PR #7, opened 2015-01-22
Added more sm versions to travis

ghost:


sarabveer:
All of your Commits have failed on TravisCI, and the current script is fine. I will be only using SourceMod releases, not snaphots.

PR #12, opened 2015-02-10
Fixed Steam3 Support

ghost:


ghost:
$steam is wrong, replace this with $match ( https://github.com/Sarabveer/SourceBans/blob/master/web_upload/pages/admin.kickit.php#L108 )

sarabveer:
Done!

PR #15, opened 2015-03-29
Migrate Travis to the new container-based infrastructure

winstliu:
:racehorse: :dash: :zap: 3000% faster build times!  Also greatly simplified the build.

@WildCard65 asked me to do this.


winstliu:
Question: Some of the includes that you're bundling in this repo can be accessed remotely-would you rather that those are removed and instead downloaded each time through Travis?

sarabveer:
Weird, this one runs for 50secs, while the old one only runs for 20-30sec.

WildCard65:
It might be from DDos github is being attacked with.

sarabveer:
I just need the include for SteamWorks.

WildCard65:
Scratch my previous comment, it was this command that added the extra seconds: sudo -E apt-get -yq update &>> ~/apt-get-update.log  
The command is executed by travis itself.

winstliu:
I'll fix the conflict later(TM).

winstliu:
Ok, time to delete this fork and try again.

PR #16, opened 2015-03-29
Remove local copy of steamworks.

WildCard65:


WildCard65:
Builds may fail due to DDos git is experiencing.

PR #17, opened 2015-03-31
Migrate Travis to the new container-based infrastructure

winstliu:
:racehorse: :dash: :zap: Builds now start 3000% faster!
Replaces the very broken PR #15.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,20 @@
-language: bash
+#Use the new container-based infrastructure
+sudo: false
 
-env:
-  - VERSION=1.7.0
+#Install some apt packages needed for spcomp
+addons:
+  apt_packages:
+    - lib32stdc++6
 
-matrix:
-  fast_finish: true
-
-before_install:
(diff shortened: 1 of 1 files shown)


winstliu:
The `env:` is also gone because there's really no point when it's being used a whole once.  So I inlined it.

sarabveer:
We need the env: becuase we test build with many different SM versions.

Hey, I just remembered, who removed 1.6.3?

winstliu:
1.6.3 wasn't there when I first edited.  It's not in the master branch either :P.
So which versions specifically do you want to be included?

sarabveer:
1.7.0 and 1.6.3

And when 1.8.0 comes out, I'll add that too.

winstliu:
Done!

sarabveer:
One last thing before I merge. Can you replace $VERSION with $SOURCEMOD

winstliu:
:heavy_check_mark:

PR #30, opened 2015-07-16
Add CreateFlagLetters() Stock

stickz:
Proceduralize code to cleanup OnPluginStart().

Add constant for size control.


sarabveer:
Test the compiled plugin, see if the plugin still works like normal.

stickz:
Isn't this a development repo? If people are using it instead of stable it's likely because they want to test it :+1: Like myself who has it live on a server for when it updates. If I have to drag and drop plugins to my folder constantly, hardly nothing will get done interms of development.

sarabveer:
The dev plugin has updater, so it automatically updates. But future dev plugins wont have this.

PR #31, opened 2015-07-16
Transitional Syntax

stickz:
Add polymorphic convars and change corresponding retrieval methods.


stickz:
I might have anther commit for this pull request.

Edit: Nvm SourceSleuth is the only plugin which needs to store convars.

sarabveer:
Nice, break SM 1.6 more. LOL

PR #34, opened 2015-07-16
Improve String Handling

stickz:
Add comment line instead of using global string for a non-repeated event.

Use decl instead of new keyword for faster string casting.


sarabveer:
what, nothing is changed?

stickz:
Looks like it didn't commit properly :-1:

sarabveer:
https://github.com/Sarabveer/SourceBans-Fork/pull/35

PR #35, opened 2015-07-16
Improve String Handling

stickz:
Add comment line instead of using global string for a non-repeated event.

Use decl instead of new keyword for faster string casting.

_Do not merge this yet, I gotta reverse something_


sarabveer:
wtf, i cant merge this

sarabveer:
Too fucked up :(

PR #37, opened 2015-07-16
Improve CanUseSourcebans

stickz:
Cast directly from equals operator instead of an if statement.

Anther commit to follow if required.


sarabveer:
Look at around 244 on sourcebans.sp

stickz:
I found no code around there that can benefit from this.

PR #40, opened 2015-07-17
Cache sizeof() to Varriable

stickz:
We should store this value in a variable instead of recalling the function each time.


stickz:
Pull request #39 accidentally got integrated into this. Maybe try merging it first.

Edit: this already got merged.

PR #42, opened 2015-07-21
Replace if with switch

stickz:
Convert if statements in menu action code to switch statements.

Remove redundancy in if statement checking.

Attempt to clean code by moving some debug lines inward.


stickz:
Hopefully I didn't make any mistakes here. GitHub is a bloody mess displaying my changes.

PR #45, opened 2015-07-23
Improve FWBlock Function

stickz:
Add switch statement with labeled constants. (faster and easier to understand)

Proceduralize code by adding setMute and setGag stocks to remove switch statement redundancy. 

Use decl instead of new to create variables. (skip the initialization stage)

Move variable out of loop for a single cast. (instead of 1 x MaxClients casts)


stickz:
Forgot couple arguments, on it.

stickz:
Fixed.

PR #48, opened 2015-08-06
Email injection fix.

obtoon:
This should fix the email injection bug.

This is untested as I don't have a SourceBans: Reloaded installation.


sarabveer:
Better Fix here: https://github.com/Sarabveer/SourceBans-Fork/pull/49

Issue #60, opened 2015-09-12
Mysqli error in logs


askovpen:
Details:

```
[256] mysqli error: [1231: Variable 'optimizer_switch' can't be set to the value of 'block_nested_loop=off'] in EXECUTE("set session optimizer_switch='block_nested_loop=off';") 
Fatal error on line 77 in file /xxx/includes/adodb/adodb-errorhandler.inc.php
```

Parent Function

```
/xxx/includes/adodb/adodb-errorhandler.inc.php - 77
/xxx/includes/adodb/adodb.inc.php - 1071
/xxx/includes/adodb/adodb.inc.php - 1046
/xxx/pages/page.banlist.php - 42
```

Query String

```
p=banlist&advSearch=STEAM_X:X:XXX&advType=steamid&Submit
```

IP  X.X.X.X

ps: 'mysql  Ver 15.1 Distrib 10.0.21-MariaDB'

sarabveer:
MariaDB not supported.

cmer81:
same problem for me :/

duhow:
Same solution. No support for MariaDB. :pensive:

sarabveer:
Lol XD

PR #61, opened 2015-09-13
mariadb support

askovpen:
#60

### DIFF
--- a/web_upload/pages/page.banlist.php
+++ b/web_upload/pages/page.banlist.php
@@ -37,7 +37,7 @@ if (isset($_GET['page']) && $_GET['page'] > 0)
 	$page = intval($_GET['page']);
 	$pagelink = "&page=".$page;
 }
-if (version_compare($GLOBALS['db_version'], "5.6.0") >= 0)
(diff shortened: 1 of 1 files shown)


sarabveer:
No MariaDB support.

askovpen:
merge pull request

WildCard65:
@askovpen I think @Sarabveer means no for adding support for mariadb, I think MYSQL is only going to be supported.

sarabveer:
Next release might have a new webpanel, that might support MariaDB.

crasx:
I don't understand why this was rejected. Isn't mariadb a dropin replacement for mysql? There shouldn't be any differences in the queries themselves, just in the db name / version.

sarabveer:
@crasx  Then why does that error show?

duhow:
Despite MariaDB is a drop-in replacement for MySQL, the functions used on the PHP code are designed for MySQL. If the connector library gets updated or changed for another one that can use more DB Connectors, then MariaDB support should be added.

Until then, just use MySQL.

PR #63, opened 2015-10-01
Simplify AutoAdd

stickz:
I missed one from months ago :D


sarabveer:
Very late as 1.5.3 was already released!

PR #64, opened 2015-10-01
Simplify Return Statements on Stocks

stickz:
We can improve performance and cut down on code by moving evaluations directly to the return statement.


sarabveer:
XD

PR #65, opened 2015-10-01
Improve IsAllowedBlockLength() Returns

stickz:
This commit is very beta, but could yield a moderate improvement if successful.


stickz:
If there is no chronological order involved, this code post bellow can be removed. It will return true any-ways, after evaluating the other statements.

if (length < 0)
return true;

PR #78, opened 2015-11-03
Fix ShortMessage CreateConVar

stickz:
Use Floats instead of Integers as expected by CreateConVar.


sarabveer:
Is this mergable?

stickz:
Yes.

PR #79, opened 2015-11-03
Improve serverID Commenting

stickz:
Add AUTO_RETRIEVE constant to comment what a -1 serverID means.

Notice: The out-of-date difference.


sarabveer:
sourcebans.sp(172) : error 035: argument type mismatch (argument 2)

PR #88, opened 2015-11-23
Revert "Add Plural Check Function"

sarabveer:
Reverts Sarabveer/SourceBans-Fork#87


stickz:
I seen that the compiler segment faulted, was that because of this commit? The sourcemod compiler should never do that when functioning properly, even if it the code in question has a syntax error.

sarabveer:
yea, this commit

PR #98, opened 2015-12-29
Add compiled plugins to repo

Nerus87:
Compiled on SourcePawn Compiler 1.7.3-dev+5280


sarabveer:
We only release compiled binaries of the plugins during a actual release. Since we are developing 1.5.4-dev, the plugins source is always changing, also because compiling the plugins every commit is hard.

PR #127, opened 2016-02-23
Remove 'Invalid Handle' Error & Add Warning Message

stickz:
**Remove Handle Error. If the file is missing or inaccessible do not try to load it.**
_"[SM] Native "IsEndOfFile" reported: invalid handle 0 (error: 4)"_

**And Add Warning Message**
_"[SM] Warning File: "configs/sourcesleuth_whitelist.cfg" missing or inaccessible."
"[SM] Continuing without loading ip address whitelists."_

### DIFF
--- a/game_upload/addons/sourcemod/scripting/SourceSleuth.sp
+++ b/game_upload/addons/sourcemod/scripting/SourceSleuth.sp
@@ -253,12 +253,20 @@ public LoadWhiteList()
 	
 	new Handle:fileHandle = OpenFile(path, "r");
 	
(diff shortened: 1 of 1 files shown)


sarabveer:
Nice, but can you also fix the SourceComms blank reason issue, or messed up reason issue?

stickz:
I'm still looking at sourcecomms trying to figure out why that is happening.

sarabveer:
I can't understand either, but you have more plugin scripting knowledge then me

PR #129, opened 2016-02-23
Attempt to Fix Sourcecomms

stickz:
Revert 'decl' keyword to 'new' for reason variables. 

Possible fix for reason issue, shouldn't hurt anything to merge.


asherkin:
Don't use `decl`, put them all back to `new`.

This is exactly why `decl` has been removed from SourcePawn, it's a major source of bugs when people use it without understanding.

Sarabveer:
Replace `decl` with `new` on sourcebans.sp, sbchecker.sp, SourceSleuth.sp, sb_admcfg.sp.

I will merge when those commits are added

stickz:
I wish those issues were just reported in compile warnings. I think the lack of understanding about that feature comes from letting plugins compile without a warning about improper syntax.

asherkin:
A compiler warning for what? `decl` is exactly YOU telling the compiler that you know better than it that the variable doesn't need initialisation - if you then go and assume the variable is in a consistent state... well that's on you.

PR #131, opened 2016-02-24
Improve File Warning Message

stickz:


sarabveer:
Not needed

stickz:
The last warning message I created was horrible. And is missing a critical element relating to file syntax.

Edit: PR #127 and #128. This feature was added by myself to catch an invalid handle. And communicate to the user the actual plugin with the white list file.

sarabveer:
its the same and is uneeded

stickz:
"Incorrect Formatting" is anther issue which causes the invalid handle. Like if the user accidentally removes a bracket or puts something into the file which shouldn't be there. This was not communicated properly by the previous pull request which introduced the warning message.

sarabveer:
this seriously is not a plugin breaking issue.

PR #150, opened 2016-04-23
Cleanup SourceSleuth Code

stickz:
The purpose of a variable is to cache repeated tasks to improve efficiency; not, to explain your code and increase memory consumption. This commit replaces that bad practice with commenting; thus, reducing the compiled the plugin size and increasing code efficiency.


sarabveer:
Does it matter? Makes the code harder to read and makes the plugin like 1ms faster.

PR #171, opened 2016-07-18
Fix security vulnerability related to empty passwords

koteq:
NOTICE: though this pull request fix the vulnerability it also breaks oauth authorization because oauth use an account's password as a cookie token to perform user authorization.


sarabveer:
Fix will be merged when another commit fixing OAuth is added

koteq:
There could be several possible solutions to fix OAuth.
1. The Right Way. Create new field on sb_admins table and populate it with random token to use it in 'remember me' routines.
2. Populate password field with random string and keep every thing ugly as it is.
3. Start a session in steamopenid.php so user could use the system while session is not obsolete. After some time of inactivity (default is 24 minutes) user have to repeat OAuth authorization.

Which one do you preffer more?

sarabveer:
SourceBans itself is ugly. My concern is to keep compatibility with older SB++ Versions and the Official SourceBans. As of right now, you can still revert to 1.4.11 without any hiccup.

I would prefer to do 2 or 3, whichever will not mess with the database structure and/or compatibility as much.

koteq:
After some thoughts about the issue I've come up to idea that in previous versions of SB was a rule that account with web-permissions has to have a non empty password. It wasn't properly enforced though. Maybe OAuth break for those users with empty password is desired behaviour? Just set them a password or reset it by email and OAuth will work again.

I could add a popup to inform a user if he has no password set to restore it or ask an admin to set it for him.

koteq:
Now it should be ready to merge I think. Should I modify `SB_GITRev` somehow?

sarabveer:
Its in init.php

PR #199, opened 2016-11-22
Fix SQL and XSS injection

CrazyHackGUT:
Something user hack my SourceBans with this injection. SourceBans "generates" PHP error due to email[] in query string, and log error in system logs. After, while opening system logs, code in `<script>` tags executes from `message` because this is not escaped. It can help hackers to steal cookies admin when entering the settings.

I hacked in this way, and I do not want to hurt anyone else.
From my system log: log screen

Sorry if that, for my English. I did not know much, I'm from Russia

### DIFF
--- a/web/pages/admin.settings.php
+++ b/web/pages/admin.settings.php
@@ -126,6 +126,7 @@ foreach ($log as $l) {
     $log_item['user']     = !empty($l['user']) ? $l['user'] : 'Guest';
     $log_item['date_str'] = SBDate($dateformat, $l['created']);
     $log_item             = array_merge($l, $log_item);
+    $log_item['message']  = str_replace("\n", "<br />", htmlentities(str_replace(["<br />", "<br>", "<br/>"], "\n", $log_item['message'])));
     array_push($log_list, $log_item);
 }
 // Theme stuff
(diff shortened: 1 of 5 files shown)


koteq:
@CrazyHackGUT is right the vulnerability is real. Firstly because of dumb db query builder [here](https://github.com/sbpp/sourcebans-pp/blob/f3dc90b89e93167f0cd9dac59b4a1c2ce9787fd4/web/includes/adodb/adodb.inc.php#L1173), which should be replaced by throw statement. And secondly because of output escaping lack.

Also I have to mention that it's totally ok to use force push to modify your pull request commits.

Technoblazed:
The majority of sourcebans is completely fucked. But no one can be bothered to do a rewrite.

CrazyHackGUT:
@Technoblazed, we have on one forum is an active discussion of whether it is necessary to rewrite the SourceBans from scratch with using framework.

Technoblazed:
Honestly, I'll probably end up doing it next summer, because I'm always bored as fuck.

Groruk:
@galexrt and I are actually planning on rewriting SourceBans++ from scratch. Sadly, at the time we are busy with RL stuff but hope to start development in between Christmas and new year.

Technoblazed:
@Groruk Let me know if you need anyone else :D

Issue #234, opened 2017-02-26
[sbpp_main.smx] Query Failed


ezspot:
Server type: MySQL
Server version: 5.5.54-0ubuntu0.14.04.1-log - (Ubuntu)

Plugins:
(1.8.0.5970) by AlliedModders LLC
"SourceBans++: Bans Checker" (1.6.0) by psychonic, Ca$h Munny, SourceBans++ Dev Team
"SourceBans++: Admin Config Loader" (1.6.0) by AlliedModders LLC, SourceBans++ Dev Team
"SourceBans++: SourceSleuth" (1.6.0) by ecca, SourceBans++ Dev Team
"SourceBans++: Main Plugin" (1.6.0) by SourceBans Development Team, SourceBans++ Dev Team
"SourceBans++: SourceComms" (1.6.0) by Alex, SourceBans++ Dev Team

```
Error Message:
L 02/26/2017 - 04:47:22: [sbpp_main.smx] Query Failed: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 1
```

Fresh install using [LGSM](https://github.com/GameServerManagers/LinuxGSM/wiki)

ezspot:
Anyone? :)

Groruk:
To resolve issues faster you could atleast give some information when this error happens or compile the plugin with the [debug](https://github.com/sbpp/sourcebans-pp/blob/v1.x/game/addons/sourcemod/scripting/sbpp_main.sp#L53) constant defined to give some additional information. [Here](https://coenjacobs.me/2013/12/06/effective-bug-reports-on-github/) is also a guide on bug reporting, please read it. (I also didn't know that a MySQL server can run linux)

ezspot:
Seems like it happends when the round is over.

ezspot:
and with changemap.

Technoblazed:
Read @Groruk's previous post. Do what he asked and provide what he asked.

galexrt:
As @Groruk already mentioned:
* give some additional info
* use the issue template https://github.com/sbpp/sourcebans-pp/blob/v1.x/ISSUE_TEMPLATE.md

koteq:
[...]

PR #240, opened 2017-03-13
Remove legacy query leftover

koteq:
subj, see #234


game/addons/sourcemod/scripting/sbpp_main.sp
```diff
@@ -993,6 +993,7 @@ public GotDatabase(Handle:owner, Handle:hndl, const String:error[], any:data)
 
 	DB = hndl;
 
+	decl String:query[1024];
```
koteq: Didn't notice that it's used below. What a shame.

Issue #188, opened 2016-10-06
SourceBans Checker to include comms blocks


sneak-it:
Just a bit of a suggestion/enhancement/feature request, it would be nice if the SourceBans checker plugin also counted & announced how many comms blocks a connecting user has, in similar fashion to previous # of bans being listed.

Technoblazed:
I wrote something for this a while back, no idea where the code went though \o/. Also implemented mute evasion into sourcesleuth

sneak-it:
@Technoblazed If you ever run across it and would like to share, I'm sure several people would be interested.

Technoblazed:
I'll take a look for it this weekend :)

sarabveer:
@Technoblazed if you ever find it, re-implement it with the latest version of SourceSleuth and submit a PR.

DataWorm:
Seems he didn't find it or forgot about it. Anyway I also missed that feature so I just checked the code and modified it today and in the end successfully tested it with some help. sm_listcomms is now implemented but we might try to make a few more improvements. Going to make a merge request when I am done. 
Also I should mention that I never programmed on source scripts before so sorry if there is a little but code duplication which I preferred to keep it simple and avoid additional bugs.

DataWorm:
@sneak-it Job is done now. If you still need that feature you can just checkout the version from my merge request.

Groruk:
Added with #398

PR #398, opened 2018-02-16
integrated comms in checker plugin

DataWorm:
As it was already mentioned in #188 this merge request adds support for comms to the checker plugin. So it does add the command sm_listcomms and also shows the amount of comms (if any) when a player joins. Code was successfully tested on our servers, all features seems to work just fine as expected.

### DIFF
--- a/game/addons/sourcemod/scripting/sbpp_checker.sp
+++ b/game/addons/sourcemod/scripting/sbpp_checker.sp
@@ -28,6 +28,7 @@
 
 #define VERSION "1.6.3-PRE"
 #define LISTBANS_USAGE "sm_listbans <#userid|name> - Lists a user's prior bans from Sourcebans"
+#define LISTCOMMS_USAGE "sm_listcomms <#userid|name> - Lists a user's prior comms from Sourcebans"
 #define INVALID_TARGET -1
 
 new String:g_DatabasePrefix[10] = "sb";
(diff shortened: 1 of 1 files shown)


rumblefrog:
Good job on bringing up forgotten issues.

sneak-it:
This is awesome, thank you. Tested and working well across multiple servers, completely error-free. Incredibly helpful!

sarabveer:
We don't really need backwards compatibility for this command.

Groruk (approved):

SkippYSB:
Good job! Great feature. 
I have a problem though after i updated the sourcebans plugins with this new versions. On several game servers when administrators ban via menu the menu just disappears without any action, nothing in error logs  either. Is there any context? Still works via commands in-game. This problem occurs on both CS:S/CS:GO. SM 1.8

DataWorm:
I don't think it is related to this ticket cause we only made changes on the checker plugin that doesn't contain code for ban menu nor does it create any bans. Anyway maybe you can tell the exact version you are using now and especially the version you used before if you still know that. That could help to identify the commits in which plugins got modified.

SkippYSB:
Okay, thank you for a quick reply.
The old version i used was the first commit from 1.6.2, now i use the commit where you updated the checker. When i updated i just downloaded the SourceBans++ 1.6.3 - PRE.

Edit:
Shall i try go back to 1.6.2 without changing checker?

DataWorm:
Guess that might be a good test. Checked the history and most likely that commit might be related to your issue:
https://github.com/sbpp/sourcebans-pp/commit/8518f3129fcdc9e8ef46a5c5e5ad7319e98136e6#diff-de25bbf1235c858d4088ab16c0f45aed

SkippYSB:
Test = success. The problem is now solved. Yes probably the reason, thank you for your time and help!

PR #458, opened 2018-05-02
Updater rework

Groruk:
## Description

- Reworked updater to use a new Updater class.
- Fixed template issues
- All Updates now use PDO instead of ADODB
- Charset is now passed as parameter for Database class
- Removed update support for SourceBans RC versions


Bara:
Any reason why you set the default charset to utf8 but here https://github.com/sbpp/sourcebans-pp/pull/458/files#diff-8bdf163b47c46dcc7485c0e0a0370224R15 to utf8mb4? It's a bit confusing

Groruk:
All update files are for older SourceBans or SourceBans++ versions to update (`utf8mb4` support was officially added in 1.6.0 (update file [600](https://github.com/sbpp/sourcebans-pp/blob/v1.x/web/updater/data/600.php))) so every update before that will still be `utf8` (just for compatibility reason).

PR #470, opened 2018-05-18
SteamID lib added

Groruk:
## Description

Added my 'new' SteamID conversion library to replace the existing SQL conversion functions.

## Motivation and Context

The new lib supports `GMP`, `bcmath` and as fallback `SQL` for Steamid conversion
and with that could have a massive performance increase when using `GMP` or `bcmath`.
It also supports a better way to resolve steam's custom urls.

## Features needed until merging
- [x] Add special support for xPaw's SourceQuery lib (SourceQuery.php L307-319) 
      Turns out bcmath doesn't support logical operation (e.g AND, OR, XOR) and SQL would be a mess... so since we don't use the server steamid it will return 0 now if it cant find a suitable calculation method.
- [x] Force admin Steam2 ids to begin with `STEAM_0`

## Testing needed until merging

- [x] Steam Openid login (steamopenid.php)
- [x] Edit comm bans (admin.edit.comms.php)
- [x] Edit bans (admin.edit.ban.php)
- [x] Edit admin details (admin.edit.admindetails.php)
- [x] banned_user.cfg import (admin.bans.php)
- [x] checkMultiplePlayers() (system-functions.php)
- [x] GetCommunityName() (system-functions.php)
- [x] BlockPlayer() (admin.blockit.php)
- [x] KickPlayer() (admin.kickit.php)
- [x] xajax_AddAdmin() (sb-callback.php)
- [x] xajax_KickPlayer() (sb-callback.php)
- [x] xajax_PasteBan() (sb-callback.php)
- [x] xajax_AddBan() (sb-callback.php)
- [x] xajax_BanMemberOfGroup() (sb-callback.php)
- [x] xajax_BanFriends() (sb-callback.php)
- [x] xajax_ViewCommunityProfile() (sb-callback.php)
- [x] xajax_AddBlock() (sb-callback.php)
- [x] xajax_PasteBlock() (sb-callback.php)

## Types of changes

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.


Bara:
How about https://github.com/xPaw/SteamID.php ?

Groruk:
>The new lib supports `GMP`, `bcmath` and as fallback `SQL` for Steamid conversion.

xPaw's library only supports `GMP`...

Bara:
I hope it works with CommunityIDs like 76561193872069853. Example output: https://steamid.io/lookup/76561193872069853

rumblefrog:
Would a new updater data be made to properly convert the admin steamids? ie. correcting the universe

Groruk:
@RumbleFrog yeah I will push some changes that forces admin Steam2 ids to begin with `STEAM_0` until we switch to Steam64 IDs

CrazyHackGUT:
> we switch to Steam64 IDs

Bad-bad-bad. Maybe you should use Steam Account ID? In SourceMod API [we have function for obtaining him](https://sm.alliedmods.net/new-api/clients/GetSteamAccountID), and Account ID is guaranteed integer and takes 4 bytes from memory.

rumblefrog:
> Bad-bad-bad. Maybe you should use Steam Account ID? In SourceMod API we have function for obtaining him, and Account ID is guaranteed integer and takes 4 bytes from memory.

Perhaps, we should take all forms of Steam IDs, but handles it one way internally

Groruk:
>Bad-bad-bad. Maybe you should use Steam Account ID? In SourceMod API we have function for obtaining him, and Account ID is guaranteed integer and takes 4 bytes from memory.

I still think Steam's 64bit IDs are the way to go. They would be a `uint64` (8 bytes of data), can also be retrieved with the function [`GetClientAuthId()`](https://sm.alliedmods.net/new-api/clients/GetClientAuthId), but would contain all information stored in a [SteamID](https://developer.valvesoftware.com/wiki/SteamID).

Also Steam64 is the only format accepted from most of Steam's WebAPI endpoints and since SourceBans++ is more and more relying on the Steam WebAPI for complex features it would be good to store the complete ID in the database (would mean less internal conversions / calculations).
And there could be a high chance of Steam64 being the ID format for Source2 (effectively future proofing SourceBans++ a bit)

CrazyHackGUT:
> can also be retrieved with the function `GetClientAuthId()`

In TF2 sometimes SteamID64 not returned by this function.
Also noticed the same bug in CS:S.

I don't know how it works, but to deny the existence of these bugs is meaningless. So i continue think about Steam Account ID.

good-live:
The namespace for the SQL Class seems to be wrong.

In https://github.com/sbpp/sourcebans-pp/blob/v1.x/web/includes/SteamID/SteamID.php#L17 it try's to load SteamID\SQL
but it's declared as SteamID\calc\SQL https://github.com/sbpp/sourcebans-pp/blob/v1.x/web/includes/SteamID/calc/SQL.php#L3

PR #510, opened 2018-09-05
Hide host name

8guawong:
## Description
use gethostbyname to show resolved IP

## Motivation and Context
My server was getting ddosed so i switched to using dynamic ip to host my server.  Now for the server to show up on dashboard i need to use a dynamic dns for the Server IP/Domain field so i don't have to manually update the ip when ip changes.  However, i noticed if server is down it'll show the dynamic dns that was used and attackers would just ddosed the dynamic dns.  Hence my motivation to edit sb++ to show resolved IP instead of dynamic dns.  

## How Has This Been Tested?
Add a server using a dynamic dns provided by no-ip.org and turn the server off and check to make sure it shows the ip not the dynamic dns  

## Types of changes

## Checklist:

### DIFF
--- a/web/includes/sb-callback.php
+++ b/web/includes/sb-callback.php
@@ -1223,9 +1223,9 @@ function ServerHostPlayers($sid, $type="servers", $obId="", $tplsid="", $open=""
         $players = $query->GetPlayers();
     } catch (Exception $e) {
         if ($userbank->HasAccess(ADMIN_OWNER)) {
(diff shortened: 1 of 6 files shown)


8guawong:
i have no clue what to check for

My code follows the code style of this project.
My change requires a change to the documentation.
I have updated the documentation accordingly.

geominorai:
This would be better as a toggle option in the settings for those who prefer showing users an easier to remember host name instead of the resolved IP.

8guawong:
Don't really know how to do that...
And I think server owner who wants can just put the domain name in the Intro Text field

PR #526, opened 2018-10-18
Fix #521

### LINKED ISSUE #521: Security: XSS vulnerability from dash_intro_text
$dash_intro_text is initialised on `init.php:28` from an untrusted $_GET source. This value is used only in `pages/admin.settings.php:239`.

This variable normally contains the markup text from the WYSIWYG editor in the SourceBans edit global settings dialogue. This value seems to deliberately subvert the FILTER_SANITIZE_STRING that takes place in `init.php:33` to allow server operators to set their introduction text to have markup. Arbitrary HTML can be provided to this value which is then naively used to update the database with its contents, which are now rendered to each visitor of the SourceBans installation.

An issue like this would normally be considered low-impact, as the only ones who can change the MOTD text are the server owners themselves, and thus could generally modify the page to their wishes anyway. Unfortunately, as seen in #520, there are sometimes weaknesses that allow attackers to thus use this as an XSS avenue.

Sanitising arbitrary HTML is not an easy task, and may be worth visiting using BBcode as a markup in your WYSIWYG editor instead of HTML. You can then parse the BBcode to render the page, which would be a much safer alternative.

radek3911:
## Types of changes

## Checklist:

### DIFF
--- a/web/init.php
+++ b/web/init.php
@@ -24,10 +24,6 @@
 		Licensed under CC BY-NC-SA 3.0
 		Page: <http://www.sourcebans.net/> - <http://www.gameconnect.net/>
 *************************************************************************/
-//Hotfix for dash_intro_text
-if (isset($_POST['dash_intro_text'])) {
-    $dash_intro_text = $_POST['dash_intro_text'];
-}
 //Filter all user inputs
 //Should be changed to individual filtering
 $_GET = filter_input_array(INPUT_GET, FILTER_SANITIZE_STRING);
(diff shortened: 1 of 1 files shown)


Groruk:
This PR is not a fix at all.

The dashboard intro text should be customizable by the end user with the use of a WYSIWYG editor (TinyMCE in this case), stripping out HTML tags would defeat this whole feature. The correct way to fix this would be the use of either `BBCode` (what the issue suggests) or something like `markdown`, instead of HTML tags.

PR #572, opened 2019-05-22
Changed the Name of the Server to actually Server in the DB

DJPlaya:
## Description
This actually had to be done. For example, if you have a Plugin using the SBPP_BanPlayer Native, you have to use an Admin ID, 0 stands for the Server, so it also should be named like that on the WebPanel.

## Motivation and Context
Bans given by the Server(Admin ID 0), should also have the Admin name "Server" instead of "CONSOLE" on the WebPanel. "CONSOLE" could be confusing for most Players and i dont see a Point to keep this Name.

## How Has This Been Tested?
I allways do so when i set up SBPP.

## Types of changes

## Checklist:

### DIFF
--- a/web/install/includes/sql/data.sql
+++ b/web/install/includes/sql/data.sql
@@ -57,4 +57,4 @@ INSERT INTO `{prefix}_settings` (`setting`, `value`) VALUES
 ('auth.maxlife.steam', '10080');
 
 INSERT INTO `{prefix}_admins` (`aid` ,	`user` , `authid` ,	`password` , `gid` , `email` ,	`validate` , `extraflags`, `immunity`) VALUES
-(0 , 'CONSOLE', 'STEAM_ID_SERVER', '', '0', '', NULL, '0', 0);
+(0 , 'Server', 'STEAM_ID_SERVER', '', '0', '', NULL, '0', 0);


Groruk:
"Server" could also be confusing, in that a user could interpret it as a singular (game) server and not as all actions done via natives / rcon / etc. (Yes, this may or may not be trivial, but I wouldn't count it out, especially not with some of the people encountered on the SourceBans++ Discord)

Maybe some further input from other users would be good on this topic.

---

Also I would make the name more distinguishable from other admins by making it uppercase like the current "CONSOLE" name ("Server" -> "SERVER")

---

Since this is a change to the database of SourceBans++ you would also require to write an [updater file](https://github.com/sbpp/sourcebans-pp/tree/v1.x/web/updater/data) and add it to the [update storage](https://github.com/sbpp/sourcebans-pp/blob/v1.x/web/updater/store.json).

After that increment the [config.version](https://github.com/sbpp/sourcebans-pp/blob/v1.x/web/install/includes/sql/data.sql#L53) in the installer to the new updater file number.

geominorai:
An admin triggering a ban via rcon would also show up as this, so changing that could be misleading.

DJPlaya:
Hmm, we maybe should use both Options in this Case.
Will there be any Problem if we use a negative Value as Admin ID? We may could use ID 0 for CONSOLE and -1 for SERVER or PLUGIN.

PR #579, opened 2019-07-20
Update ban page SteamID check

rumblefrog:
There are users with a much shorter Steam ID, and when attempting to ban those individuals, the check fails.

## Description

Patches length

## Motivation and Context

Fixes an issue with banning users with much shorter Steam ID

https://steamcommunity.com/profiles/76561197960265729 Should be the lowest with 11 characters.

## Types of changes

## Checklist:

### DIFF
--- a/web/pages/admin.bans.php
+++ b/web/pages/admin.bans.php
@@ -795,7 +795,7 @@ function ProcessBan()
         $('nick.msg').setStyle('display', 'none');
     }
 
-    if (($('steam').value.match(/(?:STEAM_[01]:[01]:\d+)|(?:\[U:1:\d+\])|(?:\d{17})/) && !$('ip').value) {
+    if (($('steam').value.test(/(?:STEAM_[01]:[01]:\d+)|(?:\[U:1:\d+\])|(?:\d{17})/) && !$('ip').value) {
         $('steam.msg').setHTML('You must enter a valid STEAM ID or Community ID');
         $('steam.msg').setStyle('display', 'block');
         err++;
--- a/web/pages/admin.bans.php
+++ b/web/pages/admin.bans.php
@@ -795,7 +795,7 @@ function ProcessBan()
         $('nick.msg').setStyle('display', 'none');
     }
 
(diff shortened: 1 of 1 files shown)


TheByKotik:
Why not just do it through XMLHttpRequest (JS Ajax) (or any ajax else)?
```
http://api.steampowered.com/ISteamUser/GetPlayerSummaries/v0002/?key=$SteamAPIKey&steamids=76561197960265729
```
Or (for custom url).
```
http://api.steampowered.com/ISteamUser/ResolveVanityURL/v0001/?key=$SteamAPIKey&vanityurl=$UserVanityURL
```

CrazyHackGUT:
Because for this checks, Steam API key is required.
XMLHttpRequest requires send for user Steam API key. This is insecure, because with key, user can generate GSLT tokens, for example: https://steamapi.xpaw.me/#IGameServersService

TheByKotik:
> Because for this checks, Steam API key is required.
> XMLHttpRequest requires send for user Steam API key. This is insecure, because with key, user can generate GSLT tokens, for example: https://steamapi.xpaw.me/#IGameServersService

No-no-no.
Sorry for my "russian" english.
I'll just print a scheme.
From JS -> /ajax/is-player-valid.php -> call to SteamAPI -> return results to JS.

Groruk:
I wouldn't use steam API calls for ID validation, simply because you would risk running into rate limiting and with akamai's agressive rate limiting and IP blacklisting that would be counter productive (also one of the reasons we use the steam API so sparsely in SBPP)

TheByKotik:
> I wouldn't use steam API calls for ID validation, simply because you would risk running into rate limiting and with akamai's agressive rate limiting and IP blacklisting that would be counter productive (also one of the reasons we use the steam API so sparsely in SBPP)

Anyway, I just suggested using more accurate methods. And maybe display more info about player in ban page.
Okay, I'll know that Steam API is not welcome here.

geominorai:
You could convert and copy the regex from https://github.com/sbpp/sourcebans-pp/blob/v1.x/web/includes/SteamID/SteamID.php#L64 for consistency.

Groruk (approved):

PR #632, opened 2019-11-05
Select Steam ID as default Ban Type on PasteBan action

borzaka:
## Description
Steam ID is the most effective ban type, and btw Sleuth will catch a player on the same IP.

## Motivation and Context
PasteBan action used, when you rightclick a player and select Ban/Block Comms. A form is automatically filled out with the players data. Without my modification is will select Ban Type: IP Address.
I think a few people would use IP Address as a primary Ban Type.

## How Has This Been Tested?
Modified the corresponding file on my webserver, and used right click / Ban a player, and Steam ID was selected instead of IP Address.

## Types of changes

## Checklist:


CrazyHackGUT:
I think, it should be moved in options.

borzaka:
Yes, it would be nice to move this into a global option, but every other time you want to issue a ban, the Ban Type defaults to Steam ID. So this is just an update to make the Ban Type consistent everywhere.

PR #662, opened 2020-04-09
change installation process instead of requiring navigating / deleting files yourself

Naleksuh:
## Motivation and Context

This change is not required, but makes installing much easier and less reliant on manual control. (To prevent malicious users from completing the installation, the config.php file will not be written automatically, which is a security issue on its own, but will be handled in a later PR).
## How Has This Been Tested?

I have tested them through independent deleting of directories and Header Locations to point to the install folder. They were not used under SourceBans thus far, but is planned.
## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:


Naleksuh:
In addition, GitHub automatically changed the formatting codec of system-functions.php, so it will show as though the entire document has been deleted and readded. In reality, the only change there is adding the function to delete folders.

rumblefrog:
Old, system has changed a lot since then, closing.

PR #678, opened 2020-06-14
Implement SteamID autoconvert

TheByKotik:
## Description

Add `ConvertSteamID` function.
Add `GetAccountIDBySteamID64` function with support elder browsers.
Add but not used `ResolveVanityURL` function.
Use `onblur` event for `<input id="steamid">`, `<input id="steam">` and `<input name="SteamID">` to autoconvert SteamID.

## How Has This Been Tested?
Windows:
1. Last stable Google Chrome.
2. Last stable Mozilla Firefox.
3. Microsoft Edge 44.18362.449.0

## Types of changes

## Checklist:


web/scripts/sourcebans.js
```diff
@@ -1511,3 +1511,161 @@ function swapTab(tab) {
...
+
+const GetAccountIDBySteamID64 = typeof BigInt !== typeof undefined ?
+	(SteamID) =>
+	{
+		return Number( BigInt( SteamID ) & BigInt( 0xFFFFFFFF ) );
+	}
+	:
+	(SteamID) =>
+	{
+		const BigBit = /* Bits (1 << 63...32) as decimal integers as numeric array. */
```
sapphonie, line 1523: what the hell

TheByKotik: Not all browsers support BigInt.
It's something like `var & 0xFFFFFFFF`.
This code subtracts bits `1 << 63...32` from number.

CrazyHackGUT: You can implement auto-converting feature on backend side, what should also resolve CSRF problems when user inserts URL with custom address.

TheByKotik: > You can implement auto-converting feature on backend side, what should also resolve CSRF problems when user inserts URL with custom address.

I think this is unwanted load for server and absolutely not needed for simple converting [U:1:116845363] -> STEAM_0:1:58422681.
If someone wants it can to implement call [ResolveVanityURL](https://partner.steamgames.com/doc/webapi/ISteamUser#ResolveVanityURL) through SteamAPI on server-side.

And... I already did it for client-side (except ResolveVanityURL; This blocked by CORS).

CrazyHackGUT: This is simple math, you can't make a high load for server with simple calculating SteamID v2 from v3/community id.
I would agree with your position only if "Add ban/comm/admin" page would be haunted, but no. This is another case.

TheByKotik: > I would agree with your position only if "Add ban/comm/admin" page would be haunted, but no. This is another case.

This function also must been used for search on main pages.

> This is simple math, you can't make a high load for server with simple calculating SteamID v2 from v3/community id.

And no matter how simple the functions are, we must not forget that thousands and tens of thousands of clients call them.
And the functions themselves in web server are at least several hundred.
Such a simple converting is quite possible to entrust the client.

PR #688, opened 2020-08-18
Add a native to ban a steam ID with a parameter to bind the steam ID with a name and a parameter to kick if online.

eyal282:
This is the native's structure, without it it's impossible to ban a steam ID with a name attached to it.

`/*********************************************************
 * Ban Identity from server
 *
 * @param authid		The authid to ban
 * @param admin			The admin's client index
 * @param time			The time to ban the player for (in minutes, 0 = permanent)
 * @param reason		The reason to ban the player from the server
 * @param kick			True if to kick the banned player if he is inside the server.
 * @param name			The name of the banned client that owns the identity.
 * @noreturn
 *********************************************************/
native void SBBanAuthId(const char[] AuthId, int iAdmin, int iTime, const char[] sReason, bool kick=true, const char[] Name="");`

### DIFF
--- a/game/addons/sourcemod/scripting/include/sourcebanspp.inc
+++ b/game/addons/sourcemod/scripting/include/sourcebanspp.inc
@@ -1,7 +1,7 @@
 // *************************************************************************
 //  This file is part of SourceBans++.
 //
-//  Copyright (C) 2014-2016 SourceBans++ Dev Team <https://github.com/sbpp>
+//  Copyright (C) 2014-2019 SourceBans++ Dev Team <https://github.com/sbpp>
 //
 //  SourceBans++ is free software: you can redistribute it and/or modify
 //  it under the terms of the GNU General Public License as published by
--- a/game/addons/sourcemod/scripting/include/sourcebanspp.inc
+++ b/game/addons/sourcemod/scripting/include/sourcebanspp.inc
@@ -73,6 +73,7 @@ native void SBBanPlayer(int iAdmin, int iTarget, int iTime, const char[] sReason
  * @noreturn
  *********************************************************/
 native void SBPP_BanPlayer(int iAdmin, int iTarget, int iTime, const char[] sReason);
+
 /*********************************************************
  * Ban Identity from server
  *
--- a/game/addons/sourcemod/scripting/include/sourcebanspp.inc
+++ b/game/addons/sourcemod/scripting/include/sourcebanspp.inc
@@ -44,8 +44,8 @@ public SharedPlugin __pl_sourcebanspp =
 public void __pl_sourcebanspp_SetNTVOptional()
 {
 	MarkNativeAsOptional("SBBanPlayer");
-	MarkNativeAsOptional("SBBanAuthId");
 	MarkNativeAsOptional("SBPP_BanPlayer");
+	MarkNativeAsOptional("SBPP_BanAuthId");
 	MarkNativeAsOptional("SBPP_ReportPlayer");
 }
 #endif
--- a/game/addons/sourcemod/scripting/include/sourcebanspp.inc
+++ b/game/addons/sourcemod/scripting/include/sourcebanspp.inc
@@ -1,7 +1,7 @@
 // *************************************************************************
 //  This file is part of SourceBans++.
 //
-//  Copyright (C) 2014-2019 SourceBans++ Dev Team <https://github.com/sbpp>
+//  Copyright (C) 2014-2016 SourceBans++ Dev Team <https://github.com/sbpp>
 //
 //  SourceBans++ is free software: you can redistribute it and/or modify
 //  it under the terms of the GNU General Public License as published by
@@ -41,9 +41,10 @@ public SharedPlugin __pl_sourcebanspp =
 };
 
 #if !defined REQUIRE_PLUGIN
-public void __pl_sourcebanspp_SetNTVOptional()
+public __pl_sourcebanspp_SetNTVOptional()
 {
 	MarkNativeAsOptional("SBBanPlayer");
+	MarkNativeAsOptional("SBBanAuthId");
 	MarkNativeAsOptional("SBPP_BanPlayer");
 	MarkNativeAsOptional("SBPP_ReportPlayer");
 }
@@ -72,6 +73,19 @@ native void SBBanPlayer(int iAdmin, int iTarget, int iTime, const char[] sReason
  * @noreturn
  *********************************************************/
 native void SBPP_BanPlayer(int iAdmin, int iTarget, int iTime, const char[] sReason);
+
+/*********************************************************
+ * Ban Identity from server
+ *
+ * @param authid		The authid to ban
+ * @param admin			The admin's client index
+ * @param time			The time to ban the player for (in minutes, 0 = permanent)
+ * @param reason		The reason to ban the player from the server
+ * @param kick			True if to kick the banned player if he is inside the server.
+ * @param name			The name of the banned client that owns the identity.
+ * @noreturn
+ *********************************************************/
+native void SBBanAuthId(const char[] AuthId, int iAdmin, int iTime, const char[] sReason, bool kick=true, const char[] Name="");
 
 /*********************************************************
  * Reports a player
--- a/game/addons/sourcemod/scripting/include/sourcebanspp.inc
+++ b/game/addons/sourcemod/scripting/include/sourcebanspp.inc
@@ -84,7 +84,7 @@ native void SBPP_BanPlayer(int iAdmin, int iTarget, int iTime, const char[] sRea
  * @param name			The name of the banned client that owns the identity.
  * @noreturn
  *********************************************************/
-native void SBPP_BanAccountId(const char[] AuthId, int iAdmin, int iTime, const char[] sReason, bool kick=true, const char[] Name="");
+native void SBPP_BanAccountId(const int AccountId, int iAdmin, int iTime, const char[] sReason, bool kick=true, const char[] Name="");
 
 /*********************************************************
  * Reports a player
--- a/game/addons/sourcemod/scripting/include/sourcebanspp.inc
+++ b/game/addons/sourcemod/scripting/include/sourcebanspp.inc
@@ -41,7 +41,7 @@ public SharedPlugin __pl_sourcebanspp =
 };
 
 #if !defined REQUIRE_PLUGIN
-public __pl_sourcebanspp_SetNTVOptional()
+public void __pl_sourcebanspp_SetNTVOptional()
 {
 	MarkNativeAsOptional("SBBanPlayer");
 	MarkNativeAsOptional("SBBanAuthId");
--- a/game/addons/sourcemod/scripting/include/sourcebanspp.inc
+++ b/game/addons/sourcemod/scripting/include/sourcebanspp.inc
@@ -45,7 +45,7 @@ public void __pl_sourcebanspp_SetNTVOptional()
 {
 	MarkNativeAsOptional("SBBanPlayer");
 	MarkNativeAsOptional("SBPP_BanPlayer");
-	MarkNativeAsOptional("SBPP_BanAuthId");
+	MarkNativeAsOptional("SBPP_BanAccountId");
 	MarkNativeAsOptional("SBPP_ReportPlayer");
 }
 #endif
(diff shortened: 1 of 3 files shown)


game/addons/sourcemod/scripting/sbpp_main.sp
```diff
@@ -2328,6 +2354,73 @@ public int Native_SBBanPlayer(Handle plugin, int numParams)
...
+	int time = GetNativeCell(3);
+	char reason[128];
+	GetNativeString(4, reason, 128);
+	
+	bool kick = GetNativeCell(5);
+	char name[64];
+	GetNativeString(6, name, 64);
+
+	if (reason[0] == '\0')
+		strcopy(reason, sizeof(reason), "Banned by SourceBans");
```
TheByKotik: Since this is `native` player will not be banned through SB I suppose.

game/addons/sourcemod/scripting/sbpp_main.sp
```diff
@@ -2328,6 +2354,73 @@ public int Native_SBBanPlayer(Handle plugin, int numParams)
...
+	char authid[64];
+	GetNativeString(1, authid, sizeof(authid));
+	
+	int admin = GetNativeCell(2);
+	int time = GetNativeCell(3);
+	char reason[128];
+	GetNativeString(4, reason, 128);
+	
+	bool kick = GetNativeCell(5);
+	char name[64];
```
TheByKotik: I think we must use `MAX_NAME_LENGTH` (128) const. Too short for now.

game/addons/sourcemod/scripting/sbpp_main.sp
eyal282: Any specific suggestions? "Banned by Sourcemod plugin" for now so if you think it's good it can just be merged instead of a suggestion. Also take a note that I just copied the reason from SBPP_BanPlayer:

https://github.com/sbpp/sourcebans-pp/blob/v1.x/game/addons/sourcemod/scripting/sbpp_main.sp#L2308

game/addons/sourcemod/scripting/sbpp_main.sp
eyal282: It's actually MAX_NAME_LENGTH + 1 because of the character "End Of String" ( EOS ) and it appears throughout the whole plugin.

eyal282:
@TheByKotik 

I asked in the Alliedmods Discord and according to asherkin, the issue in the checks is SBPP and not mine.

game/addons/sourcemod/scripting/include/sourcebanspp.inc
```diff
@@ -1,106 +1,120 @@
...
+ *
+ * @param authid		The authid to ban
+ * @param admin			The admin's client index
+ * @param time			The time to ban the player for (in minutes, 0 = permanent)
+ * @param reason		The reason to ban the player from the server
+ * @param kick			True if to kick the banned player if he is inside the server.
+ * @param name			The name of the banned client that owns the identity.
+ * @noreturn
+ *********************************************************/
+native void SBBanAuthId(const char[] AuthId, int iAdmin, int iTime, const char[] sReason, bool kick=true, const char[] Name="");
```
Groruk: New SourceBans++ natives have the prefix `SBPP_` as to not confuse them with natives from the original SourceBans, so you should change the name to `SBPP_BanAuthId`

game/addons/sourcemod/scripting/sbpp_main.sp
```diff
@@ -2328,6 +2354,73 @@ public int Native_SBBanPlayer(Handle plugin, int numParams)
...
+	GetNativeString(6, name, 64);
+
+	if (reason[0] == '\0')
+		strcopy(reason, sizeof(reason), "Banned by Sourcemod plugin");
+
+	char adminIp[24], adminAuth[64];
+
+	if (!time && admin && !(CheckCommandAccess(admin, "sm_unban", ADMFLAG_UNBAN | ADMFLAG_ROOT)))
+	{
+		ReplyToCommand(admin, "You do not have Perm Ban Permission");
```
Groruk: Instead of hardcoding an english string, a [Sourcemod translation](https://wiki.alliedmods.net/Translations_(SourceMod_Scripting)) would be more fitting.

game/addons/sourcemod/scripting/sbpp_main.sp
```diff
@@ -2781,4 +2875,25 @@ stock void AccountForLateLoading()
...
+{
+	char iAuthId[35];
+	
+	for(int i=1;i <= MaxClients;i++)
+	{
+		if(!IsClientInGame(i))
+			continue;
+			
+		else if(!IsClientAuthorized(i))
+			continue;
```
Groruk: Could be simplified to 
```sourcepawn
if (!IsClientInGame(i) || !IsClientAuthorized(i))
    continue;
```

game/addons/sourcemod/scripting/sbpp_main.sp
```diff
@@ -2781,4 +2875,25 @@ stock void AccountForLateLoading()
...
+			
+		GetClientAuthId(i, AuthId_Engine, iAuthId, sizeof(iAuthId));
+		
+		if(StrEqual(AuthId, iAuthId, true))
+			return i;
```
Groruk: Would this check work if a plugins calls the native with a Steam2 ID (e.g `STEAM_0:1:47783783`), but the server engine uses Steam3 IDs (e.g `[U:1:95567567]`)?

Groruk (changes requested):

game/addons/sourcemod/scripting/sbpp_main.sp
TheByKotik: Actually this param must be replaced by Steam Account ID and just formatted before SQL INSERT.

eyal282:
@Groruk 

I tried evening everything my fork was behind, possibly doing harm. Tell me if you see anything obvious please.

game/addons/sourcemod/scripting/sbpp_main.sp
```diff
@@ -2781,4 +2875,22 @@ stock void AccountForLateLoading()
...
+	
+	for(int i=1;i <= MaxClients;i++)
+	{
+		if(!IsClientInGame(i) || !IsClientAuthorized(i))
+			continue;
+			
+		GetClientAuthId(i, AuthId_Steam2, iAuthId, sizeof(iAuthId));
+		
+		if(StrEqual(AuthId, iAuthId, true))
+			return i;
```
Groruk: This is still a problem and a simple AuthIdType change will not fix this.
Best solution would be to use AccountIDs like @TheByKotik suggested.

Groruk (commented):
I fixed some of the smaller mistakes and dropped the commits that caused merge confilicts.

game/addons/sourcemod/scripting/include/sourcebanspp.inc
```diff
@@ -73,19 +73,18 @@ native void SBBanPlayer(int iAdmin, int iTarget, int iTime, const char[] sReason
...
+ * @param AccountId		The AccountId to ban as found in GetSteamAccountId
  * @param admin			The admin's client index
  * @param time			The time to ban the player for (in minutes, 0 = permanent)
  * @param reason		The reason to ban the player from the server
  * @param kick			True if to kick the banned player if he is inside the server.
  * @param name			The name of the banned client that owns the identity.
  * @noreturn
  *********************************************************/
-native void SBBanAuthId(const char[] AuthId, int iAdmin, int iTime, const char[] sReason, bool kick=true, const char[] Name="");
+native void SBPP_BanAccountId(const char[] AuthId, int iAdmin, int iTime, const char[] sReason, bool kick=true, const char[] Name="");
```
TheByKotik: AccountID have `uint32_t` type. Param must be like `const int iAccountID`.

game/addons/sourcemod/scripting/sbpp_main.sp
```diff
@@ -2876,21 +2877,28 @@ stock void AccountForLateLoading()
...
 			return i;
 	}
 	
 	return 0;
 }
+
+
+stock SteamAccountIdToSteam2(int AccountId, char[] buffer, bufferLen)
+{
+	FormatEx(buffer, bufferLen, "STEAM_1:%d:%d", AccountId % 2, RoundFloor(float(AccountId) / 2.0))
```
TheByKotik: 1. Use `%u` instead `%d` or `%i` for correct formatting account id's in future (after MAX SINGNED INT).
2. Database uses old universe steamid's as i remebmer. `STEAM_0:%u:%u`

Please, replace it by:
```
FormatEx( buffer, bufferLen, "STEAM_0:%u:%u", AccountId & 1, AccountId >> 1 );
```

game/addons/sourcemod/scripting/sbpp_main.sp
```diff
@@ -2399,8 +2399,8 @@ public int Native_SBBanAccountId(Handle plugin, int numParams)
 		GetClientAuthId(admin, AuthId_Steam2, adminAuth, sizeof(adminAuth));
 	}
 
-	char AuthId[64];
-	SteamAccountIdToSteam2(AccountId, AuthId, sizeof(AuthId));
+	char authid[64];
```
TheByKotik: You also can reduce this to `22+1 /* sizeof "STEAM_255:1:2147483647" + 1. */`.

TheByKotik:
You also changed every end-of-line in sourcebanspp.inc from `LF` to `CRLF`.
I think it would be a good idea to revert this.

game/addons/sourcemod/scripting/include/sourcebanspp.inc
eyal282: Wait, don't integers not use const and reference uses & or it's because of the uint32? I'll do what you said regardless.

TheByKotik: > Wait, don't integers not use const and reference uses & or it's because of the uint32? I'll do what you said regardless.

`const` modifier simply marks a variable for a compiler as "we don't want to change it, please warn us if we try".
Also this hint for API user about unchangeable variable in function call.

eyal282:
> You also changed every end-of-line in sourcebanspp.inc from `LF` to `CRLF`.
> I think it would be a good idea to revert this.

I have no idea what this means, I can't think of anything to fix this. All else is fixed.

eyal282:
@Groruk 

I made a simple test on my server and the plugin is working.

eyal282:
@TheByKotik 

Any issues left? I confirmed everything on plain sight indeed works.

eyal282:
I think IP Address is another thing that could and should be binded. Once the PR is accepted it may be too late. Thoughts?

game/addons/sourcemod/scripting/sbpp_main.sp
geominorai: FYI, you can check for empty strings with just `if (!reason[0])`.

game/addons/sourcemod/scripting/include/sourcebanspp.inc
```diff
@@ -1,106 +1,119 @@
...
- * @param sReason	The reason to report the player
- * @noreturn
- *********************************************************/
-forward void SBPP_OnReportPlayer(int iReporter, int iTarget, const char[] sReason);
-
-//Yarr!
+// *************************************************************************
+//  This file is part of SourceBans++.
+//
+//  Copyright (C) 2014-2016 SourceBans++ Dev Team <https://github.com/sbpp>
```
TheByKotik: You change year.

game/addons/sourcemod/scripting/sbpp_main.sp
```diff
@@ -432,9 +433,9 @@ public Action CommandBan(int client, int args)
 
 	int time = StringToInt(buffer);
 
-	if (!time && client && !(CheckCommandAccess(client, "sm_unban", ADMFLAG_UNBAN | ADMFLAG_ROOT)))
+	if (!time && client && !(CheckCommandAccess(client, "sm_unban", ADMFLAG_UNBAN)))
```
TheByKotik, line 436: You change flag but PR header tell only about adding native?

eyal282: Fixed some issues. I was told by Asherkin that it will just evaluate to ADMFLAG_ROOT, which is a bug. Just fixed some small stuff just for the sake of doing it. Also, CheckCommandAccess always returns true for root.

game/addons/sourcemod/scripting/include/sourcebanspp.inc
eyal282: Probably used an include from my folder. Is there anything else I edited by accident?

game/addons/sourcemod/scripting/sbpp_main.sp
eyal282: I use EOS always. I copied from another part so the syntax you guys use is the same.

trollollol123:
Looking forward to this. :)

eyal282:
I feel like this is going nowhere due to communication difficulties, likely on me. I hope someone can just, literally, take everything I did and add this native so we can use Ban Disconnected properly, I already have the plugin published.

Issue #702, opened 2020-10-18
Formatting error in translations/sbpp_comms.phrases.txt


trollollol123:
Multiple formatting errors in translations/sbpp_comms.phrases.txt
In some places u have } instead of ) 

https://github.com/sbpp/sourcebans-pp/blob/v1.x/game/addons/sourcemod/translations/sbpp_comms.phrases.txt

walliski:
Can you give an example of such occurrences?
Can you explain what kind of problem this causes?
Is it relevant to one language, or multiple?

rumblefrog:
Please include some examples of these "some places" and try to adhere to the issue template.

trollollol123:
https://github.com/sbpp/sourcebans-pp/blob/v1.x/game/addons/sourcemod/translations/sbpp_comms.phrases.txt#L1293
https://github.com/sbpp/sourcebans-pp/blob/v1.x/game/addons/sourcemod/translations/sbpp_comms.phrases.txt#L1265
https://github.com/sbpp/sourcebans-pp/blob/v1.x/game/addons/sourcemod/translations/sbpp_comms.phrases.txt#L1237

I think thats an error right? Im not sure if there were more.

Tiagoquix:
@Groruk #746

PR #742, opened 2021-07-23
Fix Translations

crashzk:
- Update Translations Portuguese-BR;
- Fix Translations https://github.com/sbpp/sourcebans-pp/issues/702;


Tiagoquix:
@Groruk Please do not accept this pull request. Do you remember the e-mail I sent to you a time ago? Well, the same reason applies here.

Tiagoquix:
The translations are good the way they are now. You are not fixing anything, you are reverting my changes and using English terms, that is, you are not translating anything.

crashzk:
Those responsible for the project will check this, I only heard feedback from my community because when I updated based on their translations, they asked for corrections in some sentences because, using their words, _"they translated thinking that nobody knew how to read or interpret short texts and simple?"_

Anyway, I understand your argument, but I leave it up to the developers to decide whether or not it should be merged.

PR #752, opened 2021-09-25
Fix auth

RSA-encryption:
No longer dies when using http

### DIFF
--- a/web/includes/auth/Auth.php
+++ b/web/includes/auth/Auth.php
@@ -33,7 +33,7 @@ class Auth
         self::insertNewToken($jti, $secret);
         self::updateLastVisit($aid);
 
-        self::setCookie($token, time() + $maxlife, Host::domain(), ((bool)$_SERVER['HTTPS']) ? true : false);
+        self::setCookie($token, time() + $maxlife, Host::domain(), array_key_exists('HTTPS', $_SERVER) ? (bool)$_SERVER['HTTPS'] : false);
 
         //Login / Logout requests will trigger GC routine
         self::gc();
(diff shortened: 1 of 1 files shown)


CrazyHackGUT:
`$_SERVER['HTTPS'] ?? false` is much simpler. SB++ already requires PHP with null-coalesce operator support ([added in PHP 7.0](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op)).

web/includes/auth/Auth.php
```diff
@@ -33,7 +33,7 @@ public static function login(int $aid, int $maxlife)
         self::insertNewToken($jti, $secret);
         self::updateLastVisit($aid);
 
-        self::setCookie($token, time() + $maxlife, Host::domain(), ((bool)$_SERVER['HTTPS']) ? true : false);
+        self::setCookie($token, time() + $maxlife, Host::domain(), array_key_exists('HTTPS', $_SERVER) ? (bool)$_SERVER['HTTPS'] : false);
```
alexandre433, line 36: You could also use `(bool) $_SERVER['HTTPS'] ?? false`

web/includes/auth/Auth.php
```diff
@@ -57,7 +57,7 @@ public static function logout()
             self::$dbs->execute();
         }
 
-        self::setCookie('', 1, Host::domain(), ((bool)$_SERVER['HTTPS']) ? true : false);
+        self::setCookie('', 1, Host::domain(), array_key_exists('HTTPS', $_SERVER) ? (bool)$_SERVER['HTTPS'] : false);
```
alexandre433, line 60: Same here.

Hackmastr:
Thanks! 

This has been fixed in the php81 branch.

PR #817, opened 2022-09-25
feat(main): add forward SBPP_OnClientPreAdminCheck

maximee-leroyy:
## Description
Add a forward SBPP_OnClientPreAdminCheck to control when OnClientPostAdminCheck will be called.
We can then return a boolean to block the commands that will trigger OnClientPostAdminCheck:
```
RunAdminCacheChecks(i);
NotifyPostAdminCheck(i);
```
## Motivation and Context
We are using https://github.com/R1KO/Vip-Core and we need to wait for both sourcebans-pp and vip-core to check if the player has a vip and/or an admin status so that we can trigger OnClientPostAdminCheck with the proper sourcemod admin flags

## How Has This Been Tested?
In use on https://nide.gg for about two years implemented throught this module:
- https://github.com/srcdslab/sm-plugin-VIP_SourcemodFlags/blob/master/addons/sourcemod/scripting/VIP_SourcemodFlags.sp#L140

## Types of changes

## Checklist:


maximee-leroyy:
cc @Hackmastr

PR #853, opened 2023-02-05
Improve in-game translations

Tiagoquix:
- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.

**Issue type:** bug fix (non-breaking change which fixes an issue).

This pull request:

- includes miscellaneous changes to improve the translations here and there (including English);
- adds some format strings that are missing from certain languages;
- changes some instances of "client" to "player" in the English translations;
- removes the format string from the `sourcesleuth_kicktext` string (from `sbpp_sleuth.phrases.txt`) as it does not seem to be necessary;
- copies the translations from the `Ban Not Verified` string (from `sbpp_main.phrases.txt`) to the `Player Comms Not Verified` string (from `sbpp_comms.phrases.txt`) since they are the same string and the `sbpp_main`'s string has more translations.
- improves the Brazilian Portuguese translations.

PR #854, opened 2023-02-06
Improve in-game translations (php81)

Tiagoquix:
@Hackmastr #853 

This new pull request also includes this [commit](https://github.com/sbpp/sourcebans-pp/commit/cda44c2cc50c11b76410576f42cf2ccec1fad85f#diff-36775cb8b605b223abf7f6cfcc8b0345640f32e114d1b63cfb3af05779e41d0c). However, in a safer approach, I've removed all translations from other languages for the string `"Player Comms Not Verified"`. Since the string has been changed, it's better to force English rather than trying to guess what "This player" is in other languages or using Google Translate.


Hackmastr:
Thanks!

PR #894, opened 2023-05-25
[SBPP Checker] Callback and initial values

DosMike:
## Description
As user of the SourceBans Checker API I would like to be notified when the bans and commbans counters are actually valid, or at least be provided with fallback values until they are.

## Motivation and Context
As of 1.8.0 the values in g_iBanCounts[client] and g_iCommsCounts[client] remain stale between a disconnect and the async `OnConnectBanCheck` callback. This might cause issues if you want to rely on, or react to these values as soon as they are available. If a plugin is polling values this would currently still give incorrect results for quite some time, even if the the client is ingame AND authorized, thus resetting the values to -1.

## How Has This Been Tested?
- [X] Smoke test in dev server + live env

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/game/addons/sourcemod/scripting/include/sourcebanschecker.inc
+++ b/game/addons/sourcemod/scripting/include/sourcebanschecker.inc
@@ -1,4 +1,4 @@
-﻿// *************************************************************************
+// *************************************************************************
 //  This file is part of SourceBans++.
 //
 //  Copyright (C) 2014-2023 SourceBans++ Dev Team <https://github.com/sbpp>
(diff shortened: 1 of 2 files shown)


Rushaway:
- Can you change the version to 1.8.1 ?
- Forward has to be created in the AskPluginLoad2

DosMike:
> * Can you change the version to 1.8.1 ?
> * Forward has to be created in the AskPluginLoad2

There you go :)

Rushaway (approved):

PR #896, opened 2023-05-26
Lessened server lag on player connect due to huge number of bans and gags #897

armanossiloko:
When the number of bans in the SourceBans database becomes higher, logging "Player XYZ has N bans and M comm blocks." becomes a bit more expensive, in some cases potentially lagging the server out for a short period, which can be noticeable and painful depending on the game.

## Description
The SourceBans database has a few indexes on the sb_bans and sb_comms tables. However, the query that is executed once a player connects to the game server does not utilize the indexes at all due to the REGEXP that was present in the WHERE clause. By separating bans by ban type (userId and type), the query can utilize the indexes and thus likely execute the filtering faster than before since now a full table scan is not truly required anymore. Overall, the purpose of this PR is to lessen the lag that was mentioned in #897 as it's something that has helped in one case already.

## Where was it tested?
The code was compiled and installed onto a Left 4 Dead 2 10v10 versus server.

## Types of changes

## Checklist:

### DIFF
--- a/game/addons/sourcemod/scripting/sbpp_checker.sp
+++ b/game/addons/sourcemod/scripting/sbpp_checker.sp
@@ -117,26 +117,27 @@ public void OnClientAuthorized(int client, const char[] auth)
 
 	char query[512], ip[30];
 	GetClientIP(client, ip, sizeof(ip));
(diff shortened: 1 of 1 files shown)


Rushaway:
LGTM

PR #899, opened 2023-06-05
Fix cookie domain for non-standard ports.

wrossmann:
## Description
Currently the cookie domain would be set to whatever the httpd's HTTP_HOST is. But in the case of web servers using non-standard ports, eg: foo.bar.com:8080, this would cause the cookie domain to also be set to foo.bar.com:8080.

The problem is that cookie domains do not account for ports, and subsequent requests will not send the cookie back because the domain that the browser wants is just `foo.bar.com`, not `foo.bar.com:8080`. The end result is that the browser does not send the authentication cookie on subsequent requests, and it appears that authentication simply doesn't work.

This fork adds a simple `Host::cookieDomain()` wrapper around `Host::domain()` to strip off the port, and replaces the relevant usages in the Auth class.

## Motivation and Context
Current behaviour is broken when SB web is run on a non-standard port.

## How Has This Been Tested?
Mocked the Auth workflow locally.

## Types of changes

## Checklist:

### DIFF
--- a/web/includes/auth/Auth.php
+++ b/web/includes/auth/Auth.php
@@ -31,7 +31,7 @@ class Auth
         $token = JWT::create($jti, $maxlife, $aid);
         self::updateLastVisit($aid);
 
-        self::setCookie($token->toString(), time() + $maxlife, Host::domain(), Host::isSecure());
+        self::setCookie($token->toString(), time() + $maxlife, Host::cookieDomain(), Host::isSecure());
 
         //Login / Logout requests will trigger GC routine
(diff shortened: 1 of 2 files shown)


Hackmastr:
Thanks!

conorab:
This merge fixes the issue for me. I've included a line in my Docker config to pull in the patched files from this commit: https://gitlab.com/conorab/k8s-sourcebanspp/-/commit/fc9e7ae584dc53eb0a3984262de71b6c7076c94b .

Issue #900, opened 2023-06-09
Invalid SteamID input!  - /includes/SteamID/SteamID.php on line 108


Trice1984:
Hello Team,

our problem, when we use the BAN button at the Homepage we become the error:

Fatal error: Uncaught Exception: Invalid SteamID input! in /var/www/vhosts/xloldies.de/sb.xloldies.de/includes/SteamID/SteamID.php:108 Stack trace: #0 /var/www/vhosts/xloldies.de/sb.xloldies.de/includes/SteamID/SteamID.php(84): SteamID\SteamID::resolveInputID() #1 /var/www/vhosts/xloldies.de/sb.xloldies.de/includes/SteamID/SteamID.php(56): SteamID\SteamID::to() #2 /var/www/vhosts/xloldies.de/sb.xloldies.de/pages/page.banlist.php(487): SteamID\SteamID::toSteam3() #3 /var/www/vhosts/xloldies.de/sb.xloldies.de/includes/page-builder.php(143): require_once('...') #4 /var/www/vhosts/XXXXXXX/index.php(29): build() #5 {main} thrown in /var/www/vhosts/xloldies.de/sb.xloldies.de/includes/SteamID/SteamID.php on line 108

We use the Version:
SB: 1.8.0 | Git: 1294
PHP: 8.1.19
Mysql: MariaDB

The link is: https://sb.xloldies.de/index.php?p=banlist

We hope someone can help us...

Thank u for ur help.

Trice1984:
This error is at today again back , sorry for my bad englisch.

How can i fix it?

Trice1984:
DELETE FROM sb_bans WHERE authid = 'BOT';   this one helps

Issue #906, opened 2023-07-01
Bad SteamID Format in DB can broke comms page


Rushaway:
What are the steps to reproduce this issue?
-------------------------------------------
1. insert a comms with the ban SteamID format in sb_comms table (Stop_ignoring_retvals  || steamid_pending) 
2. go on your comms pages
3. congrats, page display is broken

What happens?
-------------
Comms page doesn't display until you remove bad steamid format from the DB manualy.
[Screenshoot](https://github.com/sbpp/sourcebans-pp/assets/11679883/4736cb8f-cca1-453d-b96c-7918cd775c18)

What were you expecting to happen?
----------------------------------
Ignore the attempt to display comms performed with bad steamid format

Any logs, error output, etc.?
----------------------------

```
AH01071: Got error 'PHP message: PHP Fatal error: Uncaught Exception: Invalid SteamID input! in /includes/SteamID/SteamID.php:108
Stack trace:
#0 /includes/SteamID/SteamID.php(84): SteamID\\SteamID::resolveInputID()
#1 /includes/SteamID/SteamID.php(56): SteamID\\SteamID::to()
#2 /pages/page.commslist.php(462): SteamID\\SteamID::toSteam3()
#3 /includes/page-builder.php(143): require_once('...')
#4 /index.php(29): build()
#5 {main}\n thrown in /includes/SteamID/SteamID.php on line 108', referer: https:/yourwebsite.com/index.php?p=banlist&advSearch=STEAM_0:0:158394035&advType=steamid&Submit
```

Any other comments?
-------------------

What versions of software are you using?
----------------------------------------
**Operating System:** nginx
**SourceBans++ Version:** 1.8.0 (php 8.1 branch) with the lastest release.
**PHP Version:** 8.1.20
**MySQL Version:** 10.6.7-MariaDB-2ubuntu1 - Ubuntu 22.04
**Link to your project:** https://bans.nide.gg/index.php
**Link to a phpinfo() output:** https://upgradebans.nide.gg/phpinfo.php

Trice1984:
sory, but the problems are today again:

» [Home]» [Ban List]

Warning: Trying to access array offset on value of type null in /var/www/vhosts/xloldies.de/sb.xloldies.de/includes/system-functions.php on line 171

Warning: Trying to access array offset on value of type null in /var/www/vhosts/xloldies.de/sb.xloldies.de/includes/system-functions.php on line 171

[...]

PR #908, opened 2023-07-09
Fix: #900 & #906

Rushaway:
## Description

Check if the Steamid format is valid by using `isValidID()` from [SteamID.php](https://github.com/sbpp/sourcebans-pp/blob/php81/web/includes/SteamID/SteamID.php#L116)

For #900 I check if isset & is not empty (bcs they can be empty with sm_banip)
For #906 No need to check if set bcs all comms are based on steamid system (no ip supported)
## Motivation and Context

Prevent bad format to break page view.
Give this `$data['steamid'] = 'STEAM_0:0:00000000';` when the `isValid` return false.

## How Has This Been Tested?

For #900 I have a ban in db with `STEAM_ID_STOP_IGNORING_RETVALS` as SteamID in `authid` colume (also tested with a player name..) 
For #906 I have a comms in db with  `STEAM_ID_STOP_IGNORING_RETVALS` as SteamID in `authid` colume 
## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.

### DIFF
--- a/web/pages/page.banlist.php
+++ b/web/pages/page.banlist.php
@@ -483,6 +483,10 @@ while (!$res->EOF) {
     $data['player']      = addslashes($res->fields['player_name']);
     $data['type']        = $res->fields['type'];
     $data['steamid']     = $res->fields['authid'];
+    // Fix #900 - Bad SteamID Format broke the page view, so give them an null SteamID.
+	if (isset($data['steamid']) && !empty($data['steamid']) && !\SteamID\SteamID::isValidID($data['steamid'])) {
+		$data['steamid'] = 'STEAM_0:0:00000000';
+	}
     $data['communityid'] = $res->fields['community_id'];
     $data['steamid3']    = \SteamID\SteamID::toSteam3($data['steamid']);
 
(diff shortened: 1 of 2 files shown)


Rushaway:
nb: [3b1c56b](https://github.com/sbpp/sourcebans-pp/pull/908/commits/3b1c56be167a4d3284954683d00779909fdff564) I was thinking break can be a nice alt, but it's not a suitable solution

Hackmastr:
Hi, thanks for the PR!

I don't think breaking is the correct thing to do, instead check where the steamid is being used and avoid it (either by setting it to null or something similar.

Hackmastr:
Hi, thanks for the PR!

I don't think breaking is the correct thing to do, instead check where the steamid is being used and avoid it (either by setting it to null or something similar.

conorab:
I had this issue as well. I accidentally banned one of the bots in Left 4 Dead 2 in-game and 'BOT' was added to the ban as their SteamID. This was fine pre-1.7 but causes the page to refuse to load now. Fixed by running the following MySQL commands ('sourcebans' being my database):
`USE sourcebans`
`DELETE FROM sb_bans WHERE authid = 'BOT';`

Issue #949, opened 2024-02-15
bug: Negative ban lenght


Rushaway:
What are the steps to reproduce this issue?
-------------------------------------------
1. `sm_ban @me -1 test`

What happens?
-------------
You get banned for the session. But it's useless the ban is automatically expire.

What were you expecting to happen?
----------------------------------
Abort the ban, duration is incorrect. 

Any logs, error output, etc.?
----------------------------

N/A

Any other comments?
-------------------
N/A

What versions of software are you using?
----------------------------------------
**Operating System:** ubuntu 20.04
**SourceBans++ Version:** latest
**SourceMod Version:** 1.12
**MetaMod Version:** 1.12

sunstep:
This issue will be solved soon. Check out PR #953

Rushaway:
Fixed in bf5303b

PR #953, opened 2024-03-10
Fix negative values being accepted in ban commands (solve issue #949)

sunstep:
## Description

The changes add a check on all ban commands to see whether the given duration is 0 or more. If the duration given is smaller than zero, it will tell the player and the ban will not proceed.

## Motivation and Context

The motivation behind this pull request is that this fixes an open issue. Issue #949. According to the issue, banning with a negative time value is useless and only bans the player for the session. With this pull request banning with a negative value should not be possible anymore

## How Has This Been Tested?

The changes have been tested on my own community server on a real player. When entering a negative value, we want the entire command to stop after giving the command issuer a warning. This approach should be valid and should not affect any other areas in the code. Values of 0 or higher have also been tested and these do get accepted by the command.

## Screenshots (if appropriate):
image

## Types of changes

## Checklist:

### DIFF
--- a/game/addons/sourcemod/scripting/sbpp_main.sp
+++ b/game/addons/sourcemod/scripting/sbpp_main.sp
@@ -545,12 +545,11 @@ public Action CommandBanIp(int client, int args)
 	}
 
 	char adminIp[24], adminAuth[64];
-
 	int minutes = StringToInt(time);
 
-	if (minutes < 0)
+	if (!StringToIntEx(time, minutes) || minutes < 0)
 	{
-		ReplyToCommand(client, "%sInvalid duration. Duration must be 0 or higher", Prefix);
+		ReplyToCommand(client, "%sUsage: sm_banip <#userid|name> <time> [reason]", Prefix);
 		return Plugin_Handled;
 	}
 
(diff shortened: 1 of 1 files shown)


Rushaway:
You can keep it DRY with the existing code.
```sp
if (!StringToIntEx(buffer, time) || time < 0)
{
    ReplyToCommand(client, "%sUsage: sm_ban <#userid|name> <time|0> [reason]", Prefix);
    return Plugin_Handled;
}
```
Edit: Also please keep using tranlations, sourcebanspp is used by tons of users.

sunstep:
Implemented this way for the CommandBan function, will add more new commits soon

Rushaway:
It should also apply to every changes you made.

sunstep:
I don't see how it is possible in the other parts. The other parts don't print the right message after getting the int value of time. And since the minutes variable is used multiple times, wouldn't it be better there to just leave a separate check for if the time is negative or not?

Next to that, I can add translations for this phrase. But only when all other issues with the code I provided are resolved.

Rushaway:
To avoid create new translations, we can just stick to the current command usage phrase. 
Adjust the reply to cmd to the function ofc.
```sp
if (!StringToIntEx(time, minutes) || minutes < 0)
{
    ReplyToCommand(client, "%sUsage: sm_banip <#userid|name> <time> [reason]", Prefix);
    return Plugin_Handled;
}
```

sunstep:
Done, the approach is now used for all commands.

sunstep:
By the way, should the sourcebans version be bumped? I decided not to do it, just in case. I can always add a commit if needed.

Rushaway:
LGTM. 
No need to bump version, we will release a new version with all recent commits soon.

Hackmastr:
Thanks @Rushaway @sunstep!

Issue #965, opened 2024-06-03
[Bug] Installer - strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated 


Rushaway:
What are the steps to reproduce this issue?
-------------------------------------------
1. Run install until page 3

What happens?
-------------
![image](https://github.com/sbpp/sourcebans-pp/assets/11679883/ef580899-db90-4511-99d1-f8a45104f357)

What were you expecting to happen?
----------------------------------
Don't throw php error

Any logs, error output, etc.?
----------------------------

```
Warning: Undefined variable $sql_version in /install/template/page.3.php on line 154
Deprecated: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in /install/template/page.3.php on line 154
```

Any other comments?
-------------------

What versions of software are you using?
----------------------------------------
**Operating System:** Ubuntu 22.04
**SourceBans++ Version:** 1.8.0 - 1410
**PHP Version:** 8.2
**MySQL Version:** 8
**Link to your project:** [http://heizemod.one](http://heizemod.one/)/
**Link to a phpinfo() output:**

PR #966, opened 2024-06-03
Fix(installer): rare case where sql_version is not defined (#965)

Rushaway:
## Description

Ensure $sql_version is defined, if not give it a default value of 0.0.0 to let the rest of code run and throw error to user.

## Motivation and Context

#965 

## How Has This Been Tested?

Tested with the guys who have an issue on sbpp discord, and it's now sovled.

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly.


Rushaway:
instead of `0.0.0` maybe we can throw a message ? cc @Hackmastr

Hackmastr:
Hello,

Ideally, we should offer users the option to provide a version.
Thanks.

Rushaway:
Not sure, this not the normal behaviour to can not retrieve the version info. At this place something is already wrong. Allowing the user to provde a version does not mean the connection to the database was succesfull.

PR #1106, opened 2026-05-03
feat(ci): add staabm/phpstan-dba to type-check raw SQL (#1100)

### LINKED ISSUE #809: [8.1] Fatal Error on Logs - Search
What are the steps to reproduce this issue?
-------------------------------------------
0. Go on Sourcebans Settings -> Logs.
1. Open search bar
2. Click on Date and put only the year (ex : 2022 [See Screenshoot](https://user-images.githubusercontent.com/11679883/191668408-1313a704-a701-45e0-bdda-6f2fcc951064.png))
3. Error.

What happens?
-------------
Fatal Error.

What were you expecting to happen?
----------------------------------
Same results as Master branch (7.4)

Any logs, error output, etc.?
----------------------------

`Fatal error: Uncaught PDOException: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '' at line 1 in /var/www/vhosts/nide.gg/upgradebans.nide.gg/includes/Database.php:124 Stack trace: #0 /var/www/vhosts/nide.gg/upgradebans.nide.gg/includes/Database.php(124): PDOStatement->execute() #1 /var/www/vhosts/nide.gg/upgradebans.nide.gg/includes/Database.php(145): Database->execute() #2 /var/www/vhosts/nide.gg/upgradebans.nide.gg/includes/Log.php(78): Database->single() #3 /var/www/vhosts/nide.gg/upgradebans.nide.gg/pages/admin.settings.php(84): Log::getCount() #4 /var/www/vhosts/nide.gg/upgradebans.nide.gg/includes/page-builder.php(143): require_once('...') #5 /var/www/vhosts/nide.gg/upgradebans.nide.gg/index.php(29): build() #6 {main} thrown in /var/www/vhosts/nide.gg/upgradebans.nide.gg/includes/Database.php on line 124`

Any other comments?
-------------------

What versions of software are you using?
----------------------------------------
**Operating System:** nginx
**SourceBans++ Version:** 1.7.0 (php 8.1 branch) with the lastest release.
**PHP Version:** 8.1.10
**MySQL Version:** 10.6.7-MariaDB-2ubuntu1 - Ubuntu 22.04
**Link to your project:** https://upgradebans.nide.gg/index.php
**Link to a phpinfo() output:** https://upgradebans.nide.gg/phpinfo.php

### LINKED ISSUE #1100: Add staabm/phpstan-dba to type-check raw SQL
## Problem
The PDO migration (#1078, #1091, #1092) finished with all DB access flowing through `web/includes/Database.php`, but the SQL itself is still a string literal that PHPStan can't see into. Column renames, schema drift, and result-shape mismatches — like the `$chk['count']` vs `$chk[0]` bug called out in #1095 — are not catchable today without running the test suite against a live database.

`staabm/phpstan-dba` is purpose-built for this: it introspects the schema and type-checks raw SQL strings, including parameter binding and result fetch shapes.

## Scope
- Add `staabm/phpstan-dba` to `web/composer.json` (`require-dev`).
- Configure it in `web/phpstan.neon` with a connection to a schema-only MariaDB instance (the dev container's DB works fine; we're not executing queries, just introspecting).
- Teach it about the `:prefix_` placeholder convention used by `Sbpp\Database` — likely a small custom extension or a regex preprocessor that resolves the placeholder before analysis.
- Run a baseline pass and accept the existing violations.
- Wire CI so a fresh schema dump + PHPStan-with-DBA run is part of the gate.

## Acceptance criteria
- [ ] `phpstan-dba` installed and configured
- [ ] CI runs PHPStan with DBA enabled against the seeded dev schema
- [ ] At least one previously-shipped column-name typo or fetch-shape bug is identified and either fixed or baselined with a comment
- [ ] Documented in `docker/README.md` how the gate runs locally
- [ ] Schema changes (via `web/install/`) are required to keep CI green

## Notes
- Pairs naturally with the API-contract codegen and view-model issues. Together they cover the three remaining stringly-typed seams in the panel: JS↔PHP, PHP↔Smarty, PHP↔SQL.
- The `:prefix_` placeholder is the main integration wrinkle. Worst case we resolve it manually before passing the SQL string to phpstan-dba's analyzer.
- Catches a strict superset of what Tier 1 in #1095 catches, and does it without spinning up the full stack. Both are still worth having: phpstan-dba for the static layer, integration tests for behavior.

rumblefrog:
Closes #1100.

## Summary

- Adds `staabm/phpstan-dba` (^0.4) to `web/composer.json` and wires it into the existing PHPStan gate so raw SQL strings flowing through `Database::query(...)` are type-checked against the live MariaDB schema (column names, table names, statement syntax).
- New CI step in `.github/workflows/phpstan.yml` spins up a schema-only MariaDB 10.11 service, renders `web/install/includes/sql/struc.sql` the same way `docker/db-init/00-render-schema.sh` does locally, and points the bootstrap at it. `DBA_REQUIRE=1` keeps the gate from silently disabling itself on a creds drift or missing service.
- `./sbpp.sh phpstan` already runs in the dev container; it now exports `DBA_HOST=db` so the bootstrap reaches the running `db` service. `PHPSTAN_DBA_DISABLE=1` short-circuits when the DB is down or you're iterating on unrelated rules.
- Documented in `docker/README.md` (new "Static analysis with phpstan-dba" section).

## How the `:prefix_<table>` placeholder is handled

Two PHPStan-only files under `web/phpstan/` (excluded from the analysis path, autoloaded via a `classmap` entry in `composer.json`'s `autoload-dev`):

- `SbppPrefixAwareReflector` — decorator over `PdoMysqlQueryReflector` that does the same `str_replace(':prefix', $this->prefix, $query)` substitution `Database::setPrefix()` does at runtime, so phpstan-dba sees fully-qualified `sb_admins` instead of `:prefix_admins` (which would otherwise look like a PDO bind followed by an undeclared identifier).
- `SbppSyntaxErrorInQueryMethodRule` — replacement for phpstan-dba's built-in `SyntaxErrorInQueryMethodRule`. The upstream rule short-circuits any time `countPlaceholders() > 0`, which means `:prefix_admins` causes it to skip every query in the codebase. The new rule (1) resolves `:prefix` first, then (2) stubs remaining `:name` and `?` placeholders with the literal `1` so MariaDB will at least PARSE the statement. Quoted string literals are skipped (regex mirrors the upstream `REGEX_NAMED_PLACEHOLDER` pattern, validated by a unit smoke test) so `WHERE x = ':literal'` is preserved.

The trade-off: bound values live in follow-up `bind()` calls in scopes PHPStan can't tie back to the query string, so we get **structural validation only** — table/column existence, statement syntax, and the kind of typo this issue was filed for. That's a strict superset of what level-4 PHPStan catches today.

There's also `SbppNullReflector` (no-op) installed when `PHPSTAN_DBA_DISABLE=1` or the DB is unreachable, because phpstan-dba's eagerly-loaded service rules crash with "Reflector not initialized" otherwise.

## Real bug surfaced and fixed

`web/includes/Log.php:73,123` — the `case "date":` arm in `getQueryFromAdvancedSearch()` built `" l.created > :value AND l.created :valueOther"`, missing the comparison operator between the second `l.created` and `:valueOther`. Introduced in commit `387fface` ("Fix #809", Dec 2022). At runtime any admin-log query with `advType=date` threw SQLSTATE 42000.

The new rule flags both lines as `sbpp.dba.syntaxError`. Fix is a one-character `<` (the surrounding `mktime(... 0)` / `mktime(... 59)` pair is clearly trying to express a half-open `[start, end)` range, symmetric with the existing `>` on the start bound). Reverting the fix and re-running confirms the rule fires.

## Acceptance criteria

- [x] `phpstan-dba` installed and configured (`web/composer.json`, `web/phpstan.neon`, `web/phpstan-dba-bootstrap.php`).
- [x] CI runs PHPStan with DBA enabled against the seeded dev schema (`.github/workflows/phpstan.yml`).
- [x] At least one previously-shipped bug identified and fixed (`web/includes/Log.php`).
- [x] Documented in `docker/README.md`.
- [x] Schema changes (via `web/install/includes/sql/struc.sql`) are required to keep CI green — CI re-renders the file on every run, so a column rename without updating callers will fail.

## Test plan

- [x] `./sbpp.sh phpstan` — passes against the existing baseline + the new DBA rule.
- [x] `PHPSTAN_DBA_DISABLE=1 ./sbpp.sh phpstan` — graceful skip, still passes.
- [x] `DBA_HOST=does-not-exist DBA_REQUIRE=1 phpstan analyse` — fails with `exit 1` and prints "DBA_REQUIRE=1, aborting" (CI behaviour).
- [x] `./sbpp.sh test` — 37/37 pass, no regressions from the new files.
- [x] Stash the `Log.php` fix and re-run `./sbpp.sh phpstan` — rule fires `sbpp.dba.syntaxError` on both lines, then passes again with the fix restored.
- [x] Unit-style sanity check on `stubPlaceholders()`: 6 cases including placeholders inside single-quoted and double-quoted string literals (`':literal'`, `'has?question'`, `"a:b"`) all preserve the literal correctly.
- [ ] CI run on this PR (Actions will execute the new MariaDB-backed PHPStan job).

## Notes for reviewers

- `web/phpstan/` is intentionally outside the project's PSR-4 root (`includes/`) so these PHPStan-only files are never autoloaded by the runtime app. They're picked up via `autoload-dev` `classmap` for PHPStan's DI and via explicit `require_once` in the bootstrap for safety.
- `prefix: sb` is hard-coded in `phpstan.neon` and matches the dev/CI default (`docker-compose.yml`, `docker/db-init/`, the new CI render step). The reflector reads `DBA_PREFIX` from env and defaults to `sb` too — they have to stay in sync. PHPStan's `%env.NAME%` syntax doesn't support defaults cleanly, hence the duplication with a comment pointing at it.
- `web/includes/vendor/` is gitignored, so the new dep doesn't bloat the diff.


rumblefrog:
@copilot resolve the merge conflicts in this pull request

Issue #1123, opened 2026-05-04
Modernize the default web panel theme (2026 redesign)


rumblefrog:
Is your feature request related to a problem? Please describe.
-------------------------------------------------------------

The default theme (`web/themes/default/`) is the original SourceBans 2007–2014 visual language — fixed 984px column, Verdana 11px, chocolate `#38322c` chrome, rust `#b05015` accent, Vista-era 32×32 raster icons, half-second `all 0.5s ease` transitions. It has been in use largely unchanged for ~15 years.

Concrete pain points:

- **Not responsive.** The 984px column doesn't reflow; the panel is unusable below ~tablet width. A growing share of admins moderate from a phone.
- **Not accessible.** 11px Verdana, low-contrast pastels for ban states, no focus-visible affordances, no dark mode.
- **Visually dated.** New maintainers and contributors consistently flag the look as a barrier; it makes the project feel abandoned even when it isn't.
- **Hard to extend.** Every new feature has to fight the table-heavy fixed-grid layout to land cleanly.
- **Untestable end-to-end.** Markup has no stable hooks (no `data-testid`, sparse ARIA), state is encoded in styling rather than attributes, and the legacy interactions (xajax, MooTools holdovers, half-second transitions) are hostile to deterministic browser automation. There's currently no E2E coverage for the panel and the legacy theme is the reason adding it has been deferred.

Describe the solution you'd like
--------------------------------

A complete visual + interaction refresh, shipping as **v2.0.0** as a hard cutover. The legacy `web/themes/default/` template bundle is replaced wholesale; the theme system itself stays so admins who maintain a third-party theme (`web/themes/<name>/` + `config.theme=<name>`) are unaffected. **No per-user A/B toggle.** Rollback is `git checkout v1.x`.

**Cutover model: rename, not migrate.** Build the new theme alongside the old as `web/themes/sbpp2026/`. At the final cutover step, `git rm -r web/themes/default/` and `git mv web/themes/sbpp2026 web/themes/default`. Existing installs already have `config.theme=default`; after the rename, that value continues to point at a real directory whose contents are now the new theme. **No `sb_settings` migration needed.**

A finished design proposal exists from Claude Design and is the reference for the visual / interaction language:

- **[SourceBans++ Design System handoff (zip)](https://github.com/user-attachments/files/27382651/SourceBans%2B%2B.Design.System-handoff.zip)** — the design source of truth. Vendored to top-level `handoff/` at the start of A1, removed at D2.
[...]

PR #1134, opened 2026-05-05
[#1123 A3] Permission helper + View base extension + {has_access} Smarty plugin

rumblefrog:
Part of #1123 — Phase A3 of the v2.0.0 theme rollout. Sets up the foundation Phase B/C views consume to gate template content on web permissions.

Plan: [`sbpp2026-theme-rollout-4c702e97.plan.md`](/home/fishy/.cursor/plans/sbpp2026-theme-rollout-4c702e97.plan.md), section **A3 — Permission helper + `View` base extension for the new theme**.

## What this PR adds

1. **`Sbpp\View\Perms` static helper** — `Perms::for(?CUserManager $userbank): array` returns a flat `['can_<flag>' => bool, …]` snapshot of every `ADMIN_*` constant `init.php` defines from `web/configs/permissions/web.json`. Page handlers splat the result into a `<Page>View` named-arg constructor; PHP discards splatted keys the View doesn't declare, so each subclass opts in to only the booleans its template actually consumes. New flags added to the JSON automatically grow the array next request — no maintenance burden.
2. **`Sbpp\View\View` docblock convention** — extended the abstract base's docblock to spell out the splat-into-named-args pattern, and the rationale for keeping `can_*` properties on each concrete subclass (so `SmartyTemplateRule`'s parity check still catches dead permission references). No structural change to the class.
3. **`{has_access flag=…}…{/has_access}` Smarty 5 BLOCK plugin** — escape hatch in `SmartyCustomFunctions.php` for the rare per-row check the View couldn't precompute (e.g. inside a `{foreach}` over admins). Primary pattern stays `{if \$can_*}` on the View; this plugin is the deliberate exception.
4. **`PermsHelperTest` integration test** — pins the null-userbank, anonymous, owner-bypass, granted-flags-only, and key-set-coverage contracts.

## Files touched

Matches the plan's A3 \"Files\" list verbatim:

- **new** `web/includes/View/Perms.php`
- small edits to `web/includes/View/View.php` (docblock only — class shape unchanged)
- small edits to `web/includes/SmartyCustomFunctions.php` (one new function)

Plus one in-scope exception per the orchestrator brief:

- **new** `web/tests/integration/PermsHelperTest.php` — 5 cases / 165 assertions, paired with the new helper. Lives under `web/tests/integration/` because exercising `HasAccess()` against a real admin row requires the seeded `sb_admins` fixture (`Fixture::seedAdmin()` puts the admin's `extraflags = ADMIN_OWNER`); a unit-only test could only cover the null-userbank case.

Plus one one-line registration the plan flagged ahead of time:

- `web/init.php` — single new line `\$theme->registerPlugin(Smarty::PLUGIN_BLOCK, 'has_access', 'smarty_block_has_access');` next to the existing five `registerPlugin(...)` calls, mirroring how `help_icon`, `csrf_field`, etc. are wired. `SmartyCustomFunctions.php` does NOT expose a `register_smarty_custom_functions(Smarty)` extension point — every plugin in this repo is registered directly from `init.php` — so the only way to make `{has_access}` actually fire was to add the registration alongside its siblings. The plan's A3 \"Files\" list explicitly carved this case out: \"If you need to add a registerPlugin call, it's a one-liner that mirrors how existing plugins are registered. Document the change in the PR body.\"

## Design choice: docblock-only convention on the View base

I went with the **docblock-only** path on `View.php` instead of adding a trait or a `View::withPerms()` factory. Rationale:

- The plan offers a trait/factory option but also flags it as potentially over-engineered. PHP's named-args splat already does the work the factory would: `new BanListView(...Perms::for(\$userbank), ban_list: \$bans)` is two lines that any B-ticket can write without indirection.
- A trait that adds `can_*` properties on the base would silently defeat `SmartyTemplateRule`'s template-vs-View parity check — every concrete View would inherit booleans its template never reads, and dead permission checks would pile up across pages. Keeping the booleans on the concrete subclass forces the rule to flag drift on both directions (missing template ref → unused property; missing View prop → undeclared template var).
- A static factory (`SomeView::withPerms(\$userbank, …)`) buys nothing over the splat: the Renderer only reads public properties, the splat handles the wiring, and named-args make the call site read the same as without the helper. Adding a factory just means each subclass has to remember to override it.

So the surface area is: ONE helper class (`Perms`), ONE docblock paragraph on the base (`View.php`), and ONE plugin function (`smarty_block_has_access`). Phase B tickets get a 2-line consumer pattern.

## Sample usage (for B/C ticket authors)

```php
use Sbpp\View\Perms;
use Sbpp\View\Renderer;

global \$userbank, \$theme;

Renderer::render(\$theme, new BanListView(
    ...Perms::for(\$userbank),
    ban_list: \$bans,
    page: \$page,
    total: \$total,
));
```

The `BanListView` declares each `public readonly bool \$can_*` it needs (e.g. `\$can_add_ban`, `\$can_edit_all_bans`, `\$can_unban`, `\$can_owner`); PHP silently drops every other splatted key. The template then writes `{if \$can_add_ban}…{/if}` — no bitmask gymnastics, no `{if \$user.web_flags & …}`.

For the rare per-row gate inside a `{foreach}`:

```smarty
{foreach \$admins as \$a}
  …
  {has_access flag=\$smarty.const.ADMIN_EDIT_ADMINS}
    <a href=\"?p=admin&c=admins&o=edit&id={\$a.aid}\">Edit</a>
  {/has_access}
{/foreach}
```

## Owner-bypass semantic

`Perms::for()` ORs `ADMIN_OWNER` into every per-flag `HasAccess()` check, mirroring the convention baked into `CheckAdminAccess(ADMIN_OWNER|…)` in [`web/includes/page-builder.php`](web/includes/page-builder.php) and [`web/includes/system-functions.php`](web/includes/system-functions.php). A user holding `ADMIN_OWNER` therefore lights up every `can_*` => true; templates can write `{if \$can_add_ban}` without remembering to also check `\$can_owner`. The standalone `can_owner` key is still emitted for the rare panel that needs to gate strictly on owner status (e.g. \"transfer ownership\").

## Quality gates (all green locally)

- `./sbpp.sh phpstan` — default theme: OK
- `./sbpp.sh phpstan` (`SBPP_PHPSTAN_THEME=sbpp2026 + phpstan-sbpp2026.neon`) — sbpp2026 theme: OK
- `./sbpp.sh test` — full suite: 268 tests / 1134 assertions OK (incl. 5 new `PermsHelperTest` cases, 165 assertions)
- `./sbpp.sh ts-check` — clean
- `./sbpp.sh composer api-contract` — clean diff (no API surface change)

## Out of scope

- Templates are NOT updated to consume `\$can_*` or `{has_access}` here — that lands per-page in B1–B20.
- The `{has_access}` plugin's `flag=` parameter accepts SourceMod char-flag strings too (because `CUserManager::HasAccess()` already dispatches on `is_numeric()`); this PR doesn't introduce a separate `Sm\Perms` helper for SourceMod flags. If a B/C ticket needs one, that's a tiny follow-up.

### DIFF
--- a/web/includes/SmartyCustomFunctions.php
+++ b/web/includes/SmartyCustomFunctions.php
@@ -96,3 +96,62 @@ function smarty_function_csrf_field()
     $name = htmlspecialchars(CSRF::FIELD_NAME, ENT_QUOTES, 'UTF-8');
     return '<input type="hidden" name="' . $name . '" value="' . $token . '" />';
 }
+
+/**
+ * Smarty {has_access flag=…}…{/has_access} BLOCK plugin: render the inner
+ * content only when the current user holds the given permission flag.
+ *
+ * This is the deliberate ESCAPE HATCH for ad-hoc per-row checks (e.g.
+ * gating an action button inside a `{foreach}` over admins). The
+ * PRIMARY pattern stays precomputed `can_*` booleans on the View — see
+ * `Sbpp\View\Perms::for()` and the convention block on
+ * `Sbpp\View\View` — so most templates should keep using
+ * `{if $can_add_ban} … {/if}`.
+ *
+ * Usage examples (the `flag=` parameter accepts either a web-flag
+ * integer or a SourceMod char-flag string; `CUserManager::HasAccess()`
+ * dispatches on the type):
+ *
+ * ```smarty
+ * {has_access flag=$smarty.const.ADMIN_ADD_BAN}
+ *     <a href="?p=admin&c=bans&o=add">Add ban</a>
+ * {/has_access}
+ *
+ * {foreach $admins as $a}
+ *     {has_access flag=$smarty.const.ADMIN_EDIT_ADMINS}
+ *         <a href="?p=admin&c=admins&o=edit&id={$a.aid}">Edit</a>
+ *     {/has_access}
+ * {/foreach}
+ * ```
+ *
+ * Smarty 5 invokes block plugins twice per render — once on the
+ * opening tag (`$content === null`, `$repeat === true`) and once on
+ * the closing tag (`$content` populated, `$repeat === false`). We
+ * suppress output on the opening pass and gate on the closing pass.
+ *
+ * Reads `$userbank` from the request-global the rest of the panel
+ * exposes (set by `init.php`; same convention as every page handler's
+ * `global $userbank;`). When unset (no session, no user manager
+ * bound — should not happen in normal request flow), the plugin
+ * fails closed and emits nothing.
+ *
+ * @param array{flag?: int|string} $params
+ * @param string|null              $content
+ * @param mixed                    $template
+ * @param bool                     $repeat
+ */
+function smarty_block_has_access(array $params, ?string $content, $template, &$repeat): string
+{
+    if ($content === null) {
+        return '';
+    }
+    global $userbank;
+    if (!$userbank instanceof CUserManager) {
+        return '';
+    }
+    $flag = $params['flag'] ?? null;
+    if ($flag === null || $flag === '' || $flag === 0) {
+        return '';
+    }
+    return $userbank->HasAccess($flag) ? $content : '';
+}
--- a/web/includes/SmartyCustomFunctions.php
+++ b/web/includes/SmartyCustomFunctions.php
@@ -135,6 +135,17 @@ function smarty_function_csrf_field()
  * bound — should not happen in normal request flow), the plugin
  * fails closed and emits nothing.
  *
+ * Owner bypass (numeric web flags only): `ADMIN_OWNER` is OR'd into
+ * the mask before the `HasAccess()` call, mirroring `Sbpp\View\Perms::for()`
+ * and the `CheckAdminAccess(ADMIN_OWNER|…)` convention used throughout
+ * `web/includes/page-builder.php`. This guarantees the two helpers
+ * agree for owners — `{if $can_add_ban}` and
+ * `{has_access flag=$smarty.const.ADMIN_ADD_BAN}` evaluate identically
+ * regardless of which one the template author reaches for. Char-flag
+ * strings (SourceMod's `'a'`–`'z'`) are a separate permission system
+ * with their own root flag (`SM_ROOT='z'`) that `HasAccess()` already
+ * matches via substring scan; the OR is skipped on that path.
+ *
  * @param array{flag?: int|string} $params
  * @param string|null              $content
  * @param mixed                    $template
@@ -152,6 +163,9 @@ function smarty_block_has_access(array $params, ?string $content, $template, &$r
     $flag = $params['flag'] ?? null;
     if ($flag === null || $flag === '' || $flag === 0) {
         return '';
+    }
+    if (is_numeric($flag) && defined('ADMIN_OWNER')) {
+        $flag = (int) $flag | (int) constant('ADMIN_OWNER');
     }
     return $userbank->HasAccess($flag) ? $content : '';
 }
--- a/web/includes/View/Perms.php
+++ b/web/includes/View/Perms.php
@@ -1,0 +1,121 @@
+<?php
+declare(strict_types=1);
+
+namespace Sbpp\View;
+
+/**
+ * Snapshot the current user's web-permission flags as a flat
+ * `can_*` => bool array suitable for splatting into a {@see View}
+ * subclass's named constructor.
+ *
+ * The convention each `Sbpp\View\<Page>View` follows:
+ *
+ *   1. Each permission the template needs is a constructor-promoted
+ *      `public readonly bool $can_<flag>` property (e.g.
+ *      `$can_add_ban`, `$can_edit_all_bans`, `$can_owner`).
+ *   2. The page handler builds the View like:
+ *
+ *      ```php
+ *      use Sbpp\View\Perms;
+ *      use Sbpp\View\Renderer;
+ *
+ *      Renderer::render($theme, new BanListView(
+ *          ...Perms::for($userbank),
+ *          ban_list: $bans,
+ *          // …
+ *      ));
+ *      ```
+ *
+ *      PHP discards splatted `can_*` keys the View doesn't declare, so
+ *      a View only opts in to the booleans its template actually
+ *      consumes — `SmartyTemplateRule` keeps that opt-in honest by
+ *      flagging unused properties.
+ *   3. Templates render `{if $can_add_ban} … {/if}` — never raw
+ *      `{if $user.web_flags & …}` — and never need to know the
+ *      bitmask values.
+ *
+ * The boolean set is derived at runtime by enumerating every
+ * `ADMIN_*` constant defined by `init.php` from
+ * `web/configs/permissions/web.json`. Adding a new permission flag in
+ * that JSON automatically grows the array next request — no
+ * maintenance burden here.
+ *
+ * Naming convention: each constant `ADMIN_FOO_BAR` becomes the array
+ * key `can_foo_bar`. The `ADMIN_` prefix is stripped and the rest is
+ * lowercased, matching the `{if $can_foo_bar}` style the templates use.
+ *
+ * For the rare case a template needs an ad-hoc check the View couldn't
+ * precompute (e.g. inside a `{foreach}` over admins, gating one column
+ * per row), use the paired `{has_access flag=$smarty.const.ADMIN_FOO}`
+ * Smarty block plugin in `SmartyCustomFunctions.php`.
+ */
+final class Perms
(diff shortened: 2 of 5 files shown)


rumblefrog (commented):
## Initial review by Reviewer subagent

I'm a separate agent from the Worker who wrote this PR; I read the diff fresh against the A3 plan and AGENTS.md. **0 blocking, 1 non-blocking finding.** Verification audit below.

### Verification audit (each item independently verified, not taken on faith)

**1. Scope boundary — ✓ within A3 scope.**
Diff touches exactly: `web/includes/View/Perms.php` (new), `web/includes/View/View.php` (docblock only), `web/includes/SmartyCustomFunctions.php` (one new function), `web/init.php` (one new line), `web/tests/integration/PermsHelperTest.php` (new). The first three match the plan's A3 \"Files\" list verbatim; `init.php` is the explicitly authorized one-liner; the integration test is allowed by the dispatch's \"(Optional but recommended) A unit test for `Perms::for()`\".

**2. AGENTS anti-patterns — none reintroduced.**
No xajax/sb-callback/ADOdb/MooTools/jQuery/React/string-literal-actions/inlined-prefix in production code/HTML-decode-on-JSON/DB_CHARSET regression/WYSIWYG/data.sql or struc.sql edits/unannotated nofilter. (The test does `sprintf('INSERT INTO \`%s_admins\` …', DB_PREFIX)` via `Fixture::rawPdo()`, but that mirrors `Fixture::seedAdmin()`'s existing pattern at `web/tests/Fixture.php:113-122` — raw PDO bypasses `Database::query()`'s `:prefix_` rewrite, so this is the test-side convention, not a production anti-pattern.)

**3. View DTO compliance — ✓.**
`web/includes/View/View.php` change is purely a docblock paragraph (lines 23-53). The abstract class shape, `TEMPLATE` constant, `DELIMITERS` constant, and lack of constructor are all unchanged. `SmartyTemplateRule` introspects concrete subclasses, so it's unaffected by base-class doc additions.

**4. Permission checks — ✓ correct.**
- Naming: `ADMIN_FOO_BAR` → `can_foo_bar` via `strtolower(substr($name, strlen('ADMIN_')))` at `web/includes/View/Perms.php:89`. Spot-checked against `web/configs/permissions/web.json`: `ADMIN_ADD_BAN` (256) → `can_add_ban`, `ADMIN_OWNER` (16777216) → `can_owner`, `ADMIN_EDIT_OWN_BANS` → `can_edit_own_bans`. ✓
- Null userbank: `Perms::for(null)` returns all-`false` because the loop short-circuits on `$userbank !== null` (line 91). ✓
- Auto-discovery: `adminFlags()` enumerates `get_defined_constants(true)['user']` and filters `ADMIN_*` (lines 105-119). Because `init.php:178-181` uses `define()` to populate them from `web/configs/permissions/web.json`, they land in the `user` bucket. New JSON entries auto-flow — no double-source-of-truth maintenance burden. ✓
- `ALL_WEB` (also defined from web.json) and `SM_*` (sourcemod.json) are correctly excluded by the `ADMIN_` prefix filter.

**`HasAccess()` ADMIN_OWNER VERIFICATION (orchestrator-mandated):** I read `web/includes/CUserManager.php:109-135`. The method does **NOT** auto-OR `ADMIN_OWNER`:
\`\`\`php
if (is_numeric(\$flags)) {
    return ((int)\$this->admins[\$aid]['extraflags'] & (int)\$flags) != 0 ? true : false;
}
\`\`\`
It's a raw bitmask AND. So `Perms::for()`'s `\$mask = \$value | \$owner` at `Perms.php:90` is the **correct** convention mirror — it matches the explicit `CheckAdminAccess(ADMIN_OWNER|…)` pattern used in `web/includes/page-builder.php:51,59,73,85,95,103,111` and elsewhere. The Worker's claim is verified.

**5–10.** N/A (no JS, no JSON API handler, no API contract impact, no snapshots, no string-literal actions, no CSRF surface).

**11. PHPStan baseline NOT touched.** `git diff origin/main..HEAD -- web/phpstan-baseline.neon` is empty.

**12. CI status — all 4 gates green.** `Static analysis (theme=default)`, `Static analysis (theme=sbpp2026)`, `PHPUnit` (268 tests / 1134 assertions), `tsc --checkJs`. Confirmed via `gh pr checks`.

**13. Behavioral correctness:**
- `Perms.php` — static class (private constructor at L57), single public static `for(?CUserManager): array` method, correct return shape `array<string, bool>`. ✓
- `View.php` — docblock only, no shape change. ✓
- `smarty_block_has_access` — Smarty 5 block plugin signature `(array, ?string, $template, &$repeat): string` is correct; opening pass returns empty (\$content === null branch at L145); fail-closed on unbound userbank (L149) and missing/zero flag (L153). Edge case: passing `flag='0'` (string) bypasses the `=== 0` check, but `HasAccess()` then short-circuits via `empty(\$flags)` which handles `'0'` — so still fail-closed. ✓
- `init.php:223` — literally one line, placed contiguously with the existing four `registerPlugin(Smarty::PLUGIN_FUNCTION, …)` calls. Mirrors existing registration style verbatim. ✓
- `PermsHelperTest` — 5 cases / 165 assertions cover: null userbank → all false; logged-out admin → all false; auto-discovery key set; owner sees all true; non-owner with narrow mask sees only granted flags. ✓

### Findings

**F1 — Non-blocking — Owner-bypass inconsistency between `Perms::for()` and `{has_access}` block plugin.**

`Perms::for()` (`web/includes/View/Perms.php:90`) ORs `ADMIN_OWNER` into every per-flag mask before calling `HasAccess()`, so an owner-only user lights up every `can_*` => true (verified by `testOwnerSeesEveryCanFlag` at `web/tests/integration/PermsHelperTest.php:129-140`).

`smarty_block_has_access()` (`web/includes/SmartyCustomFunctions.php:156`) calls `\$userbank->HasAccess(\$flag)` raw — no OR. Concrete consequence: an owner-only user evaluates `{if \$can_add_ban}` as **true** but the equivalent `{has_access flag=\$smarty.const.ADMIN_ADD_BAN}` as **false**. Two ways to express the same gate produce different results for owners.

This is the documented \"escape hatch\" so a defensible argument exists for keeping it raw — but the inconsistency is a footgun for B/C ticket authors who'll reach for `{has_access}` expecting parity with the View-side helper. Recommend OR'ing `ADMIN_OWNER` into the plugin's mask to match `Perms::for()`'s contract. Will fix in this review pass.

Addressing now.

rumblefrog:
**F1 — Fixed in ac6934fe.**

Mirrored `Perms::for()`'s owner-bypass in `smarty_block_has_access`: OR `ADMIN_OWNER` into the mask for numeric web-flag checks before calling `HasAccess()`. SourceMod char-flag strings (`'a'`–`'z'`) keep raw semantics — that's a separate permission system with its own root flag (`SM_ROOT='z'`) which `HasAccess()` already substring-matches.

Local gates after the fix:
- PHPStan default theme: OK
- PHPStan sbpp2026 theme: OK
- PHPUnit (incl. existing 5 PermsHelperTest cases): 268 tests / 1134 assertions OK
- ts-check: clean

Watching CI.

rumblefrog (commented):
## Final review summary

- **Files in scope:** verified against plan A3's \"Files\" list — `web/includes/View/Perms.php` (new), `web/includes/View/View.php` (docblock only), `web/includes/SmartyCustomFunctions.php` (one new function + ac6934fe owner-bypass fix), `web/init.php` (single authorized one-line `registerPlugin`), `web/tests/integration/PermsHelperTest.php` (new — paired test).
- **AGENTS anti-patterns:** none.
- **`Perms::for()` correctness:** verified naming convention (`ADMIN_FOO_BAR` → `can_foo_bar`), null-userbank → all-false handling, ADMIN_OWNER OR-ing semantics (independently confirmed against `CUserManager::HasAccess()` and the `CheckAdminAccess(ADMIN_OWNER|…)` convention in `page-builder.php`).
- **`{has_access}` block plugin:** verified Smarty 5 block plugin signature, fail-closed paths (unbound `\$userbank`, missing/zero/empty `flag=`); owner-bypass parity with `Perms::for()` added in ac6934fe (F1 fix).
- **`View.php` extension:** docblock-only, no structural change — `SmartyTemplateRule` introspection unaffected.
- **`init.php` one-liner:** verified literally one new line at L223, placed contiguously with the existing four `registerPlugin` calls.
- **PHPStan baseline:** not modified.
- **API contract:** not affected (no JSON handler changes).
- **CI:** all 4 gates green after the fix (PHPStan default, PHPStan sbpp2026, PHPUnit 268/1134, ts-check).
- **Findings:** 0 blocking, 1 non-blocking — F1 fixed in ac6934fe (no findings declined).

Self-approval blocked by GitHub; merging via squash.

PR #1136, opened 2026-05-05
[#1123 B7] Redesign public protest form

rumblefrog:
## Summary

Phase B7 of #1123: replace the A1 sbpp2026 stub for the public ban-appeal page (`page_protestban.tpl`) with a real redesign and route the existing page handler (`web/pages/page.protest.php`) through a typed `Sbpp\View\ProtestBanView` instead of the legacy `$theme->assign(...)` chain (and the trailing MooTools-flavored `<script>` block that depended on the legacy theme's row IDs).

- **`web/themes/sbpp2026/page_protestban.tpl`** — full redesign derived from the handoff submit + login two-column shell pattern: a primary form card sits next to a slim info aside with the "what happens next" copy and the threat-warning blurb the legacy page already shipped. Vanilla-JS Steam/IP row toggle, `{csrf_field}`, every input carries `data-testid="protest-…"` per the issue's testability rule.
- **`web/includes/View/ProtestBanView.php`** — new typed view declaring the five form-state properties (`steam_id`, `ip`, `player_name`, `reason`, `player_email`). Names mirror the legacy `default/page_protestban.tpl` one-for-one so the dual-theme PHPStan matrix added in #1123 A2 stays clean on both legs until D1 collapses it. Initial Steam-vs-IP row visibility on a re-render is derived from `$ip` being non-empty (the only state the legacy POST validator can leave behind after a failed IP-typed submit) so we don't have to introduce a `$type` property the legacy template doesn't reference.
- **`web/pages/page.protest.php`** — the trailing five `$theme->assign(...)` calls + `$theme->display(...)` + raw-PHP `<script>` block collapse to a single `Renderer::render($theme, new ProtestBanView(...))`. Validation, DB writes, and the email notification flow are untouched. The deleted `<script>` referenced `$('Type').options[<?=$Type?>]` against legacy IDs that no longer exist in the redesigned template; the inline JS in the new template owns the toggle now.

Form `name=` attributes (`Type`, `SteamID`, `IP`, `PlayerName`, `EmailAddr`, `BanReason`, `subprotest`) are frozen end-to-end so the page handler's `$_POST` reads keep working unchanged on both themes. `Sbpp\View\Perms::for(...)` is intentionally NOT splatted — the public protest page renders for unauthenticated visitors and the template gates nothing on permission flags.

## Files touched

- `web/includes/View/ProtestBanView.php` *(new)*
- `web/pages/page.protest.php`
- `web/themes/sbpp2026/page_protestban.tpl`

## Local gates

| Gate | Command | Result |
|------|---------|--------|
| PHPStan (default theme) | `./sbpp.sh phpstan` | OK, no errors |
| PHPStan (sbpp2026 theme) | `SBPP_PHPSTAN_THEME=sbpp2026 ... --configuration=phpstan-sbpp2026.neon` | OK, no errors |
| PHPUnit | `./sbpp.sh test` | OK (268 tests, 1134 assertions) |
| ts-check | `./sbpp.sh ts-check` | OK |
| API contract | `./sbpp.sh composer api-contract` | unchanged |

## Smoke tests (in dev stack)

- `GET /index.php?p=protest` (theme = sbpp2026) renders inside the new chrome with sidebar + topbar, a card-based form on the left, and the info aside on the right. Steam ID type is selected by default and the IP row carries `hidden`.
- `POST /index.php?p=protest` with `Type=1` and an invalid IP re-renders with `option value="1" selected`, the `data-protest-row="steam"` div hidden, and the typed IP value preserved in the input — confirming the `$ip`-derived initial-row inference.
- `GET /index.php?p=protest` (theme = default) still renders the legacy `default/page_protestban.tpl` markup unchanged, since `ProtestBanView`'s property names match the legacy template variables one-for-one.

## Test plan

- [ ] CI: PHPStan default + sbpp2026 legs green
- [ ] CI: PHPUnit green
- [ ] CI: ts-check green
- [ ] CI: api-contract green (no diff)
- [ ] Manual: file an appeal as an unauthenticated visitor on the sbpp2026 theme and confirm the existing email-notification + DB insert flow still fires
- [ ] Manual: switch to the legacy `default` theme and confirm the form still renders and submits

### DIFF
--- a/web/includes/View/ProtestBanView.php
+++ b/web/includes/View/ProtestBanView.php
@@ -1,0 +1,37 @@
+<?php
+declare(strict_types=1);
+
+namespace Sbpp\View;
+
+/**
+ * Public ban appeal ("protest a ban") form — binds to
+ * `page_protestban.tpl`.
+ *
+ * The page is rendered for both the empty-form and post-submission
+ * re-render cases. On a re-render after validation failure the legacy
+ * page handler echoes a {@see ShowBox()} `<script>` ahead of the
+ * template; the user-supplied values are seeded back into the form
+ * via these properties so the user does not have to retype them.
+ *
+ * Variable names mirror the legacy `web/themes/default/page_protestban.tpl`
+ * one-for-one so the dual-theme PHPStan matrix added in #1123 A2 stays
+ * happy on both the default leg and the sbpp2026 leg until D1 collapses
+ * the matrix. The ban-context "type" (Steam vs IP) is intentionally NOT
+ * a separate property: the new template derives the initially-visible
+ * row from whether `$ip` is non-empty, which matches the only path that
+ * can leave `$ip` populated after a failed submit (the user picked
+ * "IP Address" as the ban type).
+ */
+final class ProtestBanView extends View
+{
+    public const TEMPLATE = 'page_protestban.tpl';
+
+    public function __construct(
+        public readonly string $steam_id,
+        public readonly string $ip,
+        public readonly string $player_name,
+        public readonly string $reason,
+        public readonly string $player_email,
+    ) {
+    }
+}
--- a/web/pages/page.protest.php
+++ b/web/pages/page.protest.php
@@ -174,20 +174,10 @@ if (!isset($_POST['subprotest']) || $_POST['subprotest'] != 1) {
     }
 }
 
-$theme->assign('steam_id', $SteamID);
-$theme->assign('ip', $IP);
-$theme->assign('player_name', $PlayerName);
-$theme->assign('reason', $UnbanReason);
-$theme->assign('player_email', $Email);
-
-$theme->display('page_protestban.tpl');
-?>
-<script type="text/javascript">
-function changeType(szListValue)
-{
-    $('steam.row').style.display = (szListValue == "0" ? "" : "none");
-    $('ip.row').style.display    = (szListValue == "1" ? "" : "none");
-}
-$('Type').options[<?=$Type;?>].selected = true;
-changeType(<?=$Type?>);
-</script>
+\Sbpp\View\Renderer::render($theme, new \Sbpp\View\ProtestBanView(
+    steam_id: (string) $SteamID,
+    ip: (string) $IP,
+    player_name: (string) $PlayerName,
+    reason: (string) $UnbanReason,
+    player_email: (string) $Email,
+));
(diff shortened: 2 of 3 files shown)


rumblefrog (commented):
Initial review. 0 blocking, 1 non-blocking; addressing now.

Reviewed against the B7 checklist on the actual PR diff at e9f0078e:

- **Scope ✓** — the three files match (handler is `page.protest.php`, plan path was a typo).
- **Anti-patterns ✓** — vanilla-JS IIFE, no MooTools/jQuery; legacy `$('Type').options[…]` block correctly deleted; no xajax/ADOdb regressions.
- **View DTO ✓** — `Sbpp\View\ProtestBanView` extends `View`, declares the five form-state props, paired with `page_protestban.tpl`. PHPStan dual-theme matrix passes both legs (so default leg's `{$steam_id}…{$player_email}` parity holds).
- **Permissions ✓** — public page; intentionally no `Perms::for(...)` splat, matches View base class doc.
- **Testability hooks ✓** — `protest-name`, `protest-email`, `protest-reason`, `protest-steam`, `protest-ip`, `protest-submit` all present; bonus `protest-form`, `protest-type` for state assertions; row visibility uses the HTML `hidden` attribute, not `style.display`, so tests can assert state without computed CSS.
- **CSRF ✓** — `{csrf_field}` on the form (the `validate(...)` call on the handler side is pre-existing scope and unchanged).
- **Inline JS ✓** — small, vanilla, IIFE, attribute-selector based, runs after DOM, no Smarty `{$…}` interpolation; relies on Smarty 5 auto_literal so `{ return; }` etc. parse fine. Confirmed not in the ts-check scope (`web/scripts/` only).
- **Behavioural toggle ✓** — `sync()` mirrors `hidden` and `required` together, so the hidden field is never `required`; both initial render (`$ip_first` derived) and `change` event keep the two in lockstep.

**Non-blocking observation (#1, not requesting a fix):** the `{assign var="ip_first" value=($ip != '')}` inference correctly identifies "user picked IP" only when `$ip` arrived non-empty. The asymmetric edge case is *Type=1 with an empty IP* (user opens form, picks IP, submits without filling it): IP validation fires "Please type a valid IP", but the re-render bumps the dropdown back to Steam mode because `$ip_first = false`. Recoverable (re-select IP) and the worker's docblock explicitly chose this trade-off to avoid declaring a `$type` property the legacy template doesn't reference. Safe to defer to D1 when the matrix collapses; flagging only so it's logged.

CI: PHPStan default + sbpp2026 + PHPUnit + ts-check all green. Mergeable, no conflicts. Approving and merging.

PR #1137, opened 2026-05-05
[#1123 B19] Add admin audit log page

rumblefrog:
Part of #1123 — Phase B, ticket B19 (NEW audit log page).

## Summary

Adds a dedicated `?p=admin&c=audit` page that surfaces the existing
`:prefix_log` table through a typed View DTO + handoff-styled template.
**No schema change** — the `:prefix_log` writer side (severity letters
`m`/`w`/`e`) is untouched.

* Owner-restricted (`CheckAdminAccess(ADMIN_OWNER)`): no
  `ADMIN_LIST_LOGS` / `ADMIN_LIST_AUDIT` flag exists in
  `web/configs/permissions/web.json`. The legacy
  `TRUNCATE :prefix_log` path in `admin.settings.php` is also
  owner-only, so the gate matches existing posture.
* **SSR**, no JSON action. Pagination + severity chips + free-text
  search are all driven by GET params. See "Plan deviations" below.
* Variable contract is the handoff-style row shape from the worker
  brief: `{tid, severity, severity_label, severity_class, time_human,
  time_iso, actor, title, detail, ip}`. `tid` ← `lid`,
  `severity_*` ← `type` enum, `actor` ← `admins.user` join, `time_*`
  ← `created` (UNIX ts), `detail` ← `message`, `ip` ← `host`.
* Severity badges render `Info` / `Warning` / `Error` with the
  expected `data-severity="info|warning|error"` test hook.
* Two `<form>`s (search vs severity) so pressing Enter inside the
  search input doesn't reset the active chip — the multi-submit-button
  trap of a single combined form.

## Files touched

| File | Why |
| ---- | --- |
| `web/includes/page-builder.php` | New `case 'audit':` arm under `case 'admin':`, gated on `CheckAdminAccess(ADMIN_OWNER)`. |
| `web/includes/View/AuditLogView.php` | New typed View DTO bound to `page_admin_audit.tpl`. |
| `web/pages/admin.audit.php` | New SSR page handler. Reads `:prefix_log` via the global `Database`, normalises rows, renders via `Sbpp\View\Renderer::render`. |
| `web/themes/sbpp2026/page_admin_audit.tpl` | New redesigned template carrying every test hook the issue's "Testability hooks" rule asks for. |
| `web/themes/default/page_admin_audit.tpl` | Minimal stub (carries the `STUB_MARKER` phrase) so the dual-theme PHPStan matrix's "template file must exist" check passes for the legacy theme too. Deleted at D1 alongside the rest of `themes/default/`. |

## Local quality gates

| Gate | Result |
| ---- | ------ |
| `./sbpp.sh phpstan` (default theme) | OK (175 files, no errors) |
| `./sbpp.sh phpstan` with `SBPP_PHPSTAN_THEME=sbpp2026` (`phpstan-sbpp2026.neon`) | OK |
| `./sbpp.sh test` (full PHPUnit) | OK — 268 tests, 1134 assertions |
| `./sbpp.sh ts-check` | OK |
| `./sbpp.sh composer api-contract` | wrote 58 actions / 32 perms / 2 typedefs — **no `git diff`** (SSR path adds no new action). |

## Plan deviations

**SSR vs `audit.list` JSON action.** The worker brief allowed B19 to
add a new `audit.list` handler in `web/api/handlers/system.php` (the
one Phase-B exception that's allowed to touch `api/handlers/*.php`).
I went **SSR**:

* The audit page is admin-only and low traffic; async pagination over
  GET-param SSR is mostly cosmetic.
* The handoff design (`handoff/pages/admin/audit.tpl`) is form-driven,
  not async-driven.
* SSR keeps the diff inside the smaller of the two B19 envelopes — no
  `web/api/handlers/system.php` change, no `_register.php` change, no
  `api-contract.js` regen, no `PermissionMatrixTest` row, no
  `__snapshots__/system/audit_list.json`. That's lower conflict risk
  with the parallel B-tickets touching adjacent paths.

If a future PR wants to add async pagination it's an additive change:
the handler already returns the same shape the JSON action would, so
the diff there would be a thin wrapper.

## Sidebar entry follow-up (out of B19 scope)

A2's `web/pages/core/navbar.php` does not yet emit an "Audit" entry
into the `$adminbar` array, so the new page is reachable only by
direct URL (or by typing `?p=admin&c=audit`). Per the worker brief,
B19 must **not** touch `web/themes/sbpp2026/core/navbar.tpl` or
`web/pages/core/navbar.php`.

Follow-up: add an entry like:

```php
[
    'title'      => 'Audit Log',
    'endpoint'   => 'audit',
    'permission' => ADMIN_OWNER,
],
```

…to the `$admin` array in `web/pages/core/navbar.php`, and an
`{elseif $admin.endpoint == 'audit'}clipboard-list{...}` arm in
`navbar.tpl`'s lucide chain. Either both at once or the navbar.tpl
arm can be deferred (the chain falls through to a `circle` icon).

## Other deviations from the brief

None. All hard rules satisfied:

* Touched only files in scope (the legacy stub at
  `web/themes/default/page_admin_audit.tpl` is required by the
  dual-theme PHPStan matrix; without it the `default/` leg of CI's
  phpstan job fails with `sbpp.view.templateMissing`).
* Permissions precomputed at handler entry — no inline
  `{if $userbank->HasAccess(...)}` in the template; the page handler
  doesn't need a `Perms::for` splat because the page has no
  action-gated UI (read-only view, owner-only entry).
* Every `nofilter` annotated — actually, **the template uses zero
  `nofilter`**: every `{$foo}` runs through Smarty's global
  auto-escape. The template-header docblock calls this out so a future
  reader doesn't add an unannotated one.
* Vanilla SSR; no `audit.js`, no Mootools, no jQuery, no fetch().
* All SQL uses `:prefix_` placeholders.
* Test hooks: `data-testid="audit-row"` + `data-id`, `data-severity`,
  `filter-chip-{all,info,warning,error}`, `audit-search`, `page-prev`,
  `page-next` — all present and verified via `curl + grep`.
* `SmartyTemplateRule` parity check passes: every public property on
  `AuditLogView` is referenced by the template, and vice versa.
* No reintroduced anti-patterns (xajax, sb-callback, ADOdb,
  htmlspecialchars_decode, utf8 charset, TinyMCE / CKEditor).

## Test plan

* [ ] Reviewer pulls, switches active theme to `sbpp2026`, navigates to
      `/index.php?p=admin&c=audit` as the seeded `admin` user.
* [ ] Confirm severity chips filter rows and the active chip shows
      `aria-pressed="true"`.
* [ ] Confirm search input filters by `title` / `message` substring.
* [ ] Confirm `page-prev` / `page-next` advance through the list and
      the disabled state matches the boundary.
* [ ] Confirm a non-owner admin (any flag set, but not `ADMIN_OWNER`)
      gets redirected by `CheckAdminAccess` and never reaches the
      handler.
* [ ] Confirm `?theme=default` (or any `config.theme` other than
      `sbpp2026`) renders the legacy stub with the
      "hasn't been redesigned yet" copy.

### DIFF
--- a/web/includes/View/AuditLogView.php
+++ b/web/includes/View/AuditLogView.php
@@ -1,0 +1,57 @@
+<?php
+declare(strict_types=1);
+
+namespace Sbpp\View;
+
+/**
+ * Admin → Audit log page (#1123 B19).
+ *
+ * Binds to `page_admin_audit.tpl`. New page, no legacy template — backed
+ * by the existing `:prefix_log` table (no schema change). Each
+ * `audit_log` row is normalised by `web/pages/admin.audit.php` into the
+ * handoff-style shape the template iterates over: `tid` (PK from `lid`),
+ * `severity` (one of `m`/`w`/`e`, mapped to `severity_label` +
+ * `severity_class` for the badge), `time_human` + `time_iso` (formatted
+ * from `created`), `actor` (admin name from `aid` join, or `"system"`
+ * when the row was emitted by an unauthenticated path), `title`,
+ * `detail` (from `message`), and `ip` (from `host`).
+ *
+ * SSR pagination + filtering: `current_severity`, `search`,
+ * `current_page`, `page_count`, plus `prev_url` / `next_url` /
+ * `has_prev` / `has_next` for the pager. We chose SSR over an
+ * `audit.list` JSON action because the page is admin-only and low
+ * traffic; the form-based filter chips + GET pagination match the
+ * handoff design without adding a JS dependency.
+ */
+final class AuditLogView extends View
+{
+    public const TEMPLATE = 'page_admin_audit.tpl';
+
+    /**
+     * @param list<array{
+     *     tid: int,
+     *     severity: string,
+     *     severity_label: string,
+     *     severity_class: string,
+     *     time_human: string,
+     *     time_iso: string,
+     *     actor: string,
+     *     title: string,
+     *     detail: string,
+     *     ip: string,
+     * }> $audit_log
+     */
+    public function __construct(
+        public readonly array $audit_log,
+        public readonly int $total_count,
+        public readonly string $current_severity,
+        public readonly string $search,
+        public readonly int $current_page,
+        public readonly int $page_count,
+        public readonly bool $has_prev,
+        public readonly bool $has_next,
+        public readonly string $prev_url,
+        public readonly string $next_url,
+    ) {
+    }
+}
--- a/web/includes/page-builder.php
+++ b/web/includes/page-builder.php
@@ -110,6 +110,14 @@ function route($fallback)
                 case 'settings':
                     CheckAdminAccess(ADMIN_OWNER|ADMIN_WEB_SETTINGS);
                     return ['SourceBans++ Settings', '/admin.settings.php'];
+                case 'audit':
+                    // Audit log is owner-restricted: there is no dedicated
+                    // ADMIN_LIST_LOGS / ADMIN_LIST_AUDIT flag in
+                    // web/configs/permissions/web.json (#1123 B19), and the
+                    // legacy `TRUNCATE :prefix_log` path in
+                    // admin.settings.php is also gated on ADMIN_OWNER.
+                    CheckAdminAccess(ADMIN_OWNER);
+                    return ['Audit Log', '/admin.audit.php'];
                 default:
                     CheckAdminAccess(ALL_WEB);
                     return ['Administration', '/page.admin.php'];
--- a/web/pages/admin.audit.php
+++ b/web/pages/admin.audit.php
@@ -1,0 +1,155 @@
+<?php
+/*************************************************************************
+This file is part of SourceBans++
+
+SourceBans++ (c) 2014-2024 by SourceBans++ Dev Team
+
+The SourceBans++ Web panel is licensed under a
+Creative Commons Attribution-NonCommercial-ShareAlike 3.0 Unported License.
+
+You should have received a copy of the license along with this
+work.  If not, see <http://creativecommons.org/licenses/by-nc-sa/3.0/>.
+*************************************************************************/
+
+// Admin → Audit log handler (#1123 B19). Auth is enforced upstream by
+// the `CheckAdminAccess(ADMIN_OWNER)` gate the matching `case 'audit':`
+// arm of `web/includes/page-builder.php` runs before dispatching here,
+// matching every other admin.*.php in this directory.
+//
+// SSR: severity chips + free-text search + numeric pager are all
+// driven by GET params, no JSON action. Variable shape mirrors the
+// handoff design (handoff/pages/admin/audit.tpl) but pivots filters
+// onto the `:prefix_log.type` enum (`m`/`w`/`e`) since the table only
+// stores severity, not the action-kind taxonomy the mockup illustrates.
+
+if (!defined("IN_SB")) {
+    echo "You should not be here. Only follow links!";
+    die();
+}
+global $userbank, $theme;
+
+$AuditPerPage = SB_BANS_PER_PAGE;
+
+$page = 1;
+if (isset($_GET['page']) && (int) $_GET['page'] > 0) {
+    $page = (int) $_GET['page'];
+}
+
+$severity = '';
+if (isset($_GET['severity']) && in_array($_GET['severity'], ['m', 'w', 'e'], true)) {
+    $severity = (string) $_GET['severity'];
+}
+
+$search = '';
+if (isset($_GET['search']) && is_string($_GET['search'])) {
+    // Cap search input to a sane length so the page can't be wedged by a
+    // pathological query string. The LIKE clause below adds the wildcards.
+    $search = mb_substr(trim((string) $_GET['search']), 0, 200);
+}
+
+$where = [];
+$params = [];
+if ($severity !== '') {
+    $where[] = 'l.type = :severity';
+    $params[':severity'] = $severity;
+}
+if ($search !== '') {
+    $where[] = '(l.title LIKE :search OR l.message LIKE :search)';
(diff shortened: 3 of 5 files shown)


rumblefrog:
Follow-up from a second worker pass (subagent for B19): pushed `dc7f06a1` to fix the `Static analysis (theme=sbpp2026)` CI failure.

## What broke

The header `{*...*}` block in `web/themes/sbpp2026/page_admin_audit.tpl` documented the row shape and Smarty syntax with literal examples — `{$foo}`, `{$foo nofilter}`, `{tid, severity, ...}`, `filter-chip-{kind}`, `data-id="{tid}"`. `SmartyTemplateRule` (web/includes/PHPStan/SmartyTemplateRule.php) scans templates with a single non-greedy `\{[^\}]*?\}` regex and doesn't strip Smarty comments before scanning. The embedded `}` characters inside the comment ended the comment match early, and the next `{$foo}` was harvested as a real `$foo` variable reference. CI failed with:

```
Template "page_admin_audit.tpl" references {$foo} but Sbpp\View\AuditLogView has no public $foo property.
```

## Fix

Rewrote the docblock prose-only — same content, no literal `{...}` patterns. No template logic / page handler / View change. The body comment around the search/severity forms (the "two `<form>`s" rationale) doesn't contain `{...}`, so it's untouched.

A more durable fix would be to teach `SmartyTemplateRule::parseTemplate` to strip `{*...*}` blocks (alongside the existing `{literal}` strip) before regex parsing, but that lives outside B19's scope (the rule was added in A3 / refined in A2 — touching it here would put it in the "out of scope" bucket).

## CI

All four gates green on `dc7f06a1`:

* **Static analysis (theme=default)** — pass (59s)
* **Static analysis (theme=sbpp2026)** — pass (39s) ← previously failing
* **PHPUnit** — pass (1m3s)
* **tsc --checkJs** — pass (9s)

## Local rendering check (no browser available for a screenshot)

Logged in as the seeded `admin` and curled `?p=admin&c=audit` against the worktree's docker stack (`http://localhost:8410`). The page returns 200 with the expected layout:

```
30 audit-row articles
26 data-severity="error"
14 data-severity="info"
20 data-severity="warning"
test hooks present: audit-search, filter-chip-{all,info,warning,error}, page-prev, page-next
badges: 13× badge--error, 10× badge--warning, 7× badge--info
pager: Previous disabled, Next -> ?p=admin&c=audit&page=2
```

(Severity `data-severity` count is higher than badge count because `<article>` carries it too.)

Worker context: `WORKER_ID=B19`, `BRANCH=feat/1123-b19-audit-log`, `AUDIT_LIST_ADDED=no` (kept the SSR path — same call as the prior commit).

rumblefrog (commented):
Initial review — 0 blocking, 1 non-blocking; no requested changes.

**Scope.** Matches B19 envelope. Two documented out-of-scope edits accepted: the `case 'audit':` arm in `web/includes/page-builder.php` (one-liner + permission check, structurally identical to the sibling arms) and the legacy `web/themes/default/page_admin_audit.tpl` stub (carries the `STUB_MARKER` phrase so `SmartyTemplateRule::isStubTemplate` short-circuits the dual-theme parity check; deleted at D1).

**Anti-patterns.** None. No xajax / ADOdb / MooTools / `:prefix_` inlining / `htmlspecialchars_decode` on params / `utf8` charset / WYSIWYG / unannotated `nofilter` (the template uses zero `nofilter`, which is the cleanest outcome).

**View DTO + SmartyTemplateRule.** `AuditLogView` declares all 10 template vars as `public readonly` properties; the template references each one and no extras. Stub-skip handled via `STUB_MARKER` for the legacy theme.

**Test hooks.** All present and verified against the issue body: `data-testid="audit-row"` + `data-id` + `data-severity`, `filter-chip-{all,info,warning,error}`, `audit-search`, `page-prev` / `page-next`.

**Variable naming.** `tid`, `severity` (m/w/e), `time_human`, `actor`, `title`, `detail`, `ip` — matches the handoff. Extras `severity_label`, `severity_class`, `time_iso` are server-precomputed for the badge / `<time datetime>`.

**Severity mapping.** `m→Info`, `w→Warning`, `e→Error` matches the `enum('m','w','e')` definition in `web/install/includes/sql/struc.sql:102` and every `Log::add()` call site (`m` for normal events, `w` for hacking-attempt warnings, `e` reserved for errors). Unknown letters fall through to a neutral `system` class.

**SQL / PDO.** Uses `:prefix_log` and `:prefix_admins` placeholders; `query → bind → single / resultset` pattern; `PDO::PARAM_INT` on the `LIMIT` binds (avoids the mysqlnd string-vs-int trap).

**CSRF / API contract / snapshots / baseline.** N/A — SSR, no state-change, no JSON action. `phpstan-baseline.neon` is unchanged vs `main`.

**CI.** All four gates green on `dc7f06a1` (PHPStan default + sbpp2026, PHPUnit, ts-check).

---

**Non-blocking flag — permission gate width.**

The Worker brief says the new gate matches the legacy `TRUNCATE :prefix_log` gate, which is true (admin.settings.php's `if ($userbank->HasAccess(ADMIN_OWNER))` around the truncate). However, the legacy *viewing* gate is the route-level `CheckAdminAccess(ADMIN_OWNER|ADMIN_WEB_SETTINGS)` on `case 'settings':` — the inner `AdminTabs` row for "System Log" is `ALL_WEB`, but it's effectively no-op behind that route gate.

So the new dedicated `?p=admin&c=audit` page tightens *viewing* from `ADMIN_OWNER|ADMIN_WEB_SETTINGS` to `ADMIN_OWNER`. `ADMIN_WEB_SETTINGS` holders aren't blocked from log content (the legacy tab remains), but they can't reach the new dedicated UI. Defensible posture (audit content includes IPs and hacking-attempt detail, so tighter is arguably better), just flagging in case the intent was parity with the legacy view gate. Not worth holding the merge.

Approving + merging.

Issue #1113, opened 2026-05-04
Fix dash_intro_text XSS and audit remaining nofilter usage


rumblefrog:
## Problem

#521 — `dash_intro_text` is rendered as raw HTML on the dashboard and is operator-controlled via the WYSIWYG settings page. With Smarty escape-by-default on (#1087), it's the most prominent surviving raw-HTML pathway. Compromising any account with `ADMIN_SETTINGS` lets an attacker inject JS executed in every visitor's session.

Shipping a 2.0 with a known XSS labeled `p: high` is not OK.

## Scope

- Replace the raw-HTML render with one of:
  - **(a)** HTMLPurifier with a strict allowlist (links, lists, headings, basic formatting), or
  - **(b)** a small Markdown-to-HTML pipeline (e.g. `league/commonmark`) since the WYSIWYG output is already structurally simple.

  Pick (b) if the WYSIWYG editor can be swapped to a Markdown editor — it removes a category of HTML-injection rather than chasing it.
- Audit every remaining `nofilter` / `{$x|smarty:nodefaults}` / `htmlspecialchars_decode` in `web/themes/default/`. Document each remaining one inline (`{* nofilter: this is admin-only and rendered inside <pre> *}`) or replace it with a safe alternative.
- Add a regression test in `web/tests/api/` that asserts a `<script>` tag in `dash_intro_text` is rendered escaped, not executed.

## Acceptance criteria

- [ ] `<script>` in `dash_intro_text` is rendered as text on the dashboard.
- [ ] Every remaining `nofilter` in default theme is annotated or removed.
- [ ] Settings form documents the new content format (Markdown or allowlisted HTML, depending on choice above).

## Notes

- Closes #521.
- 2.0 wave: C (depends on #1112 for the test scaffolding to lean on).

PR #1138, opened 2026-05-05
[#1123 B20] Redesign your account + updater

rumblefrog:
Two unrelated small templates ported to the new `sbpp2026` theme per
plan #1123 B20.

## Summary

- **Your Account** (`page_youraccount.tpl`) — replaces the legacy
  table-and-tab markup with a card-per-section layout: permissions
  summary, change password, server password, change email. Forms wire
  via inline vanilla JS + `sb.api.call(Actions.Account*)` (no
  MooTools / xajax / jQuery). Inputs and buttons carry stable
  `data-testid="account-<field>"` hooks.
- **Updater** (`updater.tpl`) — standalone HTML doc styled with the
  new theme; cards for the migration log + a "next step" panel. Each
  section header carries a `data-testid="updater-<step>"` hook. Each
  `{$update nofilter}` line carries the canonical safety comment
  pinned to `Updater.php` as the source of those strings (int versions
  + static templates; no user input).

## Delimiter rewrite (special note from the brief)

The legacy `default/page_youraccount.tpl` used the non-default
`-{ … }-` Smarty delimiter pair (a SourceBans-1.x oddity) and the
`YourAccountView` had to override `View::DELIMITERS` while the page
handler swapped `setLeftDelimiter` / `setRightDelimiter` on the live
`$theme` before / after `display()`. **The new template uses the
standard `{ }` pair end-to-end**, so:

- `Sbpp\View\YourAccountView` no longer declares a `DELIMITERS`
  constant.
- `web/pages/page.youraccount.php` no longer swaps delimiters and no
  longer double-renders via `AdminTabs` (the new card layout makes
  the tab bar redundant).
- `SmartyTemplateRule`'s tag regex matches the inner `{$var}` inside
  each legacy `-{$var}-`, so the dual-theme PHPStan matrix's "default"
  leg keeps finding every property here as "referenced" until D1
  deletes the legacy theme — no baseline entry needed for that
  interim window.

## Updater bootstrap note for reviewers

`init.php` pins `$theme_name = 'default'` whenever `IS_UPDATE` is
defined, so at runtime the updater still renders the legacy
`default/updater.tpl` until D1 deletes the legacy theme and renames
`sbpp2026/` into `default/`. SmartyTemplateRule scans whichever theme
`SBPP_PHPSTAN_THEME` selects per matrix leg, so the new template +
`UpdaterView` are statically validated under the `sbpp2026` leg today
— **no test-runtime adaptation needed**. To verify the new template
visually I rendered it through a one-off harness that swapped
`setTemplateDir(SB_THEMES . 'sbpp2026')` after init; the `<b>` markup
inside each `{$update nofilter}` line shows up bold (vs. the legacy
template emitting it as escaped literal text, which is itself a
#1113-shaped bug the new template fixes by making the `nofilter` use
explicit and annotated).

The `UpdaterView` deliberately declares only `$updates` so the same
View matches both templates under the dual-theme matrix — adding
extra fields would break the default leg's "unused property" check.

## PHPStan baseline cleanup

`global $theme;` in `web/updater/index.php` makes the previously
ignored `Variable $theme might not be defined.` warning stop firing,
so the matching baseline entry was removed (otherwise the default-leg
PHPStan would fail on `reportUnmatchedIgnoredErrors`).

## Files in scope

- `web/themes/sbpp2026/page_youraccount.tpl` (rewritten in standard
  `{ }` Smarty)
- `web/themes/sbpp2026/updater.tpl` (new template)
- `web/includes/View/YourAccountView.php` (drops `DELIMITERS`
  override; updated docblock)
- `web/includes/View/UpdaterView.php` (new view)
- `web/pages/page.youraccount.php` (switches to `Renderer::render`;
  drops `AdminTabs` + delimiter swap)
- `web/updater/index.php` (switches to `Renderer::render` with
  `UpdaterView`)
- `web/phpstan-baseline.neon` (drop the now-obsolete `$theme`
  baseline entry from `updater/index.php`)

## Test plan

- [x] `./sbpp.sh phpstan` (default leg) — clean
- [x] `SBPP_PHPSTAN_THEME=sbpp2026 phpstan --configuration=phpstan-sbpp2026.neon` — clean
- [x] `./sbpp.sh test` — 268 tests, 1134 assertions, OK
- [x] `./sbpp.sh composer api-contract` — no diff (no API change)
- [x] Manually verified `index.php?p=account` renders all four cards,
      emits all `data-testid="account-<field>"` hooks, and embeds
      `{csrf_field}` on every form
- [x] Manually verified the new `updater.tpl` renders via a one-off
      harness with `setTemplateDir(SB_THEMES . 'sbpp2026')` (the
      runtime IS_UPDATE pin still ships the legacy template until D1)

## Out of scope / explicitly NOT touched

Per the Phase B "NEVER touch" list: `web/themes/sbpp2026/{css,js,core}/*`,
`web/init.php`, `web/api/handlers/*.php`, `web/scripts/api-contract.js`,
`AGENTS.md`, `ARCHITECTURE.md`. The pre-existing
`web/updater/data/801.php` non-idempotency (already-seeded `attempts`
column) reproduces on `main`; also out of scope.

### DIFF
--- a/web/includes/View/UpdaterView.php
+++ b/web/includes/View/UpdaterView.php
@@ -1,0 +1,43 @@
+<?php
+declare(strict_types=1);
+
+namespace Sbpp\View;
+
+/**
+ * Updater wizard page — binds to `updater.tpl`.
+ *
+ * The updater is bootstrapped by `web/updater/index.php`, NOT by the
+ * normal `index.php` -> `route()` -> `pages/*.php` lifecycle. It still
+ * loads `web/init.php`, so the global `$theme` Smarty instance, the
+ * `$theme_url` assignment, and the configured template directory are
+ * all available — but `init.php` deliberately forces
+ * `$theme_name = "default"` when `IS_UPDATE` is defined (see the
+ * `defined("IS_UPDATE")` branch in `init.php`). That means the updater
+ * always renders the **legacy** `themes/default/updater.tpl` until D1
+ * deletes the legacy theme and renames `sbpp2026/` into `default/`,
+ * at which point this redesigned template takes over with no other
+ * code change.
+ *
+ * `SmartyTemplateRule` resolves the template against whichever
+ * theme directory is active (set per-leg via `SBPP_PHPSTAN_THEME` —
+ * see `phpstan.yml`'s matrix). Both legs of the matrix scan a
+ * template named `updater.tpl`; they share the same one-element
+ * variable contract (`$updates`) so the View doesn't need a
+ * per-theme alternate.
+ */
+final class UpdaterView extends View
+{
+    public const TEMPLATE = 'updater.tpl';
+
+    /**
+     * @param list<string> $updates Pre-formatted message lines from
+     *     {@see \Updater::getMessageStack()}. Each line may contain a
+     *     small subset of HTML markup (only `<b>` for emphasis around
+     *     int version numbers / static filenames built inside
+     *     `Updater.php`); no user input flows into them.
+     */
+    public function __construct(
+        public readonly array $updates,
+    ) {
+    }
+}
--- a/web/includes/View/View.php
+++ b/web/includes/View/View.php
@@ -16,9 +16,9 @@ namespace Sbpp\View;
  * See `web/includes/PHPStan/SmartyTemplateRule.php`.
  *
  * Most templates use Smarty's default `{…}` delimiters. Views whose template
- * uses the custom `-{…}-` pair (currently only `page_youraccount.tpl`) must
- * override {@see View::DELIMITERS} with the matching pair so the rule parses
- * the template correctly.
+ * uses the custom `-{…}-` pair (currently only `page_login.tpl`, in both the
+ * legacy and sbpp2026 themes) must override {@see View::DELIMITERS} with the
+ * matching pair so the rule parses the template correctly.
  *
  * ### Permission booleans
  *
--- a/web/includes/View/YourAccountView.php
+++ b/web/includes/View/YourAccountView.php
@@ -4,20 +4,31 @@ declare(strict_types=1);
 namespace Sbpp\View;
 
 /**
- * Your-account (self-service) page — binds to `page_youraccount.tpl`, which
- * is rendered with the custom `-{ … }-` delimiter pair (see
- * `page.youraccount.php`). The {@see View::DELIMITERS} override teaches
- * SmartyTemplateRule to scan for `-{$var}-` references instead of `{$var}`.
+ * Your-account (self-service) page — binds to `page_youraccount.tpl`.
+ *
+ * Historical note: prior to #1123 B20 this view rendered the legacy
+ * `default/page_youraccount.tpl`, which used the non-default `-{ … }-`
+ * Smarty delimiter pair (a SourceBans-1.x convention; see
+ * `web/pages/page.youraccount.php` history). The View overrode
+ * {@see View::DELIMITERS} so SmartyTemplateRule could parse the
+ * `-{$var}-` references; the page handler also had to swap delimiters
+ * on the live `$theme` before and after `display()`. The B20 redesign
+ * rewrites the template under `web/themes/sbpp2026/` in **standard**
+ * `{ }` Smarty delimiters so neither workaround is needed.
+ *
+ * The legacy `web/themes/default/page_youraccount.tpl` stays on disk
+ * with its `-{ }-` markers until D1 deletes the entire `default/`
+ * directory; SmartyTemplateRule's tag regex matches the inner `{$var}`
+ * inside each `-{$var}-` so the dual-theme PHPStan matrix's "default"
+ * leg keeps finding every property here as "referenced". No baseline
+ * entry is needed for that interim window.
  */
 final class YourAccountView extends View
 {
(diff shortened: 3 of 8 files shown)


rumblefrog (commented):
Initial review. 0 blocking, 4 non-blocking; addressing now.

**Scope / V iew DTO compliance / CSRF / testid hooks / baseline / CI** — all clean. The standard-delimiter rewrite of `page_youraccount.tpl` is correct, `YourAccountView` no longer overrides `DELIMITERS`, and `UpdaterView` matches the legacy-template variable contract (`$updates` only) so the dual-theme PHPStan matrix stays green on both legs (verified: `SmartyTemplateRule`'s tag regex matches the inner `{$var}` of legacy `-{$var}-`, so the default-leg parity check still finds every property as referenced — no baseline entry needed). `global $theme;` correctly takes care of the dropped `updater/index.php` baseline pair.

**Updater bootstrap** — runtime still pins `$theme_name='default'` under `IS_UPDATE`, so `Renderer::render($theme, new UpdaterView(...))` continues to render the legacy `default/updater.tpl`; the new template is statically validated only until D1 cuts over, as the View docblock says.

Non-blocking findings (will push fixes):

1. **Field-name mismatch in change-password JS** — `api_account_change_password` throws `ApiError(..., field: 'current')`, but the inline JS calls `setMsg('account-' + 'current' + '-msg', …)` against an element ID that's actually `account-current-password-msg`. Result: a wrong current password silently no-ops the form (no inline error). The email form correctly maps server fields → DOM IDs (`emailpw → email-password`, `email1 → email`, `email2 → email-confirm`); the password form needs the same mapping (`current → current-password`).

2. **Missing `env.redirect` check before `flashFailure`** — On a successful password change the dispatcher returns `{ok: false, redirect: '…'}` (Api::redirect) and `api.js` sets `window.location.href`. The `.then` callback fires before navigation completes; with `env.ok` false and no `env.error`, the code falls through to `flashFailure` (cf. the canonical `applyApiResponse()` in `web/scripts/sourcebans.js` which short-circuits on `res.redirect`).

3. **`sb.message` is a no-op in sbpp2026** — `sb.message.show()` looks up `#dialog-placement` / `#dialog-title` / etc. which only exist in `default/core/title.tpl`. The sbpp2026 theme replaces this with `SBPP.showToast({kind, title, body})` (see `themes/sbpp2026/js/theme.js`). As written, `flashSuccess` and `flashFailure` will silently disappear in this theme — the email-change / srv-password-change happy path eventually self-redirects via the legacy 5-second `sb.message.show` timeout, but the user sees no immediate feedback.

4. **Stale docblock cross-references to `YourAccountView`'s removed delimiter override** — `web/includes/View/View.php` ("currently only `page_youraccount.tpl`") and `web/includes/View/LoginView.php` ("`Sbpp\View\YourAccountView` pattern" / "B20 hasn't replaced its A1 stub yet") both reference the pre-B20 reality. Per `AGENTS.md` "Keep the docs in sync", these should be updated in this PR.

(Cosmetic / out of scope: the new `updater.tpl`'s `<link rel="icon" href="../themes/sbpp2026/images/favicon.ico">` 404s — `themes/sbpp2026/images/` doesn't exist yet; same pre-existing issue as `themes/sbpp2026/core/header.tpl`. Not introduced by this PR.)

rumblefrog:
Pushed `f351411c` addressing findings 1-3 from the initial review:

- `account-password-form` JS now passes a `{ current: 'current-password' }` field map into `showFieldError`, so a wrong current password surfaces inline (was silently no-opping because the server-side field name `current` didn't match the input's DOM id `current-password`). The map pattern matches the email form's existing `{ emailpw → email-password, email1 → email, email2 → email-confirm }` shape.
- `flashSuccess` / `flashFailure` now route through a `showToast()` helper that prefers `window.SBPP.showToast({ kind, title, body })` (the sbpp2026 surface from `themes/sbpp2026/js/theme.js`) and falls back to `sb.message.{success,error}` only when the page is rendered under the legacy default chrome — this fixes the silent-no-op symptom in sbpp2026 (which doesn't ship `#dialog-placement` etc.). The `m.redir` self-refresh after a successful srv-password / email change is preserved via a 1.5 s `setTimeout` so the toast is visible before the page reloads.
- Each `.then` callback short-circuits on `env.redirect` before the failure path so a successful `account.change_password` (which the dispatcher returns as `Api::redirect`) doesn't briefly flash "request failed" while the browser is already navigating to `/login`.

And finding 4 partially: `web/includes/View/View.php`'s docblock is updated in lockstep with the `YourAccountView` `DELIMITERS` drop. The matching stale references in `web/includes/View/LoginView.php` (which was added by B1 — not on this branch) will need a follow-up touch after both this PR and B1 are on `main`; flagging here so it isn't lost.

Local gates I ran in a worktree stack:
- `./sbpp.sh phpstan` — clean
- `./sbpp.sh test` — 268 tests, 1134 assertions, OK
- `./sbpp.sh ts-check` — clean
- `./sbpp.sh composer api-contract` — no diff
- The sbpp2026 leg of phpstan reproduces a pre-existing `DB_PORT` arg-type warning that's a dev-env-only artifact (the entrypoint-generated `config.php` defines `DB_PORT` as a string; CI uses `phpstan-bootstrap.php`'s int sentinel and is clean) — ran before/after my changes to confirm it's not introduced here.

PR #1139, opened 2026-05-05
[#1123 B14] Redesign admin comms group

rumblefrog:
## Summary
- Wires the **Add a block** and **Edit block** admin pages onto the sbpp2026 chrome, using the View DTO pattern A3 cemented (`Renderer::render` + `Sbpp\View\Perms::for`).
- Adds `Sbpp\View\AdminCommsAddView` and `AdminCommsEditView`, paired with `web/themes/sbpp2026/page_admin_comms_add.tpl` and `page_admin_edit_comms.tpl`. SmartyTemplateRule cross-checks both DTOs against both theme legs.
- `web/pages/admin.comms.php` and `web/pages/admin.edit.comms.php` switch from ad-hoc `$theme->assign(...)` chains to typed views; the edit handler also declares `global $userbank, $theme;` (instead of the legacy single-global form), so the matching `$userbank might not be defined` baseline entry drops.

## Why this is split per template
- **Delimiters.** The legacy default theme renders the edit form with non-standard `-{ … }-` Smarty delimiters; sbpp2026 uses the standard pair so the View binds with `View::DELIMITERS`' default. The handler picks per active theme so both legs of the dual-theme PHPStan matrix (#1123 A2) render correctly during the v2.0.0 rollout window.
- **Permissions.** SourceComms reuses the bans permission set — there is no `ADMIN_ADD_COMM`/`ADMIN_EDIT_COMM` flag — so the gate stays `ADMIN_OWNER|ADMIN_ADD_BAN`. The View surfaces it as `permission_addban` to match the legacy default-theme template's existing reference.
- **Hydration.** Type/length/current-reason on the edit form stay hydrated by inline `<script>selectLengthTypeReason(…)</script>` the handler emits after `Renderer::render` — mirrors the legacy convention so the same View satisfies SmartyTemplateRule on both legs.

## Conventions
- `data-testid="addcomm-<field>"` / `editcomm-<field>` on every input/select. `addcomm-submit` / `editcomm-submit` on the primary buttons; `addcomm-back` / `editcomm-back` on the secondary.
- `{csrf_field}` on the edit form (the add form posts via `sb.api.call(Actions.CommsAdd)` which threads the `X-CSRF-Token` header automatically).
- No new `{nofilter}` uses introduced.
- Vanilla JS only; no MooTools/jQuery/React reintroduced.
- All comms SQL stays on `:prefix_` literals (handler unchanged on that axis).

## Local gates
- `./sbpp.sh phpstan` (default leg): **OK** (no errors, 175 files).
- `SBPP_PHPSTAN_THEME=sbpp2026 ./sbpp.sh phpstan --configuration=phpstan-sbpp2026.neon` (sbpp2026 leg, with `DBA_REQUIRE=1`): **OK** (no errors, 175 files).
- `./sbpp.sh test`: **OK** (268 tests, 1134 assertions).
- `./sbpp.sh ts-check`: **OK** (clean — no JS touched).

## Test plan
- [x] PHPStan default + sbpp2026 legs clean (cache cleared).
- [x] PHPUnit clean.
- [x] ts-check clean.
- [ ] Reviewer: visit `?p=admin&c=comms` under both themes and confirm the form renders/submits.
- [ ] Reviewer: visit `?p=admin&c=editcomms&id=…&key=…` under both themes and confirm the type/length/reason hydrate from the inline `selectLengthTypeReason(...)` call.

Refs #1123 (Phase B14).

### DIFF
--- a/web/includes/View/AdminCommsAddView.php
+++ b/web/includes/View/AdminCommsAddView.php
@@ -1,0 +1,26 @@
+<?php
+declare(strict_types=1);
+
+namespace Sbpp\View;
+
+/**
+ * "Add a block" tab on the admin comms page — binds to
+ * `page_admin_comms_add.tpl`.
+ *
+ * SourceComms reuses the bans permission set: there is no separate
+ * `ADMIN_ADD_COMM` flag, so `permission_addban` (precomputed via
+ * {@see Perms::for()} as `can_add_ban`) gates the form. The naming
+ * matches the legacy `default/page_admin_comms_add.tpl` so a single
+ * View instance satisfies both themes during the v2.0.0 rollout window
+ * (#1123): SmartyTemplateRule scans whichever theme is active and the
+ * property name has to line up on both legs of the matrix.
+ */
+final class AdminCommsAddView extends View
+{
+    public const TEMPLATE = 'page_admin_comms_add.tpl';
+
+    public function __construct(
+        public readonly bool $permission_addban,
+    ) {
+    }
+}
--- a/web/includes/View/AdminCommsEditView.php
+++ b/web/includes/View/AdminCommsEditView.php
@@ -1,0 +1,32 @@
+<?php
+declare(strict_types=1);
+
+namespace Sbpp\View;
+
+/**
+ * Edit-existing-block form — binds to `page_admin_edit_comms.tpl`.
+ *
+ * The variable contract matches the legacy
+ * `default/page_admin_edit_comms.tpl`: only the player name and
+ * authid are surfaced through Smarty; the current type / length /
+ * reason are still hydrated from inline `<script>selectLengthTypeReason(…)</script>`
+ * the page handler emits after `Renderer::render` (legacy convention,
+ * preserved during the v2.0.0 rollout window so the same View
+ * satisfies SmartyTemplateRule on both theme legs of the CI matrix).
+ *
+ * The legacy template uses Smarty's custom `-{ … }-` delimiters; the
+ * sbpp2026 redesign moves to the standard `{ … }` pair so the View
+ * binds with the default {@see View::DELIMITERS}. The page handler
+ * picks the right delimiter set per active theme — see
+ * `web/pages/admin.edit.comms.php`.
+ */
+final class AdminCommsEditView extends View
+{
+    public const TEMPLATE = 'page_admin_edit_comms.tpl';
+
+    public function __construct(
+        public readonly string $ban_name,
+        public readonly string $ban_authid,
+    ) {
+    }
+}
--- a/web/pages/admin.comms.php
+++ b/web/pages/admin.comms.php
@@ -54,8 +54,17 @@ if (isset($_GET["rebanid"])) {
 
 echo '<div id="admin-page-content">';
 echo '<div class="tabcontent" id="Add a block">';
-$theme->assign('permission_addban', $userbank->HasAccess(ADMIN_OWNER | ADMIN_ADD_BAN));
-$theme->display('page_admin_comms_add.tpl');
+// SourceComms reuses the bans permission set: there is no
+// ADMIN_ADD_COMM flag, so the gate uses ADMIN_OWNER|ADMIN_ADD_BAN.
+// Splatting Perms::for(...) into the View pulls `can_add_ban` (and
+// the owner-bypass) without re-deriving the bitmask here; the
+// view-level property name stays `permission_addban` to match the
+// legacy default-theme template's existing reference (#1123 A3 +
+// SmartyTemplateRule's per-leg cross-check).
+$perms = \Sbpp\View\Perms::for($userbank);
+\Sbpp\View\Renderer::render($theme, new \Sbpp\View\AdminCommsAddView(
+    permission_addban: $perms['can_add_ban'],
+));
 ?>
 </div>
 <script type="text/javascript">
--- a/web/pages/admin.edit.comms.php
+++ b/web/pages/admin.edit.comms.php
@@ -22,7 +22,7 @@ if (!defined("IN_SB")) {
     die();
 }
 
-global $theme;
+global $userbank, $theme;
 
 new AdminTabs([], $userbank, $theme);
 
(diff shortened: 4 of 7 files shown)


rumblefrog (commented):
Initial review. 1 blocking, 2 non-blocking; addressing now.

## BLOCKING

### 1. `selectLengthTypeReason` is undefined under `sbpp2026` → edit-comms form silently clobbers existing values on submit

`web/pages/admin.edit.comms.php` tail emits, unconditionally on every theme:

```php
selectLengthTypeReason('<?=(int) $res['length']?>', '<?=(int) $res['type'] - 1?>', '<?=addslashes($res['reason'])?>');
```

`selectLengthTypeReason` lives in `web/scripts/sourcebans.js` (line 983) and is loaded **only** by `web/themes/default/core/header.tpl`. The `sbpp2026` chrome (`web/themes/sbpp2026/core/header.tpl`) loads `api-contract.js`, `sb.js`, `api.js`, `theme.js` — and explicitly drops `sourcebans.js` (per the B1 / B13 / D1 plan).

Trace under `config.theme=sbpp2026`:

1. Page renders. The new template hydrates `name` and `steam` via Smarty (`{$ban_name|escape}` / `{$ban_authid|escape}`). The `<select>` for `type`, `banlength`, `listReason` and the `<textarea id="txtReason">` are **not** Smarty-hydrated — every option emits without `selected`, defaulting to `value="1"` (Mute), `value="0"` (Permanent), `value=""` (the placeholder), and empty respectively.
2. Tail script runs. `window.addEvent('domready', …)` and the inline `changeReason` declaration succeed (sb.js shims those). Then `selectLengthTypeReason(…)` throws `ReferenceError`. Hydration never happens.
3. Admin sees the form with the player's name + authid filled in but the type / length / reason at defaults.
4. Admin clicks Save changes. `admin.edit.comms.php`'s POST handler reads `$_POST['type']=1`, `$_POST['banlength']=0`, `$_POST['listReason']=""`, runs the `UPDATE :prefix_comms SET type=:type, length=:length, reason=:reason …` — and the existing block's `type` / `length` / `reason` are silently overwritten with Mute / Permanent / empty.

This is data clobber on every edit through the new theme. The rest of the PR is otherwise clean, but I can't approve a path that loses customer data on the redesigned page it's introducing.

Two reasonable fixes:

* **Inline the function** alongside `changeReason` in the same tail `<script>` (mirrors how `ProcessBan` and `changeReason` are inlined in `admin.comms.php` for the add path). 10–15 lines of vanilla JS — `for (let opt of banlength.options) if (opt.value === String(length/60)) opt.selected = true;` etc.
* **Hydrate declaratively via Smarty.** Add `ban_type` / `ban_length_minutes` / `ban_reason` (+ `customreason`) to `AdminCommsEditView`, render `<option value="…" {if $ban_length_minutes == 60}selected{/if}>1 hour</option>` etc., drop the `selectLengthTypeReason` call. More aligned with the issue's "state in attributes, not styling" testability rule, but a larger diff.

Either works. The inline path is the smaller change and matches the B14 add-form's "preserve legacy convention via inline helpers" pattern.

(Note: `sourcebans.js` itself uses `if (typeof selectLengthTypeReason === 'function')` defensively at line 1374 in another callsite — exactly because the function may be absent.)

## NON-BLOCKING

### 2. Add-form response handlers (`ShowBlockBox`, `TabToReload`, `applyApiResponse`) are also `sourcebans.js`-only

`admin.comms.php` tail (lines 88–95):

```js
}).then(function (r) {
    if (r && r.ok && r.data && r.data.block) {
        ShowBlockBox(r.data.block.steam, r.data.block.type, r.data.block.length);
        if (r.data.reload) TabToReload();
        return;
    }
    applyApiResponse(r);
});
```

All three live in `sourcebans.js`. Under `sbpp2026`, the API call fires successfully (the block is added), but the response handler throws — admin gets no toast / no auto-refresh, just a silent submit.

Not data corruption, just degraded UX, and **B13 (already merged) has the same shape in `admin.bans.php`'s `ProcessBan` tail** — so this PR is consistent with the established precedent. Flagging for visibility; happy to defer to a Phase C/D follow-up that swaps the `sourcebans.js`-bound response handlers for `sb.message` toasts wholesale across the admin tabs. Not gating on this.

### 3. Per-theme delimiter conditional is a footgun for D1; B1 already established a cleaner pattern

`admin.edit.comms.php` does:

```php
$useLegacyDelims = Config::get('config.theme') !== 'sbpp2026';
if ($useLegacyDelims) { $theme->setLeftDelimiter('-{'); $theme->setRightDelimiter('}-'); }
\Sbpp\View\Renderer::render($theme, new \Sbpp\View\AdminCommsEditView(…));
if ($useLegacyDelims) { $theme->setLeftDelimiter('{'); $theme->setRightDelimiter('}'); }
```

After D1's `git mv web/themes/sbpp2026 web/themes/default`, every install ends up with `config.theme=default` pointing at the renamed sbpp2026 template (which uses `{ }`). The check evaluates to `'default' !== 'sbpp2026'` → `true` → swaps to `-{ }-` → renders the new template's `{$ban_name|escape}` as **literal text**. So this isn't quite "vestigial but harmless until cleanup" — it'll actively break the page at D1 unless the conditional is explicitly removed (D1's plan as written today doesn't list it).

The already-merged **B1** (`Sbpp\View\LoginView`, `web/pages/page.login.php`) chose the opposite tradeoff: rewrite the new sbpp2026 template to also use `-{ }-`, override `LoginView::DELIMITERS = ['-{', '}-']`, and let the page handler's unconditional `-{` / `}-` swap stay correct on both legs. SmartyTemplateRule scans both templates with the matching pair, no per-theme runtime branch, and D1's cleanup is local: rewrite the template body to `{ }` + drop `DELIMITERS` + drop the swap. No latent comparison to leak past the rename.

Either:
* **Adopt the B1 pattern** here too — rewrite `sbpp2026/page_admin_edit_comms.tpl` to use `-{ }-` and add `public const DELIMITERS = ['-{', '}-'];` to `AdminCommsEditView`. Drops the per-theme branch in the handler, kills the D1 footgun. (`page_admin_comms_add.tpl` is fine — both legacy and new use `{ }` already, no swap needed.)
* OR explicitly enumerate this conditional in #1123 D1's plan / checklist so it's not silently left behind.

I'd take the first; cheaper to fix today than to remember at cutover. Not gating on this — CI is green either way and the worker's tradeoff is defensible — but flagging because the user query explicitly asked me to assess "if it's the cleanest path".

---

Everything else is clean: scope matches the B14 dispatch (2 templates + 2 Views + 2 page handlers + 1 baseline), no anti-patterns reintroduced, `permission_addban` correctly reuses `ADMIN_OWNER|ADMIN_ADD_BAN` via `Perms::for()`, `data-testid` hooks on every input/select/button, `{csrf_field}` on the edit POST form, add form goes through `sb.api.call(Actions.CommsAdd)`, 6 baseline removals (allowed) with no additions, all 4 CI gates green.

rumblefrog:
Sweep 1.

Pushed `8813c34d` for the BLOCKING (#1).

* **Inlined `selectLengthTypeReason`** in `admin.edit.comms.php`'s tail script with raw-DOM `document.getElementById` and the same matching logic as `sourcebans.js`. Function-declaration hoisting makes the inline copy shadow the sourcebans.js definition on the legacy default theme, so both legs of the matrix get identical behaviour and the page keeps working post-D1 when sourcebans.js is deleted.
* Local gates: PHPStan default + sbpp2026, PHPUnit (268 tests / 1134 assertions), ts-check — all green.

Leaving the two non-blocking notes on the table:
* **#2** (add-form response handlers reach for `sourcebans.js`-only `ShowBlockBox` / `TabToReload` / `applyApiResponse`) — same shape as the already-merged B13. Belongs in a follow-up that sweeps the admin tabs onto `sb.message`-style toasts, not in B14.
* **#3** (per-theme delimiter conditional vs. B1's `DELIMITERS` override) — leaving the worker's choice as-is. The `Config::get('config.theme') !== 'sbpp2026'` branch will need explicit removal at #1123 D1 (otherwise the renamed `default/` template will render with `-{ }-` and `{$ban_name|escape}` will leak as literal text). Worth dropping a one-line entry in D1's checklist; happy to file that separately.

Once CI is green I'll approve + merge.

PR #1140, opened 2026-05-05
[#1123 C1] Drawer wiring + bans.detail JSON action

rumblefrog:
Part of #1123 (v2.0.0 theme rollout).
Plan stage: **C1 — Drawer JS wiring + new bans.detail JSON action.**

## Summary

- Adds `bans.detail` — a public JSON handler that returns the
  player-detail payload the sbpp2026 right-side drawer renders.
- Wires the drawer in `web/themes/sbpp2026/js/theme.js` to call
  `sb.api.call(Actions.BansDetail, { bid })` (replaces the C1 TODO
  breadcrumb A1 left).
- Adds a reference partial at
  `web/themes/sbpp2026/partials/player-drawer.tpl` mirroring the
  JS-rendered markup so a future B-phase ticket can deep-link to a
  pre-populated drawer without forking the visual.

## Response shape: **JSON** (not server-rendered HTML)

The plan asked us to pick one and document it.

`bans.detail` returns structured JSON; the drawer JS renders the
markup client-side via `renderDrawerBody()`, funnelling every dynamic
value through the existing `escapeHtml()` helper. The wire-format
typedef lives in `web/scripts/api-contract.js` (regenerated by this
PR), and four snapshots under
`web/tests/api/__snapshots__/bans/detail_*.json` lock it.

Why JSON over server-rendered HTML:
- Matches every other handler in `web/api/handlers/` — no Smarty
  machinery in the API path, handlers stay pure
  `function(array $params): array`.
- Snapshot file is a stable, reviewable data shape — HTML snapshots
  are brittle to layout tweaks.
- Wire payload is small and self-describing.
- The CSS classes used by the drawer (`drawer__header`,
  `drawer__body`, `text-faint`, `font-mono`, `skeleton`, …) all live
  in `web/themes/sbpp2026/css/theme.css` from A1, so the JS-built
  markup picks up the design tokens for free.

The trade-off is that the new partial is a *reference shape*, not the
live render path — the JS is the source of truth. The partial is in
scope per the plan and useful for designers + future SSR fallbacks
(deep-link `?p=banlist&id=N` opening the drawer pre-populated).

## Permissions

`bans.detail` is registered **public** (`public=true`) so the drawer
matches the public ban-list page's reach — clicking a row in the public
ban list should open the drawer for any visitor.

The handler then enforces the same per-field gates `page.banlist.php`
already does:

| Field          | Hidden when                                                          |
|----------------|----------------------------------------------------------------------|
| `player.ip`    | `banlist.hideplayerips=1` AND caller is not an admin                |
| `admin.name`   | `banlist.hideadminname=1` AND caller is not an admin                |
| `ban.removed_by` | (same as `admin.name`)                                              |
| `comments`     | `config.enablepubliccomments=0` AND caller is not an admin → `[]`  |

So a public caller never sees fields the page would have suppressed
for the same caller; an admin sees the underlying data. Two of the
new snapshots (`detail_public_hidden.json`,
`detail_admin_view.json`) lock both sides of that contract.

`web/tests/api/PermissionMatrixTest.php` gets a new row pinning
`(perm=0, requireAdmin=false, public=true)`.

## Drawer JS contract

`web/themes/sbpp2026/js/theme.js`:

- Replaces the C1 TODO breadcrumb.
- Click triggers accept either `data-drawer-bid="N"` (preferred) or
  `data-drawer-href="?p=banlist&id=N"` (handoff template form).
- Toggles **state attributes** for testability:
  - `data-drawer-open="true|false"` on `<aside id="drawer-root">`
  - `data-loading="true"` while the fetch is in flight
- Renders skeleton during fetch, error state on failure, populated
  detail on success — all built from `escapeHtml()`-funnelled strings.
- No string literals for action names — uses `Actions.BansDetail`.

## Files touched

- `web/api/handlers/bans.php` — new `api_bans_detail`
- `web/api/handlers/_register.php` — register `bans.detail` public
- `web/scripts/api-contract.js` — regenerated; adds `BansDetail`
- `web/tests/api/PermissionMatrixTest.php` — new matrix row
- `web/tests/api/BansTest.php` — 4 new tests
  (`testDetailRejectsMissingBid`, `testDetailReturns404ForUnknownBid`,
  `testDetailPublicViewHidesAdminFields`,
  `testDetailAdminViewExposesEverything`)
- `web/tests/api/__snapshots__/bans/detail_{bad_request,not_found,public_hidden,admin_view}.json`
- `web/themes/sbpp2026/js/theme.js` — drawer fetch + render
- `web/themes/sbpp2026/partials/player-drawer.tpl` — reference shape

## Local gates

- `./sbpp.sh phpstan` — green
- `./sbpp.sh phpstan` w/ `SBPP_PHPSTAN_THEME=sbpp2026 --configuration=phpstan-sbpp2026.neon` — green
- `./sbpp.sh test tests/api/BansTest.php` — 30 / 30 green
- `./sbpp.sh test tests/api/PermissionMatrixTest.php` — 60 / 60 green
- `./sbpp.sh ts-check` — green
- `./sbpp.sh composer api-contract` — emits 59 actions, 32 perms, 3 typedefs; idempotent (no further diff after second run)

Manual verification with `config.theme=sbpp2026`:

- Public caller (no login) → `ip=null`, `admin.name=null`,
  `comments_visible=false`, `comments=[]`. ✓
- Admin caller → `ip="203.0.113.42"`, `admin.name="CONSOLE"`,
  `comments_visible=true`. ✓
- Drawer scaffold rendered: `<aside id="drawer-root" data-drawer-open="false" hidden>` present in the page chrome. ✓
- `theme.js` loads after `api-contract.js`, `sb.js`, `api.js`. ✓
- `Actions.BansDetail = 'bans.detail'` exposed in the served contract. ✓
- `config.theme` reverted to `default` after the verification.

Live screenshots aren't included here — this CLI environment doesn't
have a headless browser available, and installing one for one
verification run would have ballooned the diff churn. Reviewers can
spin up `./sbpp.sh up`, switch `config.theme` to `sbpp2026`, click any
public ban row, and observe the drawer.

## Test plan

- [ ] `./sbpp.sh phpstan` (default + `--configuration=phpstan-sbpp2026.neon`) is green.
- [ ] `./sbpp.sh test` is green (full suite, not just the new tests).
- [ ] `./sbpp.sh ts-check` is green.
- [ ] `./sbpp.sh composer api-contract` produces no diff after running.
- [ ] Visual: with `config.theme=sbpp2026`, click a public ban row and
      confirm the drawer slides in with player + ban + (admin-only)
      comment data.
- [ ] Light + dark — toggle the theme button and confirm the drawer
      respects `<html class="dark">`.

## Notes for the rebase Reviewer

C1 and C2 both touch `web/api/handlers/bans.php`. Per the plan's
ordering guidance, the second-merging Reviewer handles the rebase
conflict. No special action requested from this PR's author.

### DIFF
--- a/web/api/handlers/_register.php
+++ b/web/api/handlers/_register.php
@@ -31,6 +31,13 @@ require_once __DIR__ . '/system.php';
 // ---- public actions (no auth required) --------------------------------
 Api::register('auth.login',         'api_auth_login',         0, false, true);
 Api::register('auth.lost_password', 'api_auth_lost_password', 0, false, true);
+// bans.detail mirrors the public ban-list page's reach: any visitor can
+// click a ban row in the sbpp2026 drawer (#1123 C1). Player IP, admin
+// name, removed-by and comments are gated *inside* the handler against
+// banlist.hideplayerips / banlist.hideadminname / config.enablepubliccomments
+// + is_admin(), matching page.banlist.php exactly so we don't leak
+// fields the page intentionally suppresses.
+Api::register('bans.detail',        'api_bans_detail',        0, false, true);
 
 // ---- account: dispatcher enforces login; handler enforces aid match ---
 Api::register('account.check_password',     'api_account_check_password');
--- a/web/api/handlers/bans.php
+++ b/web/api/handlers/bans.php
@@ -653,3 +653,183 @@ function api_bans_view_community(array $params): array
 
     throw new ApiError('player_not_found', "Can't get playerinfo for " . htmlspecialchars($name) . '. Player not on the server anymore!');
 }
+
+/**
+ * Player-detail payload for the sbpp2026 right-side drawer (#1123 C1).
+ *
+ * Returns the same data the public ban-list page already exposes, in a
+ * stable JSON shape the drawer JS renders client-side. Action is
+ * registered public so the drawer matches the public ban-list reach;
+ * fields that the public ban-list selectively hides via
+ * `banlist.hideplayerips` / `banlist.hideadminname` (player IP, admin
+ * name, removed-by) follow the same gating here so a public caller can
+ * never fan out the same data through the JSON API that the page
+ * intentionally suppresses. Comments are only included when
+ * `config.enablepubliccomments` is set or the caller is an admin —
+ * mirroring page.banlist.php's `$view_comments` switch.
+ *
+ * @param array{bid?: int|string} $params
+ * @return array{
+ *   bid: int,
+ *   player: array{name: string, type: int, steam_id: string, steam_id_3: string, community_id: string, ip: string|null, country: string|null},
+ *   ban: array{reason: string, banned_at: int, banned_at_human: string, length_seconds: int, length_human: string, expires_at: int|null, expires_at_human: string|null, state: string, unban_reason: string, removed_at: int|null, removed_at_human: string|null, removed_by: string|null},
+ *   admin: array{name: string|null},
+ *   server: array{sid: int, name: string|null, mod_icon: string|null},
+ *   demo_count: int,
+ *   history_count: int,
+ *   comments_visible: bool,
+ *   comments: list<array{cid: int, added: int, added_human: string, author: string|null, text: string, edited_at: int|null, edited_by: string|null}>
+ * }
+ */
+function api_bans_detail(array $params): array
+{
+    global $userbank;
+    $bid = (int)($params['bid'] ?? 0);
+    if ($bid <= 0) {
+        throw new ApiError('bad_request', 'bid must be a positive integer', 'bid');
+    }
+
+    // Mirror page.banlist.php: an admin (or any logged-in user with the
+    // appropriate flag) sees the IP and admin nick; public visitors
+    // get them suppressed when the corresponding hide-* settings are
+    // on. `is_admin()` is the same gate the page uses, so the JSON
+    // surface stays consistent with what the HTML page would have
+    // shown the same caller.
+    $isAdmin = $userbank->is_admin();
+    $hideIps   = Config::getBool('banlist.hideplayerips') && !$isAdmin;
+    $hideAdmin = Config::getBool('banlist.hideadminname') && !$isAdmin;
+
+    $row = $GLOBALS['PDO']->query(
+        "SELECT BA.bid, BA.type, BA.ip, BA.authid, BA.name, BA.created, BA.ends, BA.length,
+                BA.reason, BA.aid, BA.adminIp, BA.sid, BA.country, BA.RemovedOn, BA.RemovedBy,
+                BA.RemoveType, BA.ureason,
+                AD.user AS admin_name,
+                SE.ip AS server_ip, SE.port AS server_port,
+                MO.icon AS mod_icon, MO.name AS mod_name,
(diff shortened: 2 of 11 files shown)


rumblefrog (commented):
Initial review. 1 blocking, 1 non-blocking; addressing now.

Approach is sound — JSON shape (not server-rendered HTML) matches every other handler in `web/api/handlers/`, the four snapshots lock the wire format, and the `(player.ip / admin.name / removed_by / comments)` per-field gating mirrors `web/pages/page.banlist.php` exactly (`Config::getBool('banlist.hideplayerips') && !$userbank->is_admin()`, `Config::getBool('banlist.hideadminname') && !$userbank->is_admin()`, `Config::getBool('config.enablepubliccomments') || $userbank->is_admin()`). The `public=true` registration matches the public ban-list reach. CI is green across all five gates (PHPStan default + sbpp2026, PHPUnit, ts-check, api-contract). PHPStan baseline is unchanged. The new `BansDetail` action and typedef are present in `web/scripts/api-contract.js`, and the `PermissionMatrixTest.php` row pins `(perm=0, requireAdmin=false, public=true)`. The two contract-locking tests `testDetailPublicViewHidesAdminFields` and `testDetailAdminViewExposesEverything` exercise both sides of the gate.

## Blocking

1. **Drawer "X" close button is dead UI** — `web/themes/sbpp2026/js/theme.js:383`. The pre-existing close-click handler is

   ```
   if (target && target.closest('[data-drawer-close]') && !target.closest('.drawer')) closeDrawer();
   ```

   The `&& !target.closest('.drawer')` clause was fine in A1 because the drawer had no inner close target — only the backdrop (a sibling of `.drawer`) carried `data-drawer-close`. C1 introduces three new `<button data-drawer-close aria-label=\"Close\">` X buttons inside `<div class=\"drawer\">` (lines 235, 319, 339 — header, loading, error), and they are all swallowed by the `.drawer` filter (`closest('.drawer')` walks up and finds the wrapping `.drawer`, so `!truthy = false`, close is never invoked). Backdrop click and Escape still close the drawer, so it isn't a soft-lock, but a visually prominent X button that does nothing is a UX regression introduced by this PR. Fix: drop `&& !target.closest('.drawer')` from line 383 — `[data-drawer-close]` is the explicit close marker and no other element on the page carries it, so the additional filter is redundant and now actively wrong.

## Non-blocking

1. **Registration placement** — `web/api/handlers/_register.php`. `bans.detail` is registered in the top \"public actions (no auth required)\" comment block (next to `auth.login`) instead of inline in the bans group where every other `bans.*` action lives. `servers.refresh` and `servers.host_*` (also `public=true`) are registered inline in the `servers.*` block with the public flag set in the call, which keeps each topic's registrations contiguous and grep-friendly. Mention only — won't block on this.

Addressing the blocking finding now.

rumblefrog (commented):
Blocking finding addressed in `1b021886` — dropped the `&& !target.closest('.drawer')` clause from the close-click handler in `web/themes/sbpp2026/js/theme.js:383` so the new in-header X buttons (and the loading/error states' X buttons) actually close the drawer. Backdrop and Escape paths unchanged. CI green across all five gates on the new commit. Non-blocking placement nit declined as a stylistic preference — leaving it for a follow-up if anyone cares.

Squash-merging.

PR #1141, opened 2026-05-05
[#1123 B15] Redesign admin servers group

rumblefrog:
## Summary

Phase B15 of the #1123 sbpp2026 theme rollout: swap the four admin
"Servers" templates for the new theme and finish wiring the page
handler to the View DTO pattern Phase A established. Touches only
files in scope per the plan; no API handlers / `init.php` / `core/` /
`css/` / `js/` / contract changes.

The screens covered:

- **Server list** (`?p=admin&c=servers`) — card grid replacing the
  legacy table. Each card has `data-testid=\"server-tile\"` +
  `data-id=\"{$server.sid}\"`, a status pill, mod label, and gated
  RCON / Admins / Edit / Delete buttons. Inline `{literal}` script
  wires Delete confirmation through `Actions.ServersRemove`.
- **Add server** (rendered just below the grid) — full form with
  `{csrf_field}`, every input tagged `data-testid=\"addserver-*\"`,
  inline submit handler calling `Actions.ServersAdd` (sbpp2026 doesn't
  load the legacy `process_add_server()` helper from
  `web/scripts/sourcebans.js`, so the form posts via `sb.api.call`
  directly).
- **Server admin list** (`?p=admin&c=servers&o=admincheck&id=…`,
  rendered by `admin.srvadmins.php`) — table of admins mapped to a
  server, with a live status pill when the SourceQuery probe
  succeeded.
- **RCON console** (`?p=admin&c=servers&o=rcon&id=…`, rendered by
  `admin.rcon.php`) — terminal-style console with
  `data-testid=\"rcon-input\"` / `rcon-output`. Inline script speaks
  `Actions.ServersSendRcon`; the legacy `id=\"rcon\"` /
  `id=\"rcon_con\"` / `id=\"cmd\"` IDs are preserved so any external
  scripting that targets the console still finds them. Uses the
  custom `-{ … }-` delimiter pair to match how `admin.rcon.php`
  configures Smarty.

## What changed

- `web/themes/sbpp2026/page_admin_servers_list.tpl` — card grid.
- `web/themes/sbpp2026/page_admin_servers_add.tpl` — add form with
  `{csrf_field}` + JSON-API submit.
- `web/themes/sbpp2026/page_admin_servers_adminlist.tpl` — admins on
  this server.
- `web/themes/sbpp2026/page_admin_servers_rcon.tpl` — RCON console
  using the `-{ … }-` delimiter pair.
- `web/includes/View/AdminServersAdminListView.php` (new) — declares
  `\$admin_count` + `\$admin_list`.
- `web/includes/View/AdminServersRconView.php` (new) — declares
  `\$id` + `\$permission_rcon`, overrides `View::DELIMITERS` so
  SmartyTemplateRule scans the template with the `-{ … }-` pair.
- `web/pages/admin.servers.php` — extended the existing list query
  with `mo.name AS mod_name` so the new card can render the mod
  label without a second per-tile query (the legacy default-theme
  template ignores the column, so both themes share the same data
  shape). The `Renderer::render` calls were already in place from
  Phase A.

## Notes

- No new permission flags. `permission_list` / `permission_addserver`
  / `permission_editserver` / `pemission_delserver` (historical typo
  preserved to match the existing baseline) are still precomputed in
  `admin.servers.php`.
- No changes to API handlers — the four screens reuse
  `servers.add`, `servers.remove`, `servers.send_rcon`.
- Every `{literal}…{/literal}` block exists because the inline JS
  uses object literals + control-flow braces that would otherwise be
  parsed as Smarty tags. **No `nofilter` is used anywhere** in the
  four templates; every dynamic value goes through Smarty's
  auto-escape (or an explicit `|escape`) before reaching the page.
- `admin.rcon.php` and `admin.srvadmins.php` are unchanged — they
  already drive their templates through `\$theme->display(...)` and
  the paired Views satisfy SmartyTemplateRule.

## Test plan

- [x] `./sbpp.sh phpstan` (default theme matrix leg) → **OK**
- [x] `SBPP_PHPSTAN_THEME=sbpp2026 phpstan` (sbpp2026 leg) → **OK**
- [x] `./sbpp.sh ts-check` → **OK**
- [x] `./sbpp.sh composer api-contract` → no diff (handlers untouched)
- [x] `./sbpp.sh test` → 268 tests, 1134 assertions → **OK**
- [x] Smoke test in browser via `./sbpp.sh up`:
  - `?p=admin&c=servers` (list + add form) renders end-to-end.
  - `?p=admin&c=servers&o=rcon&id=…` renders the console (Access
    Denied path also verified by clearing the per-server mapping).
  - `?p=admin&c=servers&o=admincheck&id=…` renders the admin list
    with both online + offline states (verified by mocking
    \`checkMultiplePlayers\`'s output).

### DIFF
--- a/web/includes/View/AdminServersAdminListView.php
+++ b/web/includes/View/AdminServersAdminListView.php
@@ -1,0 +1,38 @@
+<?php
+declare(strict_types=1);
+
+namespace Sbpp\View;
+
+/**
+ * "Server admins" panel rendered by `pages/admin.srvadmins.php` — binds to
+ * `page_admin_servers_adminlist.tpl`. Lists the admins with access to the
+ * selected server (directly or through a server group) and, when the
+ * SourceQuery probe succeeded, the live in-game name + IP for each one.
+ *
+ * Each row in `$admin_list` mirrors the shape `admin.srvadmins.php` builds:
+ *
+ *   - `user`   string|null — admin display name from `:prefix_admins`.
+ *   - `authid` string|null — admin SteamID.
+ *   - `ingame` bool        — true when the admin is currently connected.
+ *   - `iname`  string|null — in-game name (only when `ingame` is true).
+ *   - `iip`    string|null — in-game IP (only when `ingame` is true).
+ */
+final class AdminServersAdminListView extends View
+{
+    public const TEMPLATE = 'page_admin_servers_adminlist.tpl';
+
+    /**
+     * @param list<array{
+     *     user: string|null,
+     *     authid: string|null,
+     *     ingame: bool,
+     *     iname?: string|null,
+     *     iip?: string|null,
+     * }> $admin_list
+     */
+    public function __construct(
+        public readonly int $admin_count,
+        public readonly array $admin_list,
+    ) {
+    }
+}
--- a/web/includes/View/AdminServersRconView.php
+++ b/web/includes/View/AdminServersRconView.php
@@ -1,0 +1,29 @@
+<?php
+declare(strict_types=1);
+
+namespace Sbpp\View;
+
+/**
+ * Per-server RCON console — binds to `page_admin_servers_rcon.tpl`, which
+ * is rendered with the custom `-{ … }-` delimiter pair (see
+ * `pages/admin.rcon.php`). The {@see View::DELIMITERS} override teaches
+ * SmartyTemplateRule to scan for `-{$var}-` references instead of `{$var}`.
+ *
+ * The console itself talks to `Actions.ServersSendRcon`; the page only
+ * needs the target `$id` (server sid) and a precomputed
+ * `$permission_rcon` flag combining `SM_RCON|SM_ROOT` with the per-server
+ * access loop in `admin.rcon.php`.
+ */
+final class AdminServersRconView extends View
+{
+    public const TEMPLATE = 'page_admin_servers_rcon.tpl';
+
+    /** @var array{0: string, 1: string} */
+    public const DELIMITERS = ['-{', '}-'];
+
+    public function __construct(
+        public readonly int $id,
+        public readonly bool $permission_rcon,
+    ) {
+    }
+}
--- a/web/pages/admin.servers.php
+++ b/web/pages/admin.servers.php
@@ -10,7 +10,11 @@ new AdminTabs([
     ['name' => 'Add new server', 'permission' => ADMIN_OWNER|ADMIN_ADD_SERVER]
 ], $userbank, $theme);
 
-$servers = $GLOBALS['PDO']->query("SELECT srv.ip ip, srv.port port, srv.sid sid, mo.icon icon, srv.enabled enabled FROM `:prefix_servers` AS srv
+// `mod_name` (mod display name) was added in #1123 B15 so the sbpp2026
+// card grid can render the mod label without a second per-card query.
+// The legacy `default/` template ignores the column; both themes share
+// the same data shape.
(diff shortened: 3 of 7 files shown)


rumblefrog (commented):
Initial review. 0 blocking, 0 actionable non-blocking; ready for sweep.

**Verified against B15 checklist:**

- **Scope** — 4 sbpp2026 templates + 4 Views (2 new: `AdminServersAdminListView` / `AdminServersRconView`; 2 from Phase A) + a 1-line `mod_name` enrichment in `admin.servers.php`. Both legacy + sbpp2026 templates consume the shared shape; legacy `default/page_admin_servers_list.tpl` ignores `mod_name`, no behaviour change there.
- **Anti-patterns** — Clean. Vanilla JS (no MooTools/React), no `xajax`/`sb-callback.php`, no ADOdb, no `nofilter`, no `htmlspecialchars_decode` on JSON-API params (the one `html_entity_decode` call site is in the pre-existing `api_servers_send_rcon` handler with the documented exemption — outside this PR).
- **View DTO + SmartyTemplateRule** — All 4 Views paired with their `.tpl`. `AdminServersRconView::DELIMITERS = ['-{', '}-']` correctly teaches the rule to scan with the alternate pair. Both legs of the dual-theme PHPStan matrix pass.
- **CSRF** — `{csrf_field}` on the add form, `-{csrf_field}-` on the rcon form (alternate delimiter).
- **Testability hooks** — All present per plan: `data-testid="server-tile"` + `data-id="{$server.sid}"` on each card, `addserver-<field>` on every add input, `rcon-input` / `rcon-output` on the console.
- **JS conventions** — Vanilla, defensive guards, uses `sb.api.call(window.Actions.<Name>, …)` (not string literals), `textContent`+`appendChild` on gameserver-controlled bytes (mirrors the `LoadSendRcon` defence in `sourcebans.js`), `confirm()` before destructive `ServersRemove`.
- **API contract / snapshots** — N/A; no handler changes, no diff in `api-contract.js` or `__snapshots__/`.
- **PHPStan baseline** — No additions.
- **CI** — All 4 checks green (PHPStan default + sbpp2026 legs, PHPUnit, ts-check). Mergeable: CLEAN.
- **Behavioural correctness** — The d2b59b9c rcon-comment fix is robust: I verified the entire rephrased docblock (lines 1–24) and the inner `*}-` block (41–48) contain zero `}` characters in the body, so `#-\{([^}]*?)\}-#s` cannot terminate prematurely. Every other `-{…}-` tag in the template body has a brace-free body (`$id`, `if NOT $permission_rcon`, `else`, `csrf_field`, `/if`). The script is outside any `-{…}-` block so its `}` chars don't matter to the regex.

**Observations (intentional, won't be addressed in this PR):**

1. The View constructors use `permission_list` / `permission_addserver` / `pemission_delserver` (typo preserved) rather than the `Perms::for` `can_*` convention because the Phase A `AdminServersListView` already uses these names and they match the legacy `default/` template — required for the dual-theme matrix to keep both legs green. Migrating naming is its own ticket.
2. `admin.rcon.php` and `admin.srvadmins.php` still use `$theme->assign() + $theme->display()` rather than `Renderer::render(new AdminServers…View(…))`. The Views exist purely to satisfy SmartyTemplateRule on the redesigned templates; the PR description calls this out as deliberate. Page migration is a follow-up.
3. The card-grid template uses `$server.ip` / `$server.port` / `$server.mod_name` directly rather than the handoff's `$server.host` / `$server.mod`, with `data-hydrate="players"` / `data-hydrate="map"` slots reserved for the Phase C JS to populate live host / map / player counts (sbpp2026 doesn't ship the legacy `LoadServerHost` helper).

Approving + merging next.

PR #1142, opened 2026-05-05
[#1123 B17] Redesign admin overrides

rumblefrog:
## Summary

Phase B17 of the #1123 sbpp2026 theme migration. Extracts the
SourceMod command/group overrides editor out of `admin.admins.php`'s
tab grab-bag into a single-template, single-View unit so each Phase B
ticket maps cleanly to one `Sbpp\View\*` DTO and one `.tpl` per
theme.

- New `web/pages/admin.overrides.php` holds the POST save/delete/dup
  logic (verbatim from the legacy block) and renders via
  `Sbpp\View\Renderer` with the new `AdminOverridesView` DTO.
- `web/pages/admin.admins.php` delegates the overrides tab to the
  new file via a top-level `require`, so the existing AdminTabs URL
  (`?p=admin&c=admins`) keeps its three-tab layout unchanged for
  both themes (no routing change).
- `web/includes/View/AdminOverridesView.php` declares the four
  template variables (`permission_addadmin`, `overrides_error`,
  `overrides_save_success`, `overrides_list`) the legacy default
  template still needs, so `SmartyTemplateRule` matches both themes.
  Each row carries both the legacy `id`/`name` keys and the
  redesigned `oid`/`command_or_group` aliases so the same DTO
  satisfies both `.tpl` files during the dual-theme rollout.
- `web/themes/sbpp2026/page_admin_overrides.tpl` replaces the A1
  stub with a real redesign (cards, inputs, table) that:
  - calls `{csrf_field}` on the form,
  - tags each row with `data-testid="override-row"` +
    `data-id="{$override.oid}"`,
  - tags the add-row inputs with `data-testid="addoverride-{type,name,flags}"`,
  - keeps every `{nofilter}` (just the `$overrides_error` HTML)
    behind the canonical safety comment per AGENTS.md.

## Local gates

| Gate | Result |
| --- | --- |
| `./sbpp.sh phpstan` (default theme) | OK No errors (175 files) |
| `SBPP_PHPSTAN_THEME=sbpp2026 ... phpstan-sbpp2026.neon` | OK No errors (175 files) |
| `./sbpp.sh ts-check` | OK |
| `./sbpp.sh composer api-contract` | clean diff (no handler changes) |
| `./sbpp.sh test --filter='AdminFlowTest|AdminsTest|PermissionMatrixTest'` | OK (79 tests, 311 assertions) |

The full test suite was noisy on this worker due to multiple parallel
B-task agents sharing the same dev DB volume (race conditions on
`Fixture::install()` dropping/recreating `sourcebans_test` between
them); the targeted admin-suite + matrix run above is the relevant
slice and is clean.

## Manual smoke

Hit `?p=admin&c=admins` on both themes against the worker's local
stack (`http://localhost:8390`) after seeding two override rows:

- Default theme: legacy `page_admin_overrides.tpl` still renders with
  the existing copy and form fields (no behavioural diff).
- sbpp2026 theme: renders the redesigned card layout, two
  `<tr data-testid="override-row" data-id="1">` /
  `data-id="2"` rows with the right type/name/flags pre-filled, and
  the add-row with `data-testid="addoverride-type|name|flags"` hooks.

## Test plan

- [ ] CI: PHPStan (default + sbpp2026 matrix) green
- [ ] CI: PHPUnit green (this PR doesn't touch any handler so the
      shared DB races above shouldn't recur on a clean runner)
- [ ] CI: ts-check green
- [ ] CI: api-contract green (no diff expected)
- [ ] Reviewer manual: switch a test admin to `sbpp2026` theme, hit
      `?p=admin&c=admins`, click the **Overrides** tab, add / edit / delete
      rows, confirm the success/error banners render and that
      blanking out a name still deletes that override on save (parity
      with the legacy default-theme tab).

### DIFF
--- a/web/includes/View/AdminOverridesView.php
+++ b/web/includes/View/AdminOverridesView.php
@@ -1,0 +1,50 @@
+<?php
+declare(strict_types=1);
+
+namespace Sbpp\View;
+
+/**
+ * SourceMod command/group overrides editor — binds to
+ * `page_admin_overrides.tpl`. Overrides let admins flip the required
+ * flags on any in-game command without editing plugin source.
+ *
+ * Until #1123 D1 cuts over, the same DTO has to satisfy both
+ * `web/themes/default/page_admin_overrides.tpl` (legacy markup) and
+ * `web/themes/sbpp2026/page_admin_overrides.tpl` (new redesign), so:
+ *
+ *   - The permission gate is exposed as `$permission_addadmin` (the
+ *     legacy name) rather than the canonical `can_add_admins` from
+ *     {@see Perms::for()}; renaming it would require editing the
+ *     legacy default template, which is out of scope for B17.
+ *   - Each row in `$overrides_list` carries BOTH the legacy
+ *     `id` / `name` keys (read by the default template) AND the
+ *     handoff-style `oid` / `command_or_group` aliases (read by the
+ *     sbpp2026 template). They point at the same underlying DB
+ *     columns; the SmartyTemplateRule only checks top-level vars, so
+ *     row-level alias keys cost nothing.
+ *
+ * The wiring lives in `web/pages/admin.overrides.php`. The legacy
+ * `web/pages/admin.admins.php` `require`s that file at the bottom so
+ * the existing `?p=admin&c=admins` URL keeps its three-tab layout.
+ */
+final class AdminOverridesView extends View
+{
+    public const TEMPLATE = 'page_admin_overrides.tpl';
+
+    /**
+     * @param list<array<string,mixed>> $overrides_list Each entry is a
+     *     row from `:prefix_overrides` augmented with the handoff-style
+     *     aliases. Effective row shape:
+     *     `id` (int), `oid` (int, alias of `id`),
+     *     `type` ('command'|'group'),
+     *     `name` (string), `command_or_group` (string, alias of `name`),
+     *     `flags` (string).
+     */
+    public function __construct(
+        public readonly bool $permission_addadmin,
+        public readonly string $overrides_error,
+        public readonly bool $overrides_save_success,
+        public readonly array $overrides_list,
+    ) {
+    }
+}
--- a/web/pages/admin.admins.php
+++ b/web/pages/admin.admins.php
@@ -279,90 +279,9 @@ $serverscript .= "</script>";
     server_group_list: $server_group_list,
     server_script: $serverscript,
 ));
-// Overrides
 
-// Saving changed overrides
-$overrides_error        = "";
-$overrides_save_success = false;
-try {
-    if (isset($_POST['new_override_name'])) {
-        // Handle old overrides, if there are any.
-        if (isset($_POST['override_id'])) {
-            // Apply changes first
-            $edit_errors = "";
-            foreach ($_POST['override_id'] as $index => $id) {
-                // Skip invalid stuff?!
-                if ($_POST['override_type'][$index] != "command" && $_POST['override_type'][$index] != "group") {
-                    continue;
-                }
-
-                $id = (int) $id;
-                // Wants to delete this override?
-                if (empty($_POST['override_name'][$index])) {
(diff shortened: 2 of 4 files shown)


rumblefrog (commented):
Initial review. 0 blocking, 1 non-blocking; ready to merge once CI stays green.

## Checklist

| # | Check | Result |
| - | ----- | ------ |
| 1 | Scope vs B17 plan | OK — `AdminOverridesView` + `web/themes/sbpp2026/page_admin_overrides.tpl` + new `web/pages/admin.overrides.php`. Splitting overrides out of `admin.admins.php` (one-line `require()` bridge) is a sensible refactor that keeps `?p=admin&c=admins` URL parity. |
| 2 | Anti-patterns (xajax / ADOdb / inlined `:prefix_` / WYSIWYG / unannotated nofilter) | OK — only `nofilter` is the inline `$overrides_error` block which carries the canonical safety comment. |
| 3 | View DTO compliance + `SmartyTemplateRule` parity | OK — `AdminOverridesView` declares the four template vars; PHPStan green on both `default` and `sbpp2026` themes. |
| 4 | Permission booleans | OK — `permission_addadmin` matches the existing convention used by 7+ other Views (`AdminBansAddView`, `AdminServersAddView`, etc.). `Perms::for()` / `can_*` is the new canonical shape but, as the View docblock explains, dual-template binding requires keeping the legacy name until D1 cuts over the default theme. |
| 5 | Variable naming (dual-key `oid`/`id`, `command_or_group`/`name`) | OK — documented in the View docblock; `SmartyTemplateRule` only checks top-level vars so row-level aliases are free. |
| 6 | Testability hooks | OK — `data-testid=\"override-row\"` + `data-id={\$override.oid}`, `data-testid=\"addoverride-{type,name,flags}\"`, `overrides-error`/`overrides-success`/`overrides-table`/`overrides-form`/`overrides-save`/`overrides-back`. |
| 7 | JS conventions | N/A — only inline `javascript:history.go(-1)` for Back, mirrors the legacy `onclick=\"history.go(-1)\"` `{sb_button}`. |
| 8 | CSRF | OK — `{csrf_field}` present on the form. |
| 9 | API contract regen | N/A — no handler changes. |
| 10 | Snapshot tests | N/A. |
| 11 | PHPStan baseline | OK — empty diff vs `main`. |
| 12 | CI | OK — PHPStan (default + sbpp2026), PHPUnit, ts-check all green. |
| 13 | Behavioural correctness (edit / add / delete-by-blank / dup-name guards) | OK — verbatim move of the legacy block; the only deltas are `SELECT *` → `SELECT id` in the dup-check queries (sensible, no behaviour change). |

## Non-blocking nit

`admin.overrides.php` builds the duplicate-name `$edit_errors` line as `... name \\\"<name>\\\" from the selected type.<br />`. The `\\\"` literal is a JS-string-context escape inherited from the legacy `ShowBox(\"Error\", ...)` popup the default theme renders. The new sbpp2026 template emits `{\$overrides_error nofilter}` directly into HTML, so on a duplicate-name edit error the user sees a literal `\\\"foo\\\"` instead of `\"foo\"`. Cosmetic only — security is fine because `htmlspecialchars(addslashes(...))` neutralises the user-supplied portion either way, and it perfectly mirrors the verbatim-move invariant the PR claims. Worth a follow-up around D1 (when the legacy `ShowBox`-based default template is retired) to drop the JS-context escapes from the message string.

Approving on next sweep once nothing changes.

rumblefrog:
Rebased onto current `main` to resolve a merge conflict introduced by #1153 (B11). The conflict was localized to `web/pages/admin.admins.php`: B11 replaced the legacy `$theme->assign(...) + $theme->display('page_admin_admins_add.tpl')` block with a `\Sbpp\View\AdminAdminsAddView` render call, which sat in the same region B17 was deleting from. Resolution: keep B11's render call and append B17's `require(TEMPLATES_PATH . "/admin.overrides.php")` bridge after it. The other three PR files (View DTO, new page handler, sbpp2026 template) are byte-identical to the previous head. Old SHA `893fde24` → new SHA `744c97d1`.

Issue #1102, opened 2026-05-03
Login: `auth.login` checks the wrong config flag (`enablesteamlogin`)


rumblefrog:
## Problem

`web/api/handlers/auth.php:33` gates **normal-auth login** on `config.enablesteamlogin`:

```php
if (!Config::getBool('config.enablesteamlogin')) {
    Log::add('w', 'Hacking attempt', 'Attempted to connect via Normal Auth when the feature is disabled.');
    return Api::redirect('?p=login&m=failed');
}
```

That setting controls whether the **Steam login button** is rendered (`web/pages/page.login.php:106` → `$theme->assign('steamlogin_show', Config::getBool('config.enablesteamlogin'))`), not whether normal auth is enabled. So:

- An admin who disables Steam login (intending to keep normal auth as the only path) effectively disables **all** login.
- Every legitimate normal-auth attempt while the flag is off triggers a `"Hacking attempt"` audit-log entry, polluting the log.

This was carried over from xajax (`web/includes/sb-callback.php:110` in pre-#1096 history). It is **pre-existing** and was not introduced by #1096, but it was caught during the multi-agent review of that PR.

## Scope

The fix needs a real boolean for "is normal auth enabled". `data.sql` only ships `config.enablesteamlogin`; there is no `config.enablenormallogin` to read. So the change has three parts:

1. **Schema:** add `config.enablenormallogin` to `web/install/includes/sql/data.sql` defaulting to `'1'`. Add an updater step under `web/updater/data/` that inserts the row if missing.
2. **Settings UI:** add the corresponding checkbox to `web/pages/admin.settings.php` and `page_admin_settings_settings.tpl` (parallel to the existing `enable_steamlogin` toggle).
3. **Handler:** flip the check in `web/api/handlers/auth.php:33` to `config.enablenormallogin`. Drop the `Hacking attempt` log line — disabled-feature is not the same as a hacking attempt; either downgrade to `'m'` (info) or remove.

While here, consider the symmetric issue in `web/pages/page.login.php` — when only one login mode is enabled, the form should hide the disabled mode rather than appearing to accept input that will be rejected.

## Acceptance criteria

- [ ] `config.enablenormallogin` exists in fresh installs (struc/data.sql) and updates from older installs.
- [ ] Toggling `enable_steamlogin` no longer affects normal-auth login.
- [ ] Disabling normal auth blocks `auth.login` and the form hides/disables the username/password inputs.
[...]

PR #1149, opened 2026-05-05
[#1123 B8] Redesign lost password page

rumblefrog:
## Summary

Phase B (#1123) — restyle the public lost-password form to the
SourceBans++ 2026 visual language.

- Replaces the A1 stub at `web/themes/sbpp2026/page_lostpassword.tpl`
  with a real form built from the `handoff/pages/login.tpl`
  primitives (`.label` / `.input` / `.btn--primary` inside a `.card`).
  The form posts via `sb.api.call(Actions.AuthLostPassword, {email})`,
  matching the legacy default theme exactly, so the wire contract is
  unchanged and `auth.lost_password` API tests keep passing untouched.
- Adds `Sbpp\View\LostPasswordView` and switches
  `web/pages/page.lostpassword.php` from `$theme->display(...)` to
  `Sbpp\View\Renderer::render(...)` so `SmartyTemplateRule`'s
  dual-theme matrix (#1123 A2) guards both legs of the template.
  The View has no properties — the form is static and the JSON API
  carries the only state — so `LostPasswordView` exists primarily to
  bind the template to the typed pipeline.
- Response handling is inline (no `web/scripts/sourcebans.js`
  dependency, which the new theme intentionally drops). Both the
  success and error envelopes from `sb.api.call` are funneled into
  `window.SBPP.showToast(...)` exposed by the A1 `theme.js`.
- Testability hooks per the issue: `data-testid="lostpw-email"` on
  the email input, `data-testid="lostpw-submit"` on the submit
  button. `{csrf_field}` is included so a future POST fallback would
  still validate even though `sb.api.call` already attaches the
  `X-CSRF-Token` header from the chrome's `<meta name="csrf-token">`.
- Reset-link branch (`?email=&validation=`) keeps its legacy
  `<script>ShowBox(...)</script>` shape — that branch is out of
  scope for this Phase B form restyle and is reachable only from
  the recovery email anyway.

### Adjacent fix in the same PR

Once `LostPasswordView` binds to `page_lostpassword.tpl`,
`SmartyTemplateRule`'s default-theme leg also walks
`web/themes/default/page_lostpassword.tpl`, and its regex-based tag
scanner reads `sb.$id('email')` inside the legacy `{sb_button …
onclick="…"}` as a `{$id}` reference, firing
`sbpp.view.missingProperty` against a property the View intentionally
does not declare. The minimal fix is to swap the legacy template's
`sb.$id('email')` for `document.getElementById('email')` —
`sb.$id` is literally `(id) => document.getElementById(id)` in
`web/scripts/sb.js`, so the runtime behaviour is identical. A Smarty
comment block above the `{sb_button}` line explains why the helper
isn't used in that one spot. Bundling this with the View
introduction keeps the failure window at zero.

## Test plan

- [x] `./sbpp.sh phpstan` — clean, both default and `SBPP_PHPSTAN_THEME=sbpp2026` legs.
- [x] `./sbpp.sh composer api-contract` — no diff (action surface unchanged).
- [x] `./sbpp.sh test tests/api/AuthTest.php tests/api/PermissionMatrixTest.php tests/api/DispatcherTest.php tests/integration/LoginToggleTest.php` — 80 tests, 298 assertions, OK. Includes both `auth.lost_password` cases.
- [x] Manual: hit `http://localhost:8300/index.php?p=lostpassword` against the `sbpp2026` theme — markup contains the `lostpw-form`, `lostpw-email`, `lostpw-submit` IDs and `data-testid` attributes; `{csrf_field}` rendered.
- [ ] Reviewer: visual sanity check against `handoff/pages/login.tpl` token usage.

### DIFF
--- a/web/includes/View/LostPasswordView.php
+++ b/web/includes/View/LostPasswordView.php
@@ -1,0 +1,26 @@
+<?php
+declare(strict_types=1);
+
+namespace Sbpp\View;
+
+/**
+ * Lost-password page — binds to `page_lostpassword.tpl`.
+ *
+ * The form itself takes no PHP-side data: the recovery flow runs entirely
+ * over the JSON API (`Actions.AuthLostPassword`), so the template is a
+ * static card with a vanilla-JS submit handler. The View therefore declares
+ * no properties — it exists only so the page handler can switch from the
+ * untyped `$theme->display(...)` call to the typed
+ * {@see Renderer::render()} pipeline, and so SmartyTemplateRule can guard
+ * the template against accidentally growing untyped variable references.
+ *
+ * The reset-link branch in `web/pages/page.lostpassword.php` (the
+ * `?email=…&validation=…` URL the password-reset email links to) still
+ * uses the legacy inline `<script>ShowBox(...)</script>` + `PageDie()`
+ * shape and never reaches this View. That branch is out of scope for the
+ * Phase B form restyle; a follow-up moves it onto the new toast helper.
+ */
+final class LostPasswordView extends View
+{
+    public const TEMPLATE = 'page_lostpassword.tpl';
+}
--- a/web/pages/page.lostpassword.php
+++ b/web/pages/page.lostpassword.php
@@ -56,5 +56,5 @@ if (isset($_GET['email'], $_GET['validation']) && (!empty($_GET['email']) || !em
     print "<script>ShowBox('Password Reset', 'Your password has been reset and sent to your email.<br />Please check your spam folder too.<br />Please login using this password, <br />then use the change password link in Your Account.', 'blue');</script>";
     PageDie();
 } else {
-    $theme->display('page_lostpassword.tpl');
+    \Sbpp\View\Renderer::render($theme, new \Sbpp\View\LostPasswordView());
 }
--- a/web/themes/default/page_lostpassword.tpl
+++ b/web/themes/default/page_lostpassword.tpl
@@ -18,7 +18,12 @@
         </div>
 
         <div id="loginSubmit">
-            {sb_button text=Ok onclick="sb.api.call(Actions.AuthLostPassword, {email: sb.$id('email').value}).then(applyApiResponse);" class=ok id=alogin submit=false}
+            {* `sb.$id('email')` would otherwise be parsed by Smarty as a *}
+            {* `{$id}` reference inside this tag's onclick attribute and trip *}
+            {* SmartyTemplateRule once a paired View binds to this template; *}
+            {* `document.getElementById` is the literal body of `sb.$id` *}
+            {* (web/scripts/sb.js), so the runtime behaviour is unchanged. *}
+            {sb_button text=Ok onclick="sb.api.call(Actions.AuthLostPassword, {email: document.getElementById('email').value}).then(applyApiResponse);" class=ok id=alogin submit=false}
         </div>
 
     </div>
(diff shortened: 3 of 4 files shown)


rumblefrog (commented):
Initial review. 0 blocking, 0 non-blocking; ready to approve.

**Scope (✓)** — Three in-scope files (`LostPasswordView`, `page.lostpassword.php`, `themes/sbpp2026/page_lostpassword.tpl`) plus the call-out one-expression swap on `themes/default/page_lostpassword.tpl`.

**Out-of-strict-scope swap is the cleanest fix.** Verified against `web/scripts/sb.js` line 32 (`sb.$id = (id) => document.getElementById(id)` — runtime-identical for the string `'email'`). Verified the SmartyTemplateRule failure mode: the non-greedy `#\{([^}]*?)\}#s` tag scanner over the unescaped `{sb_button onclick="…{email: sb.$id('email').value}…"}` line would pick up `$id` as a referenced variable. Alternatives weighed:
- (a) **Fix the rule's regex to handle nested `{…}` in tag attributes** — Smarty grammar problem, out of scope for B8.
- (b) **Baseline silence** — forbidden by AGENTS.md.
- (c) **Declare an artificial `\$id` property on `LostPasswordView`** — pollutes the View AND would fail the parity check from the other end (the sbpp2026 template doesn't reference `\$id`, so `sbpp.view.unusedProperty` fires under `SBPP_PHPSTAN_THEME=sbpp2026`).
- (d) **Worker's choice — 1-expression swap** — minimal blast radius, runtime-equivalent, with an in-place 5-line `{* … *}` block explaining why the helper isn't used in this one spot. Accepted.

**Anti-patterns (✓)** — no xajax / sb-callback, no ADOdb, no MooTools / React / bundler, no `htmlspecialchars_decode`, no `nofilter`. Action name uses `Actions.AuthLostPassword` (not a string literal). Inline `<script>` is wrapped in `{literal}`, IIFE'd, `'use strict'`, vanilla.

**View DTO (✓)** — `LostPasswordView extends View`, declares `TEMPLATE = 'page_lostpassword.tpl'`, no public properties (the form is static and the JSON API carries the only state). The `SmartyTemplateRule` parity check passes because both the referenced and declared sets are empty.

**Permission checks (✓)** — public page; `web/pages/page.lostpassword.php` already gates the whole route on `Config::getBool('config.enablenormallogin')` (#1102).

**Testability hooks (✓)** — `data-testid="lostpw-email"` on the email input, `data-testid="lostpw-submit"` on the submit button.

**CSRF (✓)** — `{csrf_field}` rendered on the form per AGENTS.md "every state-changing form" rule, even though `sb.api.call` already attaches `X-CSRF-Token` from the chrome's `<meta name="csrf-token">`. Forward-compatible with a future POST fallback.

**Behavioural correctness (✓)** — verified the response-handling path against `web/api/handlers/auth.php` (`api_auth_lost_password` returns `{ message: { title, body, kind: 'blue' } }` on success, throws `ApiError` otherwise) and `web/scripts/api.js` (envelope is `{ ok, data?, error?, redirect? }`). The inline handler funnels success → `toast(mapKind('blue'), title, body)`, error → `toast('error', 'Error', error.message)`, and lets `sb.api.call`'s built-in `redirect` short-circuit handle navigation. The `mapKind` covers the legacy `red`/`green`/`blue` palette, and `'warn'` (used directly for the empty-email guard) is in the toast helper's accepted set per `theme.js:187`.

**API contract / snapshots / baseline (✓)** — handler signatures unchanged, no snapshot edits, no PHPStan baseline diff.

**CI (✓)** — all four checks green: PHPStan default, PHPStan sbpp2026, PHPUnit, ts-check.

Approving + merging.

PR #1152, opened 2026-05-05
[#1123 B2] Redesign public ban list (marquee)

rumblefrog:
Part of [#1123](https://github.com/sbpp/sourcebans-pp/issues/1123) — see the [v2.0.0 theme rollout plan](https://github.com/sbpp/sourcebans-pp/blob/main/handoff/HANDOFF.md) (B2: Public ban list — MARQUEE page).

The most-visited public surface gets the marquee redesign: handoff-spec rows with avatar/state/pill, sticky filter chips, mobile card list, and the testability hooks the #1123 issue locks in. Both themes render off the same `BanListView` DTO, so the legacy `default/` theme keeps working until D1 cuts over.

## Screenshots

> Screenshots are committed to a sibling orphan branch (`screenshots/1123-b2-banlist`) so they render inline here without polluting the feature branch's diff. The branch can be deleted with the PR after merge.

**Light mode (desktop, 1440×900):**
banlist-light

**Dark mode (desktop, 1440×900):**
banlist-dark

**Mobile cards (414×800, 2× DPR):**
banlist-mobile

The light/dark screenshots show the four state derivations the marquee design distinguishes:
- `permanent` — purple pill, magenta-pink left border (`wallhack_2024`, `angry_pyro_42`)
- `active` — orange pill, orange left border (`snipezzz`, `griefer.tv`)
- `expired` — gray pill, gray left border (`spammer`, `old_player_2018`, `stale_account`)
- `unbanned` — teal pill, teal left border (`reformed_player`)

State derivation lives in `page.banlist.php` (collapses `banlength`, `ban_ends`, `RemoveType` into one of the four labels), so the template just gates classes off `{$ban.state}`.

## Files touched

| File | What changed |
| --- | --- |
| `web/themes/sbpp2026/page_bans.tpl` | Replaces the A1 stub with the 134-line handoff design. Desktop table + mobile card list both expose `data-testid="ban-row"`, `data-id="{$ban.bid}"`, and `data-drawer-href` so the C1 drawer can layer on without a per-theme view-model detour. Sticky filter chips carry `filter-chip-*`, per-row actions carry `row-action-edit` / `row-action-unban`, the search input carries `bans-search`. The trailing `{if false}…{/if}` block is the dual-theme PHPStan manifest for legacy-only View props. |
| `web/includes/View/BanListView.php` (new) | Paired View DTO. Carries every variable both themes consume — handoff names for sbpp2026 (`bid|name|state|length_human|banned_human|sname|can_edit_ban|can_unban`) plus the legacy keys for `default/` (`ban_id|player|class|reban_link|edit_link|…`) — and the comment-edit scratch pad. SmartyTemplateRule does not introspect array contents, so the per-row dual shape coexists on `$ban_list` items without conflict. |
| `web/pages/page.banlist.php` | Switches to `Renderer::render($theme, new BanListView(...))`. Per-row payload gains `bid|name|steam|state|length_human|banned_human|sname|can_edit_ban|can_unban|avatar_initials|avatar_hue` alongside the legacy keys. State derivation collapses `(banlength, ban_ends, RemoveType)` into `permanent|active|expired|unbanned` server-side. Pagination markup is rebuilt to embed `data-testid="page-prev"` / `page-next` and uses inline `window.location` JS for the page-jump select (the new theme drops `sourcebans.js` in #1123 D1, so any reach into legacy bulk JS would silently break here). DB queries and the unban POST handler are unchanged. |
| `web/scripts/banlist.js` (new) | Optional client layer: chip filter + URL state sync, copy-to-clipboard via `theme.js`'s toast helper, and the comment-edit form POSTed through `sb.api.call(Actions.Bans*)` instead of the legacy form round-trip. Vanilla JS, `// @ts-check`, feature-detected so a missing element doesn't throw. The legacy theme keeps its server-side comment POST. |
| `web/phpstan-baseline.neon` | Drops the now-stale `Variable $prev might not be defined` entry for `pages/page.banlist.php`; the rebuilt pagination always initialises `$prev`. |

## Conventions

- **Permissions** are pre-computed on the View (`bool can_edit_ban`/`can_unban` per row + top-level `general_unban`/`can_delete`/`can_export`) — the template never reaches into `$user.srv_flags|...`. The Add-ban button is gated with `{has_access flag=$smarty.const.ADMIN_ADD_BAN}` (A3 plugin).
- **Testability hooks** ship for every entry in the marquee table from #1123: `filter-chip-{permanent,active,expired,unbanned}`, `ban-row` + `data-id="{bid}"`, `row-action-{edit,unban}`, `drawer-trigger` + `data-drawer-href`, `bans-search`, `page-{prev,next}`, `[data-loading]` / `[data-skeleton]`, `<time datetime="…">` for the banned timestamp.
- **`nofilter` discipline**: every `{$… nofilter}` in the template carries the canonical `{* nofilter: <reason> *}` comment immediately above it.
- **SQL queries unchanged** — touching them was explicitly out of scope. The `:prefix_` literal pattern is preserved.

## Local gates

- `./sbpp.sh phpstan` — green for both `default` and `sbpp2026` themes
- `./sbpp.sh test` — 268 tests, 1134 assertions, OK
- `./sbpp.sh ts-check` — green
- `./sbpp.sh composer api-contract` — clean (no diff)

### DIFF
--- a/web/scripts/banlist.js
+++ b/web/scripts/banlist.js
@@ -81,13 +81,18 @@
       const sb = /** @type {any} */ (window).sb;
       const Actions = /** @type {any} */ (window).Actions;
       if (!sb || !sb.api || !Actions) {
-        cform.removeEventListener('submit', /** @type {EventListener} */ (() => {}));
+        // sb / api-contract failed to load (offline, asset 404, …).
+        // Native form.submit() bypasses listeners per the HTML spec, so
+        // we don't need to detach this handler first; the submission
+        // POSTs to the action-less form URL, which page.banlist.php
+        // doesn't handle, but that's the same fall-through the legacy
+        // theme has when sourcebans.js fails to load.
         cform.submit();
         return;
       }
 
       const action = cid > 0 ? Actions.BansEditComment : Actions.BansAddComment;
-      sb.api.call(action, { bid: bid, cid: cid, ctype: ctype, text: value })
+      sb.api.call(action, { bid: bid, cid: cid, ctype: ctype, ctext: value, page: page })
         .then(() => {
           window.location.href = 'index.php?p=banlist' + (page > 0 ? '&page=' + page : '');
         })
--- a/web/scripts/banlist.js
+++ b/web/scripts/banlist.js
@@ -1,0 +1,116 @@
+// @ts-check
+/* ============================================================
+   banlist.js — sbpp2026 public ban list interactions
+
+   Layered on top of the server-rendered table from
+   themes/sbpp2026/page_bans.tpl. The template is fully usable
+   without this script (filters fall through as a server-side
+   ?searchText= query, copy buttons no-op, comment-edit is a
+   normal POST round-trip), so this file only adds:
+
+     1. Status-filter chips   — toggle row visibility client-side,
+        sync the active state into the URL hash so deep links
+        survive a refresh.
+     2. SteamID copy buttons  — the {has_access}-gated copy is
+        already inert without JS; this just wires
+        navigator.clipboard + the toast helper from theme.js.
+     3. Comment edit form     — POSTs through sb.api.call()
+        instead of the legacy `<form action="">` so the new
+        theme doesn't need to navigate away.
+
+   Each interaction is wrapped in a feature-detect (`if (foo)`)
+   so a missing element on the comment-only branch doesn't throw.
+   ============================================================ */
+(function () {
+  'use strict';
+
+  /** @type {HTMLElement | null} */
+  const root = /** @type {HTMLElement | null} */ (document.getElementById('banlist-root'));
+
+  // ---- STATUS FILTER CHIPS ---------------------------------
+  /** @type {NodeListOf<HTMLButtonElement>} */
+  const chips = document.querySelectorAll('[data-state-filter]');
+  /** @type {NodeListOf<HTMLElement>} */
+  const rows = document.querySelectorAll('.ban-row[data-state]');
+
+  /**
+   * @param {string} state '' to show all, otherwise one of permanent|active|expired|unbanned.
+   * @returns {void}
+   */
+  function applyStateFilter(state) {
+    rows.forEach((r) => {
+      const match = !state || r.dataset.state === state;
+      r.style.display = match ? '' : 'none';
+    });
+    chips.forEach((c) => {
+      c.setAttribute('aria-pressed', c.dataset.stateFilter === state ? 'true' : 'false');
+    });
+    try {
+      const url = new URL(window.location.href);
+      if (state) url.searchParams.set('state', state); else url.searchParams.delete('state');
+      window.history.replaceState({}, '', url.toString());
+    } catch (e) { /* URL unsupported in some old browsers; chip filter still works */ }
+  }
+
+  chips.forEach((c) => {
+    c.addEventListener('click', () => applyStateFilter(c.dataset.stateFilter || ''));
+  });
+
+  try {
+    const initial = new URL(window.location.href).searchParams.get('state') || '';
+    if (initial) applyStateFilter(initial);
+  } catch (e) { /* ignore — default "All" is already applied server-side */ }
+
+  // ---- COMMENT EDIT FORM (sbpp2026 only) -------------------
+  // Legacy theme keeps the `<form method=post>` round-trip; the
+  // new theme's form has no `action`, so submit hits this handler
+  // and goes through the JSON API. Falls back to navigation on
+  // non-OK responses.
+  /** @type {HTMLFormElement | null} */
+  const cform = /** @type {HTMLFormElement | null} */ (document.getElementById('banlist-comment-form'));
+  if (cform) {
+    cform.addEventListener('submit', (/** @type {SubmitEvent} */ e) => {
+      e.preventDefault();
+      const bid = parseInt(cform.dataset.bid || '0', 10);
+      const cid = parseInt(cform.dataset.cid || '0', 10);
+      const ctype = cform.dataset.ctype || 'B';
+      const page = parseInt(cform.dataset.page || '-1', 10);
+      const text = /** @type {HTMLTextAreaElement | null} */ (document.getElementById('banlist-comment-text'));
+      const value = text ? text.value : '';
+
+      const sb = /** @type {any} */ (window).sb;
+      const Actions = /** @type {any} */ (window).Actions;
+      if (!sb || !sb.api || !Actions) {
+        cform.removeEventListener('submit', /** @type {EventListener} */ (() => {}));
+        cform.submit();
+        return;
+      }
+
+      const action = cid > 0 ? Actions.BansEditComment : Actions.BansAddComment;
+      sb.api.call(action, { bid: bid, cid: cid, ctype: ctype, text: value })
+        .then(() => {
+          window.location.href = 'index.php?p=banlist' + (page > 0 ? '&page=' + page : '');
+        })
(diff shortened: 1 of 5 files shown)


web/themes/sbpp2026/page_bans.tpl
```diff
@@ -1,6 +1,290 @@
...
+     filter triggers a re-render. Kept here so the testid hook the
+     marquee issue locks in (`[data-skeleton]`) is always present in
+     the DOM, even when the live list is populated. *}
+<div data-skeleton hidden aria-hidden="true">
+  <div class="skel" style="height:2.5rem;margin-bottom:0.5rem"></div>
+  <div class="skel" style="height:2.5rem;margin-bottom:0.5rem"></div>
+  <div class="skel" style="height:2.5rem"></div>
 </div>
+
+<script src="./scripts/banlist.js" defer></script>
```
rumblefrog, line 279: **BLOCKING.** The `<script src="./scripts/banlist.js" defer>` tag lives inside the `{else}` branch (the listing view), so it is **not loaded when `$comment` is truthy** (the comment add/edit view at lines 23-60).

Effect: opening `?p=banlist&comment=N&ctype=B` (Add comment) or `&cid=M` (Edit comment) renders the form at line 32, but `banlist.js`'s submit handler never attaches. The `<button type="submit">` at line 37 then triggers a native form submission to the current URL with no `action`, and `page.banlist.php` has no POST handler — the comment save is silently dropped.

Fix: move `<script src="./scripts/banlist.js" defer></script>` outside the `{if $comment}{else}{/if}` block (e.g. after the closing `{/if}` on line 272, or before `{if $comment}` on line 23). The IIFE in `banlist.js` already feature-detects `#banlist-comment-form` and `[data-state-filter]`, so it's safe to load in either mode.

web/scripts/banlist.js
```diff
@@ -0,0 +1,116 @@
...
+      const sb = /** @type {any} */ (window).sb;
+      const Actions = /** @type {any} */ (window).Actions;
+      if (!sb || !sb.api || !Actions) {
+        cform.removeEventListener('submit', /** @type {EventListener} */ (() => {}));
+        cform.submit();
+        return;
+      }
+
+      const action = cid > 0 ? Actions.BansEditComment : Actions.BansAddComment;
+      sb.api.call(action, { bid: bid, cid: cid, ctype: ctype, text: value })
```
rumblefrog: **BLOCKING.** Param-name mismatch with the JSON API. `api_bans_add_comment` / `api_bans_edit_comment` read `(string)(\$params['ctext'] ?? '')` (`web/api/handlers/bans.php:224,261`), but this call sends `text:`. The HTTP call succeeds; the comment body is silently saved as `''`.

For reference, the legacy caller in `web/scripts/sourcebans.js:872` (`ProcessComment()`) correctly sends `ctext:` and `page:`. Mirror that:

```js
sb.api.call(action, { bid: bid, cid: cid, ctype: ctype, ctext: value, page: page })
```

Including `page` here also fixes the redirect: the handler reads `$page` and builds `?p=banlist&page=N` from it, so without it the user is bounced back to page 1 every time after editing a comment from a deeper page.

web/scripts/banlist.js
```diff
@@ -0,0 +1,116 @@
...
+      const cid = parseInt(cform.dataset.cid || '0', 10);
+      const ctype = cform.dataset.ctype || 'B';
+      const page = parseInt(cform.dataset.page || '-1', 10);
+      const text = /** @type {HTMLTextAreaElement | null} */ (document.getElementById('banlist-comment-text'));
+      const value = text ? text.value : '';
+
+      const sb = /** @type {any} */ (window).sb;
+      const Actions = /** @type {any} */ (window).Actions;
+      if (!sb || !sb.api || !Actions) {
+        cform.removeEventListener('submit', /** @type {EventListener} */ (() => {}));
```
rumblefrog: **Nit (non-blocking).** This `removeEventListener` call is a no-op — the function reference passed in (`() => {}`) is a fresh arrow function, not the actual handler registered at line 72, so nothing matches and nothing is removed. The native `cform.submit()` on the next line already bypasses the submit-event chain per the HTML spec, so the listener doesn't need to be removed at all. Drop the line.

rumblefrog (commented):
Initial review pass on B2 (marquee public ban list).

**2 blocking, 1 non-blocking; addressing now.**

The marquee redesign is in great shape — testability hooks are all present (`filter-chip-*`, `ban-row` + `data-id`, `row-action-*`, `drawer-trigger`, `bans-search`, `page-prev`/`page-next`, `[data-loading]`, `[data-skeleton]`, `<time datetime>`), state derivation correctly distinguishes natural-expiry from admin-driven unban (matches the design's separate pills), the View DTO's dual-shape strategy with the `{if false}` PHPStan manifest is well-documented, `nofilter` discipline holds, and the legacy `default/` theme is untouched so dual-theme PHPStan stays green. Drawer-href markup is compatible with C1's reader (`?p=banlist&c=details&id={bid}` matches the handoff).

Both blockers cluster in the comment-edit path — banlist.js and the template never agree on whether the script is even loaded for that view, and even if it were, the param name sent to the JSON API would silently drop the comment text. Details inline.

Full checklist:
- Scope: ✅ four files match the plan list (template, View DTO, page handler, optional `web/scripts/banlist.js` is plan-authorized)
- Anti-patterns: ✅ no jQuery/MooTools/React, vanilla JS with `// @ts-check`, no `Actions` string literals, `:prefix_` preserved, no `htmlspecialchars_decode` on JSON params, no install/sql edits, no WYSIWYG
- View DTO + SmartyTemplateRule: ✅ dual-theme PHPStan green
- Permissions: ✅ pre-computed `can_edit_ban`/`can_unban` per row + top-level `can_export`/`general_unban`/`can_delete`/`view_bans`; `{has_access}` for ad-hoc Add Ban gate
- Variable naming: ✅ handoff names (`bid|name|state|length_human|banned_human|sname`) — `class` correctly replaced with `state`
- Testability hooks: ✅ all marquee-locked surfaces present, `data-id` is integer bid (not formatted timestamp)
- CSRF: ✅ N/A for unban (preserved existing GET+postkey pattern, out of scope); comment form goes via `sb.api.call` which auto-sets `X-CSRF-Token`
- API contract: ✅ N/A (no handler change)
- Snapshots: ✅ N/A
- PHPStan baseline: ✅ single removal is the resolved-by-pagination-rewrite entry; no additions
- CI: ✅ all four checks green
- Behavioural: ✅ state derivation traces correctly through all four branches; View constructor mapping verified for `bid`/`name`/`steam`/`length_human`/`banned_human`/`sname`/`state`

web/themes/sbpp2026/page_bans.tpl
rumblefrog: Addressed in ca380303 — moved `<script src="./scripts/banlist.js" defer>` outside the `{if $comment}{else}{/if}` block (now sits between `{/if}` and the EOF PHPStan manifest) so the comment-edit branch also gets `banlist.js`.

web/scripts/banlist.js
rumblefrog: Addressed in ca380303 — `text:` -> `ctext:` and added `page: page` to preserve the active pagination on redirect. Now matches the `ProcessComment()` call shape from `web/scripts/sourcebans.js`.

web/scripts/banlist.js
rumblefrog: Addressed in ca380303 — dropped the dead removeEventListener call and replaced it with a comment explaining the form-submit fall-through path (matches the legacy theme's behaviour when sourcebans.js fails to load).

PR #1154, opened 2026-05-05
[#1123 B5] Redesign public servers list

rumblefrog:
Part of #1123 (Phase B, ticket B5).

## Summary

Replaces the A1 stub at `web/themes/sbpp2026/page_servers.tpl` with a
responsive card grid sourced from `handoff/pages/servers.tpl`. Pairs
with a docblock-only extension to `Sbpp\View\ServersView` and a
one-liner enrichment to `web/pages/page.servers.php` so the template
has a `mod` display name to render before the live UDP query lands.

- **Card grid** (auto-fill `minmax(20rem, 1fr)`) — one card per
  configured server, mod icon + hostname + ip:port + status pill +
  map + players row + Connect/Players/Refresh actions. Inline JS
  walks every `[data-testid="server-tile"]` and fires
  `sb.api.call(Actions.ServersHostPlayers, ...)` per card to
  progressively populate live data; mirrors the legacy
  `LoadServerHost(...)` flow without depending on
  `web/scripts/sourcebans.js` (which D1 deletes).
- **Live state** — each tile starts at `data-status="loading"` and
  flips to `online` or `offline` based on the handler's
  `{ error: 'connect' }` envelope. Status lives on the
  `data-status` attribute (per #1123 "state in attributes, not
  styling"); the same hook drives the loading/online/offline pill.
- **Auto-expand** — `?p=servers&s={index}` keeps working: the matching
  card auto-expands its player panel once the live response lands.
  No additional fetch — the same `Actions.ServersHostPlayers`
  response already includes `player_list[]`.
- **Permissions** — `$access_bans` (legacy field, unchanged) gates
  the right-click hint at the top of the page. Per-server admin
  shortcuts (RCON, Players-as-admin) from the handoff are
  intentionally cut — they live on the admin servers list (B15).
- **Empty state** — friendly card when no servers are configured.

## Files

- `web/themes/sbpp2026/page_servers.tpl` — replaces A1 stub.
- `web/includes/View/ServersView.php` — docblock-only extension; no
  property changes (preserves the parity B3's `HomeDashboardView`
  relies on while reusing this view's `$server_list`).
- `web/pages/page.servers.php` — adds `md.name AS mod_name` to the
  SELECT and exposes it as `$info['mod']` for the new card label.
  `Renderer::render` was already in place from earlier theme
  groundwork.

## Hard rules

- Touched only the 3 in-scope files.
- No edits to chrome (`core/*`), `init.php`, `api/handlers/*`,
  `api-contract.js`, `AGENTS.md`, or `ARCHITECTURE.md`.
- Variable naming follows the handoff's `$server.{sid,name,host,port,
  players,map,mod,...}` shape; new keys added, none renamed/removed
  (B3's dashboard transitively reads `$server_list` row entries).
- No `{nofilter}` introduced.
- Vanilla JS only (`{literal}` block; relies on the chrome's
  `api-contract.js` + `sb.js` + `api.js`).
- All SQL via `:prefix_` literal placeholder; no inline prefix.
- Testability hooks: `data-testid="servers-list"`,
  `data-testid="server-tile"`, `data-id={sid}`, `data-index={index}`,
  `data-status="loading|online|offline"`, plus per-element ids on
  status / host / map / players / panel / toggle / connect /
  refresh.
- Source query data handled gracefully when offline — preserves
  legacy behaviour (UDP probe may time out; row falls back to
  Offline pill).

## Plan link

#1123 issue body, Phase B, ticket B5 ("public servers").

## Local gates

| Gate | Result |
| --- | --- |
| `./sbpp.sh phpstan` (default theme leg) | Same 4 baseline-mismatch errors as clean `main` (pre-existing). No new errors from this PR. |
| `SBPP_PHPSTAN_THEME=sbpp2026 ... phpstan-sbpp2026.neon` | Same 3 pre-existing `Database` constructor type errors as clean `main` (`init.php`, `CUserManager.php`, `install/template/page.6.php`). No new errors from this PR. |
| `./sbpp.sh ts-check` | Clean. |
| `./sbpp.sh test --testsuite api` | 238 tests, 874 assertions, all green. |
| `./sbpp.sh composer api-contract` | No diff. |

## Screenshots

Hosted on the disposable orphan branch `screenshots-b5-1154` (never
merged). Light-mode initial render, with the inline
`sb.api.call(Actions.ServersHostPlayers, ...)` probe still in flight:

Public servers — initial loading state

After the probe lands, with the API mocked to return online responses
for three of the four configured servers (the fourth's mock returned
`{ error: 'connect' }`, demonstrating the offline pill and the
disabled Players button):

Public servers — online state with players + map populated

Same online state with the first card's player panel expanded (clicking
the Players toggle reveals the cached `player_list[]` from the same
API response — no extra fetch):

Public servers — expanded player panel

Real offline state (every configured server is unreachable, so every
card flips to Offline; Players button stays disabled, Connect link
stays valid since `steam://` resolution is the client's job):

Public servers — all servers offline

Dark mode — `localStorage.sbpp-theme = 'dark'` applied via the topbar
theme toggle (the Connect button rendering as neutral instead of brand
in dark mode is a CSS-specificity quirk inherited from the A1 vendored
`theme.css`, not something this PR introduces):

Public servers — dark mode

Mobile width (~420px) — sidebar collapses, cards stack to a single
column:

Public servers — mobile width

Empty state — no enabled servers configured:

Public servers — empty state

## Test plan

- [ ] Visit `?p=servers` with no servers configured → "No servers are
      configured yet." card renders.
- [ ] Visit `?p=servers` with several servers → grid of cards, each
      starting in loading then degrading to offline (for unreachable
      hosts) or online with hostname + map + player count.
- [ ] Click a card's Players button on an online server → expands and
      lists players (name + frags + time) inline.
- [ ] Click Refresh on a card → re-fires the live query, status flips
      back to loading then resolves.
- [ ] `?p=servers&s={index}` deep-links → card with that index
      auto-expands its player panel after the live response.
- [ ] Mobile width (~360px–420px) → cards stack to single column,
      sidebar collapses.
- [ ] Toggle dark mode (theme toggle in topbar) → contrast / borders
      hold up.

### DIFF
--- a/web/themes/sbpp2026/page_servers.tpl
+++ b/web/themes/sbpp2026/page_servers.tpl
@@ -1,6 +1,427 @@
-<div class="card">
-    <div class="card__body">
-        <h1>{$title}</h1>
-        <p class="text-muted">This page hasn't been redesigned yet.</p>
+{*
+    SourceBans++ 2026 — page_servers.tpl
+
+    Public servers list. Replaces the A1 stub. Pair:
+    web/pages/page.servers.php + web/includes/View/ServersView.php
+    (#1123 B5).
+
+    Why this differs from handoff/pages/servers.tpl
+    -----------------------------------------------
+    The handoff card grid is treated as the visual reference for the
+    layout shell (mod-letter avatar, hostname + ip:port, status pill,
+    players bar, map row), not as a 1:1 markup port:
+
+      * The handoff renders an "Add server" button and per-row "RCON" /
+        "Players" admin shortcuts gated on SourceMod char-flags
+        (`$user.srv_flags|strpos:'z'`). The legacy public servers page
+        carries neither — both belong on the admin servers list page
+        (#1123 B15). Reproducing them here would require Perms / SM
+        char-flag plumbing the public View doesn't need today, so
+        they're intentionally cut.
+      * The handoff's status (online/offline) is set server-side from a
+        precomputed `$s.online`. We don't have that cheaply: each card
+        starts in a loading state and progressively flips to
+        online/offline as the inline `sb.api.call(
+        Actions.ServersHostPlayers, …)` per-card response lands. This
+        mirrors the legacy `LoadServerHost(...)` UDP-poll flow.
+
+    Live data flow
+    --------------
+    SourceQuery (xpaw/php-source-query-class) is a UDP probe — the
+    server may be down, behind NAT, or rate-limiting. The handler
+    returns `{ error: 'connect', ip, port, is_owner }` on failure
+    instead of throwing, so each card has to handle three terminal
+    states: `loading` (initial), `online` (data came back clean),
+    `offline` (handler returned `error: 'connect'` or the call itself
+    failed). Status is persisted in the `data-status` attribute so
+    Playwright (and humans) can assert state without reading computed
+    CSS — see #1123 issue body's "State in attributes, not just
+    styling" rule.
+
+    Auto-expand
+    -----------
+    Mirrors the legacy `?p=servers&s={index}` pattern: when the page
+    handler sets `opened_server >= 0`, the matching card auto-expands
+    its player list once the live response lands. The expansion fetches
+    no additional data — the same `Actions.ServersHostPlayers`
+    response already includes `player_list[]`.
+
+    Per-row JS without ./scripts/sourcebans.js
+    ------------------------------------------
+    sourcebans.js is dropped at #1123 D1, so we can't use the legacy
+    `LoadServerHost()` / `InitAccordion()` helpers; they'd be undefined
+    here. The card markup carries `data-sid` / `data-index` /
+    `data-status` and the inline `<script>` walks
+    `[data-testid="server-tile"]` directly. `sb.api.call` and the
+    `Actions.*` constants come from the chrome's
+    api.js + api-contract.js (loaded by core/header.tpl).
+
+    Testability hooks (per #1123 issue + B5 brief)
+    ----------------------------------------------
+      data-testid="servers-list"   — outer container
+      data-testid="server-tile"    — each card (one per server)
+      data-testid="server-toggle"  — expand/collapse trigger
+      data-testid="server-connect" — steam:// connect button
+      data-id=$server.sid          — stable integer (NOT the index)
+      data-status="loading|online|offline" — terminal state on each card
+      data-index=$server.index     — 0-based offset for the dashboard
+                                     ?p=servers&s=N deep link
+*}
+<div class="p-6 space-y-4" style="max-width:1400px;margin:0 auto;width:100%">
+
+    <header class="flex items-end justify-between gap-3" style="flex-wrap:wrap">
+        <div>
+            <h1 style="font-size:var(--fs-2xl);font-weight:600;margin:0">Servers</h1>
+            <p class="text-sm text-muted m-0 mt-2"
+               data-testid="servers-summary"
+               data-total="{$server_list|@count}">
+                {$server_list|@count} configured
+                <span data-online-count>&middot; <span data-online-num>0</span> online</span>
+            </p>
+        </div>
+        {if $IN_SERVERS_PAGE && $access_bans}
+            <p class="text-xs text-muted m-0"
+               role="note"
+               data-testid="servers-rcon-hint"
+               style="max-width:24rem;text-align:right">
+                <i data-lucide="info" style="width:12px;height:12px;vertical-align:-2px"></i>
+                Right-click a player on an expanded card to kick, ban, or message them.
+            </p>
+        {/if}
+    </header>
+
+    {if $server_list|@count == 0}
+        <div class="card" data-testid="servers-empty">
+            <div class="card__body text-sm text-muted">
+                No servers are configured yet.
+            </div>
+        </div>
+    {else}
+    <div class="grid gap-4"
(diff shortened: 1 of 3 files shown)


rumblefrog (commented):
Initial review.

CI is fully GREEN across all four legs (PHPStan default + sbpp2026, PHPUnit, ts-check). Verified the Worker's "pre-existing PHPStan errors" claim is correct in the only sense that matters — they're all baseline-managed (`reportUnmatchedIgnoredErrors: false` on the sbpp2026 leg per `phpstan-sbpp2026.neon`), so CI passes cleanly. PR diff stays inside the three in-scope files (`web/includes/View/ServersView.php`, `web/pages/page.servers.php`, `web/themes/sbpp2026/page_servers.tpl`); no chrome / api / baseline edits.

View / DTO + B3 parity: confirmed — extension is **docblock-only** (no public property added or removed), `HomeDashboardView::$server_list` is typed `list<array<string,mixed>>` so the new `mod` row key is transparently forwarded into the dashboard's existing `$serversView->server_list` plumbing in `web/pages/page.home.php:212`, and the legacy `web/themes/default/page_servers.tpl` ignores extra row keys. B3's dashboard tile is unaffected.

Other checklist items confirmed clean: `$access_bans` correctly gates the right-click hint only; testability hooks (`data-testid="servers-list" / "server-tile" / "server-status" / "server-toggle" / "server-connect" / "server-refresh"`, `data-id={sid}`, `data-status="loading|online|offline"`, `data-index={index}`) are all present; inline `<script>` is wrapped in `{literal}` so SmartyTemplateRule strips it before scanning and ts-check (which only walks `web/scripts/*.js`) doesn't touch it; the `servers.host_players` handler is registered `public=true, perm=0` so the unauthenticated public page can call it; `error: 'connect'` envelope is handled with the disabled Players button + Offline pill.

0 blocking, 2 non-blocking; both about the Lucide icon-swap path inside `setStatus()`. Reply on the PR shortly.

web/themes/sbpp2026/page_servers.tpl
```diff
@@ -1,6 +1,427 @@
...
+            pill.classList.add(status === 'online' ? 'pill--online' : 'pill--offline');
+            var lbl = pill.querySelector('[data-status-label]');
+            if (lbl) lbl.textContent = label;
+            var icon = pill.querySelector('i');
+            if (icon) {
+                icon.setAttribute('data-lucide',
+                    status === 'online' ? 'check-circle-2'
+                    : status === 'offline' ? 'x-circle'
+                    : 'loader');
+            }
```
rumblefrog, line 268: **Non-blocking — visual icon never updates after the API response.**

Lucide's `createIcons()` (called by `theme.js` after the chrome's `core/footer.tpl` parses `lucide.min.js`) does `t.parentNode.replaceChild(_$, t)` on every `<i data-lucide="…">` (verified at the only `replaceChild` callsite in the vendored `lucide.min.js`, in the `x0` function). After that pass, the original `<i>` element is gone — replaced by an `<svg data-lucide="loader">`.

The API call is async, so its `.then` handler always runs after Lucide's first pass. By the time `setStatus()` runs, `pill.querySelector('i')` returns `null`, the `if (icon)` branch is skipped, and the loader spinner stays in the pill forever — even on `online` / `offline` cards. The pill's text label and color (`pill--online` / `pill--offline`) update fine, so users still distinguish the states; the icon is the only thing stuck.

Fwiw the screenshots in the PR body are too small to tell whether the swap is silently broken or whether I'm wrong about Lucide's behavior here, so first step is probably to load a real online server in the dev stack and inspect what's actually in the pill after the response. If it is broken, two cheap fixes:

1. Switch the selector to `pill.querySelector('[data-lucide]')` so it finds the swapped `<svg>`, then call `window.lucide.createIcons()` inside `setStatus()` after the swap so Lucide re-renders just the new attribute.
2. Or keep the selector but skip Lucide for the status icon entirely — render a tiny inline SVG triple (loader / check-circle-2 / x-circle) directly inside `setStatus()`, bypassing the swap-then-rewrite dance.

Not a merge blocker — basic state communication works — but worth a follow-up either way.

web/themes/sbpp2026/page_servers.tpl
```diff
@@ -1,6 +1,427 @@
...
+            refresh.addEventListener('click', function (ev) {
+                ev.preventDefault();
+                loadTile(tile);
+            });
+        }
+    });
+
+    // Re-render Lucide icons we just swapped via setAttribute('data-lucide').
+    if (typeof window.lucide !== 'undefined' && typeof window.lucide.createIcons === 'function') {
+        window.lucide.createIcons();
```
rumblefrog, line 423: **Non-blocking — this `lucide.createIcons()` call is unreachable / dead.**

The IIFE runs synchronously when the `<script>` tag in `{$page_content}` is parsed, which is **before** `core/footer.tpl` later in the document loads `lucide.min.js` and `theme.js`. At this exact line, `window.lucide` is still `undefined`, so the `typeof` guard short-circuits and the body never runs. (The comment above it talks about "icons we just swapped via setAttribute" but the synchronous IIFE doesn't swap anything — the swap happens later, inside the `.then` of `sb.api.call`.)

If the icon-swap path in `setStatus()` ends up needing `lucide.createIcons()` (per the other thread above), it'd live there, not at the bottom of the IIFE. Either way this block can be deleted as a no-op cleanup.

PR #1155, opened 2026-05-05
[#1123 B1] Redesign login page

rumblefrog:
Part of #1123. Plan: [`sbpp2026-theme-rollout-4c702e97.plan.md`](../tree/main/.cursor/plans/sbpp2026-theme-rollout-4c702e97.plan.md) — phase **B1 (Login page)**.

Replaces the A1 stub at `web/themes/sbpp2026/page_login.tpl` with the sbpp2026 login redesign and routes the page handler through the typed View DTO pattern.

## Highlights

- Centred sign-in card matching the new chrome. The handoff demo's full-viewport 2-column layout is dropped here because `web/includes/page-builder.php` mandatorily wraps every page in `core/{header,navbar,title,footer}.tpl` (added in #1123 A2) — a full-viewport split would collide with the chrome. The B1 dispatch explicitly allows this trim.
- Vanilla JS form posts via `sb.api.call(Actions.AuthLogin, …)` — no string action name, no jQuery, no MooTools. `redirect: ''` lands on `'?'` after success (= dashboard) via `Api::redirect('?' . $redirect)` in `web/api/handlers/auth.php`.
- `?m=…` status messages are re-routed from the legacy `ShowBox()` JS dialogs (which depend on the about-to-be-deleted `web/scripts/sourcebans.js`) to `window.SBPP.showToast(…)` driven client-side off `URLSearchParams`. The page handler still emits the legacy `ShowBox()` blocks for the default theme, guarded with `if (typeof ShowBox === 'function')` so they no-op cleanly when the sbpp2026 chrome is active.
- Stable `data-testid` hooks per the B1 dispatch:
  `login-form`, `login-username`, `login-password`, `login-remember`, `login-submit`, `login-steam`, `login-disabled`, `login-lostpwd`.
- `{csrf_field}` on the form (`CSRF::validate` runs in `route()` for POSTs and `sb.api.call` sends `X-CSRF-Token` for the API path).

## LoginView quirks worth noting

Both items are covered in detail by the docblock on `LoginView` and an inline comment in the template — short version:

- The view declares the custom `'-{', '}-'` delimiter pair so `SmartyTemplateRule` scans the template with the matching pair, and pins it with a literal `@var array{0: '-{', 1: '}-'}` annotation. The base `View::DELIMITERS` carries a wider `array{0: string, 1: string}` annotation which PHPStan reflection inherits onto overridden constants, widening the literal types to plain `string` and breaking `SmartyTemplateRule::delimitersFor()` (it then silently falls back to `{ … }` and starts treating JS object literals as Smarty tags).
- `$redir` stays on the View as a legacy compatibility shim. The legacy `web/themes/default/page_login.tpl` inlines it directly into its login button's `onclick=` and Enter/Space `keydown` handler as a JS expression (`DoLogin('');`, depending on `DoLogin` from `web/scripts/sourcebans.js`). The new template never reads it for login wiring (it posts via `sb.api.call`) but echoes it on a dead `data-legacy-redir` attribute so the dual-theme PHPStan matrix added in #1123 A2 sees the property as referenced. Property + attribute both disappear with #1123 D1.

## Files touched (3, the immutable B1 boundary)

- `web/pages/page.login.php`
- `web/themes/sbpp2026/page_login.tpl` (replaces A1 stub)
- `web/includes/View/LoginView.php` (new)

## Local gates (all green)

| Gate | Result |
| --- | --- |
| `./sbpp.sh phpstan` (default leg) | OK |
| `./sbpp.sh phpstan` (sbpp2026 leg, `phpstan-sbpp2026.neon`) | OK |
| `./sbpp.sh test` | 268 tests, 1134 assertions OK |
| `./sbpp.sh ts-check` | OK |
| `./sbpp.sh composer api-contract` | regenerates byte-identical (no diff) |

## Screenshots

Deferred — the local dev environment lacks a headless browser (no Chromium / Playwright available in the runtime). Visual verification gated to CI / human reviewer with Docker.

## Out of scope for this PR

Per the B1 dispatch: chrome assets (`web/themes/sbpp2026/{css,js,core}/*`), `init.php`, API handlers, `web/scripts/api-contract.js`, `AGENTS.md`, and `ARCHITECTURE.md` are explicitly off-limits — none of them are touched.

### DIFF
--- a/web/includes/View/LoginView.php
+++ b/web/includes/View/LoginView.php
@@ -1,0 +1,81 @@
+<?php
+declare(strict_types=1);
+
+namespace Sbpp\View;
+
+/**
+ * Login page — binds to `page_login.tpl`.
+ *
+ * Both the legacy `web/themes/default/page_login.tpl` and the
+ * sbpp2026 redesign render with the custom `-{ … }-` delimiter pair
+ * so the inline `<script>` can use `{` / `}` for JS object literals
+ * without `{literal}` wrapping. The page handler swaps delimiters
+ * around `Renderer::render()` (mirroring the
+ * `Sbpp\View\YourAccountView` pattern); {@see View::DELIMITERS}
+ * teaches `SmartyTemplateRule` to scan with the matching pair so
+ * template/View parity is still enforced in both PHPStan legs.
+ *
+ * The login screen is anonymous-only: the page handler short-circuits
+ * to the dashboard when the caller already has a session, so this View
+ * never needs the `Sbpp\View\Perms::for()` boolean splat — there are
+ * no `{if $can_*}` gates in the template. The property surface is
+ * intentionally limited to what BOTH templates actually consume so the
+ * dual-theme PHPStan matrix added in #1123 A2 stays clean without
+ * baseline entries during the rollout window:
+ *
+ *   - `$normallogin_show` — gated by `config.enablenormallogin`,
+ *     hides the username/password form when off.
+ *   - `$steamlogin_show`  — gated by `config.enablesteamlogin`,
+ *     hides the "Continue with Steam" button when off.
+ *   - `$redir`            — *legacy compatibility shim*. The legacy
+ *     `web/themes/default/page_login.tpl` inlines this string as a
+ *     JavaScript expression into its login button's `onclick=` and
+ *     into the Enter/Space `keydown` handler (currently the literal
+ *     `DoLogin('');`, which depends on `DoLogin` from
+ *     `web/scripts/sourcebans.js` — loaded only by the legacy chrome).
+ *     The new sbpp2026 template does NOT use it for login wiring (it
+ *     posts via `sb.api.call(Actions.AuthLogin, …)` with a hardcoded
+ *     `redirect: ''`); the property exists on this View purely so the
+ *     legacy theme keeps rendering correctly during the rollout window
+ *     AND the sbpp2026 template's dead `data-legacy-redir` attribute
+ *     keeps SmartyTemplateRule's "every declared property must be
+ *     referenced" check happy across BOTH legs of the dual-theme
+ *     PHPStan matrix added in #1123 A2. Both the property and the
+ *     legacy template go away with #1123 D1.
+ *
+ * Inline error banner content (failed login, locked account, no
+ * access, …) is intentionally NOT a property: the sbpp2026 template
+ * surfaces the `?m=…` query param via `window.SBPP.showToast(...)`
+ * on page load, so the message text lives client-side. The legacy
+ * template still emits `ShowBox()` JS dialogs from the page handler
+ * for backwards compatibility; both branches go away when #1123 D1
+ * deletes the legacy bundle.
+ */
+final class LoginView extends View
+{
+    public const TEMPLATE = 'page_login.tpl';
+
+    // The `@var` is intentionally narrower than the base View::DELIMITERS
+    // (which is `array{0: string, 1: string}`). PHPStan's reflection
+    // inherits the base annotation onto overridden constants, which widens
+    // the literal `'-{'` / `'}-'` types to plain `string` and breaks
+    // `SmartyTemplateRule::delimitersFor()` — that helper inspects each
+    // element via `getConstantStrings()` to pick the matching scan
+    // delimiters, and a non-literal `string` returns zero strings, so the
+    // rule silently falls back to the default `{ … }` pair and starts
+    // matching JS object literals as Smarty tags. Pinning the literal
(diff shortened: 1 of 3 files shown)


rumblefrog (commented):
## Initial review — 0 blocking, 2 non-blocking observations; proceeding to merge.

Read the diff fresh against B1's plan slot (the `sbpp2026-theme-rollout-4c702e97.plan.md` "B1 — Login" entry), `AGENTS.md` (Conventions, Anti-patterns, "nofilter discipline"), and `ARCHITECTURE.md` (Templates + View DTOs). Walked the full reviewer checklist.

### Checklist

| # | Gate | Status |
| --- | --- | --- |
| 1 | Scope boundary — only `web/themes/sbpp2026/page_login.tpl`, `web/includes/View/LoginView.php`, `web/pages/page.login.php` touched | clean — `git diff --stat` = 3 files |
| 2 | AGENTS anti-patterns (jQuery/MooTools/React/xajax/sb-callback/ADOdb, string-literal actions, prefix inlining, `htmlspecialchars_decode` on JSON params, `utf8` charset, WYSIWYG, install/sql edits, unannotated `nofilter`) | none present (no `nofilter` at all; `Actions.AuthLogin` is the autogen contract, not a string literal) |
| 3 | View DTO compliance — template/property parity | `LoginView::{normallogin_show,steamlogin_show,redir}` matches the three template references; `SmartyTemplateRule` would catch drift, both PHPStan legs green |
| 4 | Permission checks | N/A (public page; `is_logged_in()` short-circuit retained) |
| 5 | Variable naming | form-driven, fields match legacy IDs / handoff names |
| 6 | Testability hooks | required `login-username` / `login-password` / `login-submit` / `login-steam` present, plus extras (`login-form`, `login-remember`, `login-disabled`, `login-lostpwd`); inputs have explicit `<label for=…>` |
| 7 | JS conventions | inline `<script>` (not in `web/scripts/`), so `// @ts-check` not applicable; uses `sb.api.call(Actions.AuthLogin, …)`, vanilla JS |
| 8 | CSRF + JSON API | `{csrf_field}` rendered (defends the JS-disabled POST fallback that `route()` runs `CSRF::rejectIfInvalid()` against); JS path uses `sb.api.call` which sets `X-CSRF-Token` from the chrome's `<meta name="csrf-token">` |
| 9 | API contract regen | N/A (no handler changes) |
| 10 | Snapshot tests | N/A |
| 11 | PHPStan baseline | unchanged (verified `git diff … -- web/phpstan-baseline.neon web/phpstan-sbpp2026.neon` is empty) |
| 12 | CI status | all 4 green: PHPStan default, PHPStan sbpp2026, PHPUnit, ts-check |
| 13 | Behavioural correctness | `?m=…` legacy `ShowBox()` calls now `if (typeof ShowBox === 'function')` guarded so the legacy theme keeps working and the sbpp2026 chrome (which drops `sourcebans.js`) silently no-ops; new template surfaces the same messages via `window.SBPP.showToast(…)` driven off `URLSearchParams`; `redirect: ''` → `Api::redirect('?')` lands at the default route post-login |

### `LoginView::DELIMITERS` literal-narrowing override (verified)

Confirmed the docblock claim against `SmartyTemplateRule::delimitersFor()` (web/includes/PHPStan/SmartyTemplateRule.php:214-239) — it calls `getConstantStrings()` on each array element and bails to `['{', '}']` if the count isn't exactly 1, which is exactly what would happen if PHPStan inherited the base's `array{0: string, 1: string}` annotation. The narrower `array{0: '-{', 1: '}-'}` override is the right fix. `YourAccountView` gets away with the looser annotation only because `page_youraccount.tpl` is still an A1 stub and `SmartyTemplateRule::isStubTemplate()` short-circuits before delimiter scanning; B20 will need the same narrowing once it replaces the stub.

### Non-blocking observations (not gating merge)

1. **Inline `<script>` (~50 lines) at the bottom of the template.** The B1 dispatch allows extracting to `web/scripts/page_login.js` with `// @ts-check`. Keeping it inline is fine for B1's scope (legacy `page_login.tpl` also relies on inline JS; D1 cleans up the legacy bundle) and the script is too small to justify a per-page entry. Worth revisiting if the per-page-JS pattern becomes load-bearing for the page later.
2. **`sb.api.call(Actions.AuthLogin, …)` is fire-and-forget — no error feedback if the network drops.** Matches the legacy `web/scripts/sourcebans.js` `DoLogin` shape, so not a regression. `api.js` returns a synthetic `network` error envelope on fetch failure, which is currently dropped here. Could surface via `sb.message.error` later, but not in scope for B1.

Proceeding to merge — `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`, base is current main (95006a03).

PR #1157, opened 2026-05-05
[#1123 followup] Redesign edit ban form (B13 follow-up; drops -{}- delims + MooTools)

rumblefrog:
Part of #1123. Pre-flight follow-up to unblock D1.

The B13 ticket landed seven of the eight admin-bans templates in the new
sbpp2026 theme but missed `page_admin_edit_ban.tpl` — D1's pre-flight
check (`find … -name 'page_admin_edit_ban.tpl' | xargs grep -l "hasn't
been redesigned yet"`) flagged the A1 stub as still present, blocking
the cutover. This patch ships the redesign so D1 can proceed.

## Files touched

- `web/themes/sbpp2026/page_admin_edit_ban.tpl` — replaces the A1 stub
  with the redesigned form. Standard `{ }` Smarty (no `-{ }-`
  override). Mirrors the `page_admin_bans_add.tpl` shape from B13:
  card layout, vanilla form, `{csrf_field}`, every input/select/button
  carries a stable `data-testid="editban-<field>"` hook
  (`editban-name`, `editban-type`, `editban-length`, `editban-steam`,
  `editban-ip`, `editban-reason`, `editban-reason-custom`,
  `editban-demo`, `editban-submit`, `editban-cancel`). The single
  `nofilter` (the `$ban_demo` "Uploaded: <b>safename</b>" snippet)
  carries the canonical safety annotation pointing at the #1113 escape
  contract on the handler side.

- `web/includes/View/AdminBansEditView.php` — **new**. Typed View DTO
  bound to the new template via the `TEMPLATE` constant. Variable
  contract is the intersection both themes consume (`ban_name`,
  `ban_authid`, `ban_ip`, `ban_demo`, `customreason`) plus the
  `can_edit_ban` defense-in-depth gate the new template uses for the
  access-denied card. Length / type / current-reason / per-field
  validation errors / success-redirect are NOT View properties —
  they're emitted by the page handler's vanilla tail script after
  `Renderer::render` so the View stays narrow enough to satisfy
  `SmartyTemplateRule` against both the legacy `-{ … }-` template and
  the new `sbpp2026` one on the dual-theme PHPStan matrix (#1123 A2).
  Does **not** override `View::DELIMITERS`; the legacy template's
  inner `{$var}` matches the standard tag regex so the default leg
  keeps finding every property as referenced.

- `web/pages/admin.edit.ban.php` — rewritten:
  - Drops `setLeftDelimiter('-{')` / `setRightDelimiter('}-')` calls
    and the trailing `$theme->display(…)` chain. Switches to
    `Sbpp\View\Renderer::render($theme, new AdminBansEditView(…))`.
    No per-theme delimiter conditional — the B14 stopgap pattern is
    deliberately skipped because D1 will delete the legacy template
    outright (per spec).
  - Drops `<script>ShowBox(…)</script>` (lives in `web/scripts/sourcebans.js`
    which sbpp2026 doesn't load and #1123 D1 deletes outright).
    Replaces with an `emitEditBanToastAndRedirect()` helper that
    surfaces a toast via `window.SBPP.showToast` (theme.js) with
    `sb.message.{success,error}` (sb.js) as a fallback.
  - Drops the MooTools-style `$('id').setStyle(…)` /
    `$('id').setHTML(…)` `$errorScript` accumulator. Per-field
    validation errors now collect into a structured PHP array, get
    JSON-encoded into the inline tail script, and a vanilla
    `setMsg(id, text)` writes them into the matching `*.msg` divs on
    `DOMContentLoaded`.
  - Drops the call into `selectLengthTypeReason` from
    `sourcebans.js`. Inlines a self-contained vanilla replacement
    that hydrates the type / banlength / listReason `<select>`s
    post-mount with the row's stored values (mirrors B14's
    `admin.edit.comms.php` shim, adjusted for ban-edit's
    banlength-in-seconds and the type 0/1 select).
  - Drops the legacy `$('id').value = …` / `$('id').setHTML(…)`
    inside the popup `demo(id, name)` callback. The new vanilla
    version uses `textContent` + `appendChild` so any HTML
    metacharacters in the demo's original filename (admin-supplied;
    see #1113 audit) render literally instead of parsing as markup.
  - Keeps every DB op + form-validation branch identical to the
    legacy code: same SteamID validation, same duplicate-ban check,
    same admin-immunity guard, same archive-other-submissions
    update, same demo `:prefix_demos` REPLACE, same `Log::add` on
    length change, same redirect target.

- `web/phpstan-baseline.neon` — net `+15 / −6`:
  - Removes the `pages/admin.edit.ban.php` `Variable $userbank might
    not be defined` entry (count: 9). The new handler declares
    `global $userbank` so all nine occurrences resolve.
  - Adds one `AdminBansEditView::$can_edit_ban` unused-property
    entry gating the default-theme leg (the legacy template doesn't
    carry the access-denied card; the `phpstan-sbpp2026.neon` config
    disables `reportUnmatchedIgnoredErrors` so the entry doesn't
    have to fire under the sbpp2026 leg). D1 deletes the legacy
    template and this entry collapses with it.

## Hard rules satisfied

- Stub check: `find web/themes/sbpp2026 -name 'page_admin_edit_ban.tpl' | xargs grep -l "hasn't been redesigned yet"` → 0 hits.
- No `setLeftDelimiter('-{')` / `setRightDelimiter('}-')` / `-{ }-` markers in the new template, View, or handler.
- No MooTools (`setStyle`, `setHTML`, `window.addEvent`, `$('id')` selector); no `ShowBox`; no call into `selectLengthTypeReason` from `sourcebans.js`.
- Permission gate via `can_edit_ban` (computed from `ADMIN_OWNER | ADMIN_EDIT_ALL_BANS`, plus the row-aware `EDIT_OWN_BANS` / `EDIT_GROUP_BANS` branches). Template gates via `{if NOT $can_edit_ban}`.
- `{csrf_field}` on the form.
- Every `{nofilter}` carries a `{* nofilter: <reason> *}` comment immediately above it.
- All SQL uses `:prefix_` literals.

## Test plan

- [x] `./sbpp.sh phpstan` (default leg, 208 files) — green.
- [x] sbpp2026 phpstan leg (`SBPP_PHPSTAN_THEME=sbpp2026 phpstan-sbpp2026.neon`, 208 files) — green.
- [x] `./sbpp.sh test` — 280 tests / 1189 assertions, all green.
- [x] `./sbpp.sh ts-check` — green.
- [x] `./sbpp.sh composer api-contract` — no diff.
- [x] End-to-end smoke against the sbpp2026 dev stack: `GET edit-ban` renders the form populated with the row's current name/authid/ip/length/type/reason; `POST` with a valid edit updates the DB (verified `bid=1` `name`+`length`+`reason` updated) and emits the "Ban updated" toast + redirect script; `POST` with a missing custom reason replays `{"reason":"You must type a reason"}` into `#reason.msg` via vanilla DOM; bad-URL-key, missing-id, and deleted-ban paths emit the vanilla toast + redirect script (no `ShowBox`).

Refs #1123 B13.

### DIFF
--- a/web/includes/View/AdminBansEditView.php
+++ b/web/includes/View/AdminBansEditView.php
@@ -1,0 +1,64 @@
+<?php
+declare(strict_types=1);
+
+namespace Sbpp\View;
+
+/**
+ * Edit-existing-ban form on the admin bans page — binds to
+ * `page_admin_edit_ban.tpl`.
+ *
+ * Variable contract is the intersection the legacy
+ * `default/page_admin_edit_ban.tpl` and the new sbpp2026 redesign both
+ * consume:
+ *
+ *   - `$can_edit_ban`    Defense-in-depth template gate. The page handler
+ *                        already early-`PageDie`'s on access denial, so this
+ *                        is `true` whenever the form actually renders. The
+ *                        sbpp2026 template uses `{if NOT $can_edit_ban}` to
+ *                        render an "Access denied" card if it is ever
+ *                        reached without the gate being honoured upstream.
+ *   - `$ban_name`        Current player name on the ban row.
+ *   - `$ban_authid`      Current Steam2 authid on the ban row.
+ *   - `$ban_ip`          Current IP on the ban row.
+ *   - `$ban_demo`        Pre-built "Uploaded: <b>safename</b>" snippet
+ *                        (or empty), htmlspecialchars'd in the page handler
+ *                        per #1113. Rendered with `{nofilter}` from both
+ *                        themes; the safety annotation lives at the call
+ *                        site in the sbpp2026 template.
+ *   - `$customreason`    `false` when custom reasons are disabled, otherwise
+ *                        the list of reason strings from `bans.customreasons`.
+ *
+ * Length / type / current-reason / post-validation errors / success-redirect
+ * are NOT Smarty vars: the page handler emits a vanilla `<script>` tail
+ * after `Renderer::render` to hydrate the `<select>`s, surface per-field
+ * validation errors, and trigger the "saved" toast → redirect. This mirrors
+ * the legacy convention preserved during the v2.0.0 rollout window so the
+ * View stays narrow enough that the legacy `-{ … }-` template's variable
+ * references match the View's properties on the dual-theme PHPStan matrix
+ * (#1123 A2).
+ *
+ * The legacy template uses Smarty's custom `-{ … }-` delimiters; the
+ * sbpp2026 redesign moves to the standard `{ … }` pair so the View binds
+ * with the default {@see View::DELIMITERS}. The page handler does NOT swap
+ * delimiters per active theme — D1 deletes the legacy template outright, so
+ * carrying a per-theme conditional here would just be dead-on-arrival
+ * scaffolding (see `web/pages/admin.edit.ban.php` for the rationale).
+ */
+final class AdminBansEditView extends View
+{
+    public const TEMPLATE = 'page_admin_edit_ban.tpl';
+
+    /**
+     * @param false|list<string> $customreason `false` when custom reasons
+     *     are disabled, otherwise the list of reason strings.
+     */
+    public function __construct(
+        public readonly bool $can_edit_ban,
+        public readonly string $ban_name,
+        public readonly string $ban_authid,
+        public readonly string $ban_ip,
+        public readonly string $ban_demo,
+        public readonly false|array $customreason,
+    ) {
+    }
+}
--- a/web/pages/admin.edit.ban.php
+++ b/web/pages/admin.edit.ban.php
@@ -22,16 +22,57 @@ if (!defined("IN_SB")) {
     die();
 }
 
-global $theme;
+global $userbank, $theme;
 
 new AdminTabs([], $userbank, $theme);
 
+/**
+ * Vanilla replacement for sourcebans.js' ShowBox(): emits an inline
+ * <script> that surfaces a toast via window.SBPP.showToast (theme.js,
+ * sbpp2026) with sb.message.* (sb.js, both themes) as a fallback, then
+ * redirects after a short delay. Used in place of the legacy
+ * `<script>ShowBox(…)</script>` calls so the page works under sbpp2026
+ * (which does NOT load sourcebans.js) without losing the legacy theme.
+ *
+ * The strings are JSON-encoded so embedded quotes / newlines / non-ASCII
+ * survive the round-trip into the script body without further escaping.
+ */
+function emitEditBanToastAndRedirect(string $kind, string $title, string $body, string $redirect): void
+{
+    $kindJs = json_encode($kind, JSON_THROW_ON_ERROR | JSON_HEX_TAG | JSON_HEX_AMP | JSON_HEX_APOS | JSON_HEX_QUOT);
+    $titleJs = json_encode($title, JSON_THROW_ON_ERROR | JSON_HEX_TAG | JSON_HEX_AMP | JSON_HEX_APOS | JSON_HEX_QUOT);
+    $bodyJs = json_encode($body, JSON_THROW_ON_ERROR | JSON_HEX_TAG | JSON_HEX_AMP | JSON_HEX_APOS | JSON_HEX_QUOT);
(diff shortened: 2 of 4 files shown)


rumblefrog (commented):
Initial review pass. **0 blocking, 4 non-blocking observations.** CI green on all 4 gates (PHPStan default + sbpp2026 + PHPUnit + ts-check); mergeable; no file collision with #1158 or #1159. Approving + merging in the next sweep.

Checklist (against the prompt):

- **Scope** — 4 files (template / View / handler / baseline), as spec'd. ✓
- **Anti-patterns** — no MooTools (`setStyle`, `setHTML`, `window.addEvent`, `$('id')`); no `ShowBox`; no call into `selectLengthTypeReason` from `sourcebans.js`. The replacement shim is inline in the handler tail script. ✓
- **View DTO** — `AdminBansEditView` extends `View`, has `TEMPLATE` constant, **does not** override `DELIMITERS`. The legacy `-{$ban_name}-` substitutions still match the standard `{$ban_name}` regex inside the `-` literals, so SmartyTemplateRule keeps finding `ban_name`/`ban_authid`/`ban_ip`/`ban_demo`/`customreason` on the default leg too — only `can_edit_ban` (sbpp2026-only) baselines out. ✓
- **Permission gate** — `can_edit_ban` precomputed via `ADMIN_OWNER | ADMIN_EDIT_ALL_BANS` plus row-aware `EDIT_OWN_BANS` / `EDIT_GROUP_BANS`; template gates with `{if NOT $can_edit_ban}`; handler also early-`PageDie`s. ✓
- **Variable naming, testability hooks, CSRF, vanilla JS** — all good (`editban-<field>` testids, `{csrf_field}`, no string-literal action names). ✓
- **PHPStan baseline** — net `+15/-6` matches: `-9` on `pages/admin.edit.ban.php` `Variable $userbank might not be defined` (resolved by `global $userbank`), `+1` on `AdminBansEditView::$can_edit_ban` unused-property (default-leg dual-theme bridging, collapses at D1). ✓
- **Behavioural** — drops `setLeftDelimiter('-{')`/`setRightDelimiter('}-')` ✓; switches to `Renderer::render(new AdminBansEditView(...))` ✓; preserves DB ops + form-validation byte-for-byte ✓; toast pattern matches B13/B20 (`window.SBPP.showToast` with `sb.message.{success,error}` fallback) ✓.

### Non-blocking observations

1. **Default-theme rendering trade-off (most substantive).** The B14 `admin.edit.comms.php` pattern keeps a `$useLegacyDelims = Config::get('config.theme') !== 'sbpp2026'` conditional so the legacy `-{ }-` template still renders correctly until D1. This PR drops the conditional outright. With default `{ }` delimiters parsing the legacy template, the `-` characters surrounding each tag become literal output, so default-theme operators see `value=\"-VanillaName-\"` in inputs and `<option value=\"-Custom1-\">-Custom1-</option>` in the reason dropdown. If an admin clicks **Save** without manually correcting the dashy values, `$reason = -Custom1-` gets persisted (data corruption). PR description acknowledges the trade-off (\"D1 will delete the legacy template outright\") and the worker followed the prompt, so this is deliberate — but the corruption window is real until D1 lands. Worth pinning the D1 PR to land in close succession.

2. **Latent privilege-escalation bug fixed in the new access logic** (worth crediting in the PR body). Legacy denial used `A && (B && C) && (D && E)` which granted access to *anyone with `EDIT_OWN_BANS`* regardless of ban ownership (the `&&` short-circuits when the user has the perm). New code uses the correct `||`-of-positive-grants shape, so EDIT_OWN_BANS now actually restricts to own bans. Strict tightening — could surprise operators who relied on the buggy permissive behavior.

3. **`!$res` flow change.** Legacy code echoed `ShowBox(\"...ban deleted?\")` but **did not** `PageDie`, then fell through to render an empty form. With the legacy `&&`-bug access check above the `!$res` branch, this path was effectively dead (access check tripped first on null `$res['aid']`/`$res['gid']`). New code reorders: `!$res` check first + `PageDie`. Correct fix; just noting it's a behavioral change.

4. **Cosmetic.**
   - `<input type=\"hidden\" name=\"insert_type\" value=\"add\">` is copy-paste from `page_admin_bans_add.tpl`; the edit handler never reads `$_POST['insert_type']`. Harmless but could be removed.
   - Template's leading docblock lists `nick.msg` among the DOM IDs but the actual ID is `name.msg` (no `nick.msg` exists in this template — that's the add-ban template's id).
   - Two `DOMContentLoaded` listeners in the tail IIFE; could be merged for clarity.

None of these block the merge. Proceeding to approve.

PR #1159, opened 2026-05-05
[#1123 followup] Vanilla replacements for unguarded sourcebans.js callers (D1 prep)

rumblefrog:
Part of #1123. Pre-flight follow-up to unblock D1. Replaces 8 unguarded `sourcebans.js` function call sites (and the adjacent `ShowBox` / `LoadPaste*` siblings in the same files that would also `ReferenceError` post-D1) with vanilla equivalents or guarded shims so D1's `git rm web/scripts/sourcebans.js` doesn't break the panel.

## Summary

| File | Sites addressed | Replacement strategy |
|------|-----------------|----------------------|
| `web/pages/admin.bans.php` | `ProcessBan` (`ShowKickBox` / `TabToReload` / `applyApiResponse`), 4× `<select onchange="changePage(...)">`, `LoadPrepareReban` / `LoadPasteBan`, 2× `ShowBox` banners | Inline `sb.message.show` + `setTimeout(location.reload)`; direct `window.location.href` updates; `sb.api.call(Actions.BansPrepareReban / Actions.BansPaste)` + new `window.__sbppApplyBanFields` tail helper (replaces `applyBanFields`). |
| `web/pages/admin.comms.php` | `ProcessBan` (`ShowBlockBox` / `TabToReload` / `applyApiResponse`), `LoadPrepareReblock` / `LoadPrepareBlockFromBan` / `LoadPasteBlock`, 2× `ShowBox` banners | Same shape as `admin.bans.php`, plus `window.__sbppApplyBlockFields` helper (replaces `applyBlockFields`). |
| `web/pages/admin.groups.php` | `InitAccordion` | `sb.accordion(...)` (lives in `sb.js`, survives D1). |
| `web/pages/admin.edit.adminservers.php` | `ShowRehashBox` + `TabToReload` + `ShowBox` | Inline `sb.api.call(Actions.SystemRehashAdmins)` + `sb.message.show`. |
| `web/pages/page.servers.php` | `LoadServerHost` script-builder | New `SbppServerQryHelpers()` PHP function (function_exists-guarded) emits `window.__sbppLoadServerHost{,Property}` definitions; the loop emits `__sbppLoadServerHost(SID);` calls. |
| `web/pages/page.home.php` | `LoadServerHostProperty` script-builder | Mirror of above; both files ship the same helper definition so either can lead. |
| `web/themes/sbpp2026/page_admin_groups_add.tpl` | `applyApiResponse(r)` | Inline envelope handler that prefers `window.SBPP.showToast` and falls back to `sb.message`. |
| `web/themes/sbpp2026/page_admin_bans_groups.tpl` | `LoadGetGroups` | Inline `sb.api.call(Actions.BansGetGroups)` with the same DOM-ops shape, so `TickSelectAll` / `CheckGroupBan` keep operating on the same `#steamGroupsTable` rows. |

Replacements: 13 call sites across 8 files (8 from the original D1 pre-flight scan + 5 adjacent `sourcebans.js`-only sites in the same files that would also `ReferenceError` post-D1). All inline replacements; no guarded shims needed.

The new `sbpp2026/core/footer.tpl` deliberately drops `{$query nofilter}`, so `__sbppLoadServerHost{,Property}` only runs under the legacy default theme + any third-party theme that keeps the legacy footer shape.

## Test plan

Local gates (all green):
- `./sbpp.sh phpstan` — 0 errors (208 files).
- `./sbpp.sh ts-check` — clean.
- `./sbpp.sh composer api-contract` — no diff (no handler changes).
- `./sbpp.sh test` — 280 tests / 1189 assertions, all passing.
- `php -l` clean on every modified PHP file.

Manual smoke-test:
- [ ] `/?p=home` renders under sbpp2026 theme; no `ReferenceError` in console.
- [ ] `/?p=servers` renders under sbpp2026 theme; per-card `Actions.ServersHostPlayers` call still fires (uses the template's own JS, unrelated to the helper).
- [ ] `/?p=admin&c=bans` admin tab list pagination + Add a ban form (incl. paste-from-server flow) under both default and sbpp2026 themes.
- [ ] `/?p=admin&c=comms` Add a block form (incl. reblock + block-from-ban + paste flows) under both themes.
- [ ] `/?p=admin&c=groups` accordion open/close under both themes.
- [ ] `/?p=admin&c=admins&o=editservers` rehash dialog after submitting under both themes.

Out of scope (separate followups, per the task brief):
- 4 search-box stub redesigns.
- `page_admin_edit_ban.tpl` redesign.
- Script-builders in `admin.{bans,admins,comms}.search.php` that emit `LoadServerHost` (handled by the search-stubs follow-up).

### DIFF
--- a/web/pages/admin.bans.php
+++ b/web/pages/admin.bans.php
@@ -32,9 +32,10 @@ new AdminTabs([
 ], $userbank, $theme);
 
 if (isset($_GET['mode']) && $_GET['mode'] == "delete") {
-    echo "<script>ShowBox('Ban Deleted', 'The ban has been deleted from SourceBans', 'green', '', true);</script>";
+    // Inlined sourcebans.js helper (#1123 D1 prep): ShowBox is removed at D1; sb.message is in sb.js.
+    echo "<script>sb.message.show('Ban Deleted', 'The ban has been deleted from SourceBans', 'green', '', true);</script>";
 } elseif (isset($_GET['mode']) && $_GET['mode']=="unban") {
-    echo "<script>ShowBox('Player Unbanned', 'The Player has been unbanned from SourceBans', 'green', '', true);</script>";
+    echo "<script>sb.message.show('Player Unbanned', 'The Player has been unbanned from SourceBans', 'green', '', true);</script>";
 }
 
 if (isset($GLOBALS['IN_ADMIN'])) {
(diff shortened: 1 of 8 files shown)


rumblefrog (commented):
Initial review. 2 blocking, 4 non-blocking; addressing the blockers now.

## Blocking

1. **`admin.comms.php` ProcessBan drops the blockit iframe**, so successful comm-blocks never propagate to gameservers. The bans path correctly preserves `<iframe src="pages/admin.kickit.php?...">`; the comms path replaced `ShowBlockBox(steam, type, length)` with a plain "The block has been successfully added." toast. Without the `pages/admin.blockit.php?check=…&type=…&length=…` iframe loading, `api_blockit_block_player` (which fires the rcon `sc_fw_block …` per enabled server) never runs. The DB row is created, the user sees a green toast, but the player is never actually silenced/muted on live servers — silent functional regression. (The legacy default theme is still served `sourcebans.js`, but `admin.comms.php`'s tail script now invokes the rewritten branch regardless of theme, since this is the page-level `<script>` block.)

2. **`page_admin_groups_add.tpl` calls `window.SBPP.showToast({ type, message })` with the wrong shape.** The actual signature in `web/themes/sbpp2026/js/theme.js` is `{ kind, title, body }`. Passing `type`/`message` falls through to `kind = 'info'` (default) and `title = undefined`, so the user sees a generic info-styled toast that literally renders the string `undefined` (because `escapeHtml(undefined) === 'undefined'`). Both the success and error branches are affected. Should match the shape used elsewhere in sbpp2026 (e.g. `page_admin_servers_list.tpl`, `page_admin_bans_email.tpl`).

## Non-blocking (will not fix in this PR unless asked)

3. `page.servers.php`'s rewire from `LoadServerHost(...)` to the simplified `__sbppLoadServerHost(sid)` drops the `inHome=false` branches (no players placeholder, mapimg, accordion auto-open). Only affects the legacy default theme + third-party themes copying its footer pattern, only during the brief D1 prep window. Worker's helper comment already documents the trade-off; flagging for awareness.

4. `__sbppApplyBlockFields` correctly omits `ip` and `demo.msg` (they don't exist on the comms add form). Worth a one-line comment that this is deliberately narrower than legacy `applyBanFields`, which gets reused for both flows.

5. `page_admin_groups_add.tpl` reload uses 1500 ms vs legacy `TabToReload` 2000 ms. Cosmetic.

6. `LoadGetGroups` replacement in `page_admin_bans_groups.tpl` drops `td.className = 'listtable_1'`. Likely intentional under sbpp2026 (no such class), but legacy parity would set it.

CI is currently green (PHPStan dual-theme, PHPUnit, ts-check all pass).

rumblefrog (commented):
Both blockers fixed in f5904018:

- Comms `ProcessBan` now emits the `pages/admin.blockit.php?check=&type=&length=` iframe with `encodeURIComponent` on each query param, mirroring the bans/kickit branch one above. `noclose` flipped back to `true` so the iframe has time to dispatch.
- `page_admin_groups_add.tpl` now calls `SBPP.showToast({kind, title, body})` per the actual contract in `themes/sbpp2026/js/theme.js`. Reload aligned to 2 s for parity with the rest of the panel.

CI green on all four gates (PHPStan dual-theme, PHPUnit 280/1189, ts-check). Self-merging.

PR #1173, opened 2026-05-05
tests/e2e: flow-admin-ban-lifecycle (add → list → edit → unban → confirm)

rumblefrog:
Refs #1124. Slice 4 of the E2E rollout.

## What

`web/tests/e2e/specs/flows/admin-ban-lifecycle.spec.ts` walks the
admin ban moderation surface end-to-end on the v2.0.0 chrome:

1. Open `?p=admin&c=bans`, fill the "Add a ban" form
   (`addban-*` testids), submit, assert the `Ban added` toast.
2. Navigate to `?p=banlist`, find the new row by SteamID, capture
   its `data-id` (`bid`) for cross-step anchoring, assert
   `data-state="permanent"`.
3. Confirm the same row surfaces in the public list view.
4. Click `row-action-edit`, change the reason via the edit-ban
   form (`editban-*` testids), submit, assert the `Ban updated`
   toast and the new reason in the row.
5. Click `row-action-unban`, assert the row's `data-state` flips
   to `unbanned`, the unban affordance disappears, and the public
   list still shows the row in its `unbanned` state.

`_screenshots.spec.ts` picks up an appended
`@screenshot flow-admin-ban-lifecycle` describe block — five
deterministic frames × `light` + `dark`:

- `addban-empty`
- `list-after-add`
- `editban-prefilled`
- `list-with-unban-action` *(closest analogue to the issue's
  "unban modal/prompt open" frame — the legacy `confirm()` lived
  in `sourcebans.js`, gone since #1123 D1; v2.0.0's unban path is
  a direct GET, so the closest "about to unban" state is the row
  with the affordance visible)*
- `list-after-unban`

Each frame `truncateE2eDb()`s before its setup so the captures are
hermetic. Mobile chromium is skipped per the same constraint the
spec uses (shared `sourcebans_e2e` would race the truncate across
projects; Slice 7's responsive gallery covers the mobile chrome
separately). Within the desktop project the block is
`mode: 'serial'` so concurrent workers don't trample each other.

Drive-by fix in `web/pages/page.banlist.php` (three call sites,
identical shape): the paginated `SELECT … LIMIT ?,?` queries
bound limit/offset positionally via `PDOStatement::execute(array)`,
which (with the default `ATTR_EMULATE_PREPARES => true` PDO
carries on mysql) sends the two ints as quoted strings
(`LIMIT '0','30'`). MariaDB 10.11 strict mode rejects that with a
1064 syntax error, so the public banlist 500'd as soon as a
page-2 request landed. Switched the three sites to
`LIMIT " . intval($BansStart) . "," . intval($BansPerPage)`
(intval-clamped interpolation, no SQL-injection surface) — the
canonical PDO LIMIT idiom that the rest of the codebase uses for
LIMIT clauses, and what step 4 of the flow needs to land green.

## Why one linear test, not five

Steps depend on data the previous step created — the new ban's
`bid` for cross-navigation row anchoring, the row's edit/unban
hrefs that embed the per-session `admin_postkey`. Splitting into
separate `test()` blocks would re-truncate the DB between them
and force every step to redo the earlier setup. Keeping the
lifecycle as one linear test mirrors the user-facing flow and
keeps the trace artifact a single scrubbable timeline on failure.

## Hooks audit

Every `data-testid`, `data-state`, and `data-id` the spec asserts
on already exists in the templates from #1123 B13 + #1157
(`page_admin_bans_add.tpl`, `page_admin_edit_ban.tpl`,
`page_bans.tpl`). No template tweaks needed.

## Divergences from the literal slice spec

- **Mobile chromium for the gallery is skipped.** The flow
  mutates `:prefix_bans` heavily and the per-test
  `truncateE2eDb()` is the only resetter; running both projects
  concurrently against the shared schema makes the post-truncate
  re-seed PK-collide on the admin row. The flow's UI surface is
  the same desktop chrome at mobile width; Slice 7's responsive
  gallery (`specs/responsive/`) locks the mobile chrome
  separately, so the mobile frame here would be duplicate
  coverage anyway.
- **"Unban modal/prompt open" frame substituted with
  `list-with-unban-action`.** The legacy `confirm('Unban...?')`
  modal lived in `sourcebans.js`, which #1123 D1 removed; the
  current chrome unbans via a direct GET on the row's
  `row-action-unban` anchor. The closest deterministic "about to
  unban" frame is the row hovered with the affordance visible —
  that's what the gallery captures.
- **Foundation gap noticed (not fixed in this PR).** The base
  `docker-compose.yml` hard-pins `DB_NAME=sourcebans` on the
  apache container, so `./sbpp.sh up` HTTP requests hit the dev
  DB while `truncateE2eDb()` resets `sourcebans_e2e`. Slice 0's
  smoke spec doesn't notice (login works against any DB that has
  admin/admin seeded — both do); Slice 3 lands one row in
  `sourcebans` and gets away with it. Any state-mutating flow
  (this slice) sees the dev DB. Worked around in this worktree's
  gitignored `docker-compose.override.yml`
  (`environment: { DB_NAME: sourcebans_e2e }` + a one-shot
  `rm config.php && docker compose up -d --force-recreate web`
  to regenerate the panel's config from the new env). A
  permanent fix likely belongs in either `./sbpp.sh e2e` (force
  the apache container's DB_NAME for the duration of the suite)
  or the entrypoint (regenerate `config.php` on every boot
  rather than only when missing). Out of scope for this slice;
  flagging here so a follow-up can land it once Slice 3/5's
  state-mutating flows ship.

## Quality gates (local, parallel-stack via `docker-compose.override.yml`)

- `./sbpp.sh phpstan` → `[OK] No errors`.
- `./sbpp.sh test` → `OK (280 tests, 1189 assertions)`.
- `./sbpp.sh ts-check` → clean.
- `./sbpp.sh composer api-contract` → zero diff.
- `./sbpp.sh e2e` → 5 passed (admin ban lifecycle on chromium +
  4 smoke logins on chromium + mobile-chromium), 25 skipped
  (mobile lifecycle + screenshot gallery without `SCREENSHOTS=1`).
- `SCREENSHOTS=1 ./sbpp.sh e2e --grep @screenshot` → 14 passed
  (login × 2 themes × 2 projects + flow-admin-ban-lifecycle × 5
  frames × 2 themes desktop), 10 skipped (mobile flow gallery).

## Acceptance criteria (from #1124)

- [x] Critical flow: admin ban lifecycle (add → list → edit → unban
  → confirm) — this PR.

### DIFF
--- a/web/pages/admin.edit.ban.php
+++ b/web/pages/admin.edit.ban.php
@@ -76,14 +76,19 @@ if (!isset($_GET['id']) || !is_numeric($_GET['id'])) {
 }
 $_GET['id'] = (int) $_GET['id'];
 
+// Native PDO prepares (#1175 / Slice 3) reject the same named placeholder
+// reused across positions, so split the bid lookup into two distinct
+// names and bind both. The subquery's `:demo_bid` and the outer
+// `:bid` both pull from `$_GET['id']`, which is already int-cast above.
 $GLOBALS['PDO']->query("
-    				SELECT bid, ba.ip, ba.type, ba.authid, ba.name, created, ends, length, reason, ba.aid, ba.sid AS ba_sid, ad.user, ad.gid, CONCAT(se.ip,':',se.port) AS server_addr, se.sid AS se_sid, mo.icon, (SELECT origname FROM `:prefix_demos` WHERE demtype = 'b' AND demid = :id) AS dname
+    				SELECT bid, ba.ip, ba.type, ba.authid, ba.name, created, ends, length, reason, ba.aid, ba.sid AS ba_sid, ad.user, ad.gid, CONCAT(se.ip,':',se.port) AS server_addr, se.sid AS se_sid, mo.icon, (SELECT origname FROM `:prefix_demos` WHERE demtype = 'b' AND demid = :demo_bid) AS dname
     				FROM `:prefix_bans` AS ba
     				LEFT JOIN `:prefix_admins` AS ad ON ba.aid = ad.aid
     				LEFT JOIN `:prefix_servers` AS se ON se.sid = ba.sid
     				LEFT JOIN `:prefix_mods` AS mo ON mo.mid = se.modid
-    				WHERE bid = :id");
-$GLOBALS['PDO']->bind(':id', $_GET['id']);
+    				WHERE bid = :bid");
+$GLOBALS['PDO']->bind(':bid', $_GET['id']);
+$GLOBALS['PDO']->bind(':demo_bid', $_GET['id']);
 $res = $GLOBALS['PDO']->single();
 
 isset($_GET["page"]) ? $pagelink = "&page=" . urlencode($_GET["page"]) : $pagelink = "";
--- a/web/tests/e2e/specs/_screenshots.spec.ts
+++ b/web/tests/e2e/specs/_screenshots.spec.ts
@@ -956,3 +956,208 @@ test.describe('@screenshot flow-comms-gag-mute', () => {
         });
     }
 });
+
+/**
+ * Per-flow gallery — admin ban lifecycle (#1124 Slice 4).
+ *
+ * Five frames, each its own deterministic test so a single failure
+ * scopes to one PNG (the gallery rebuilds the DB per test):
+ *
+ *   1. addban-empty           — pristine "Add a ban" form
+ *   2. list-after-add         — admin list with the new permanent ban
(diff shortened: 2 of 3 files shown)


rumblefrog:
## Screenshots (flow-admin-ban-lifecycle)

| `login` |  |  |  |  |

_Re-shot post-rebase onto `main` (Slices 2 + 3 + 6 landed). The lifecycle frames are desktop-only by design (DB race + Slice 7 owns the mobile gallery); `login` is the shared baseline from the gallery describe._

PR #1174, opened 2026-05-05
tests/e2e: flow-comms-gag-mute (gag → list → unblock)

rumblefrog:
Refs #1124. Critical-flows partial (3/4 once Slices 3+4 are also in).

## Summary

Adds Slice 5 of the #1124 Playwright suite: a critical-path e2e spec
covering the admin comms (gag) → admin list → public list → unblock →
admin list flow, plus an `@screenshot` block for the per-PR gallery.

- `web/tests/e2e/specs/flows/comms-gag-mute.spec.ts` — single-context
  flow spec, truncates per-test, anchors on the row's `data-id` across
  navigations.
- `web/tests/e2e/specs/_screenshots.spec.ts` — append-only addition of
  `@screenshot flow-comms-gag-mute` (empty add form / active list /
  unblocked list, light + dark, desktop only — see "Divergences").

Local quality gates (parallel stack on ports 8240–8244):

| Gate | Result |
| --- | --- |
| `./sbpp.sh phpstan` | clean |
| `./sbpp.sh test` | OK (280 tests, 1189 assertions) |
| `./sbpp.sh e2e` | 5 passed, 9 skipped (mobile-chromium for state-mutating, see below) |
| `SCREENSHOTS=1 ./sbpp.sh e2e --grep '@screenshot flow-comms-gag-mute'` | 2 passed, 2 skipped (mobile-chromium) |

## Pre-flight bug fix (`prep` commit)

`/index.php?p=commslist` fataled on every fresh request with:

    SQLSTATE[42000]: 1064 ... near ''0','30'' ... at line 13 in Database.php:110

`page.commslist.php` passed `LIMIT ?,?` placeholders via
`resultset([..., intval($BansStart), intval($BansPerPage)])`. PDO
emulated prepares (the default) stringify positional `execute()`
arguments, so MariaDB 10.11 sees `LIMIT '0','30'` and rejects. Latent
on `main` because no spec actually loaded `/commslist` until this
slice. Fixed in all three call sites (default, searchText, advSearch)
by mirroring the inline `intval()` pattern that `admin.admins.php`
already uses for the same construct.

**Follow-up:** `page.banlist.php` carries the same anti-pattern in
three more sites (`page.banlist.php:285,312,497`). Left alone here for
Slice 1 (`smoke-public`) coordination — that slice's specs hit
`/banlist` so the bug surfaces there too. Whichever PR lands first
should pull the same `intval()` inline pattern; the other can rebase
trivially.

## Divergences from the brief

1. **Brief step 3 says goto `/index.php?p=admin&c=comms` for the admin
   list.** That route renders the add form, not a list. The list lives
   at `?p=commslist` for both admin and public; the spec threads the
   brief's intent on that URL.
2. **Brief mentions `data-testid="row-action-unblock"`.** The actual
   template emits `row-action-unmute` (#1123 B14 picked the verb that
   matches the SQL `RemoveType`, "U" / "Unmuted"). Asserting against
   what's there; no template re-tag — not a missing-hook escalation.
3. **No unblock confirmation modal.** The legacy `UnGag()` JS confirm
   was removed at #1123 D1. The row's `unmute_url` is a direct anchor;
   we append `&ureason=e2e: lifted` to the href to honour the brief's
   "provide an unblock reason" step (the server reads `ureason` from
   `$_GET`).
4. **No "unblock prompt/modal" screenshot.** There's no modal to
   capture (see #3). Gallery has three frames per (theme × desktop)
   instead of four.

## Slice 0 / harness follow-ups (flagged, not fixed here)

- **`truncateE2eDb()` is not parallel-project-safe.** Two workers
  (chromium + mobile-chromium) racing on `beforeEach` for the same
  test fail with `1062 Duplicate entry '0' for key 'PRIMARY'` when
  their seed `INSERT`s interleave. The flow spec and the screenshot
  block both pin to `chromium` until `Fixture::truncateOnly()` grows
  a per-DB advisory lock (`GET_LOCK('sbpp_e2e_truncate', N)`). Mobile
  viewport coverage for the comms-add chrome will move into the
  gallery in that follow-up.
- **`web` container's `DB_NAME` defaults to `sourcebans` (the dev
  schema), not `sourcebans_e2e`.** `truncateE2eDb()` cleans the e2e
  schema; Apache writes to dev. State-mutating specs silently collide
  with leftover dev rows. Worked around in this slice's
  `docker-compose.override.yml` via `services.web.environment.DB_NAME:
  sourcebans_e2e`. Each parallel-stack worker currently has to opt in
  the same way; long-term `./sbpp.sh e2e` should rebuild the web
  service against `sourcebans_e2e` (or the entrypoint should read an
  `E2E_*` env split) so the override isn't load-bearing.
- **`sb.message.show()` is a silent no-op in the new theme** — the
  legacy `#dialog-placement` shell isn't part of `sbpp2026`.
  `web/pages/admin.comms.php`'s `ProcessBan()` calls `sb.message.show`
  on success, so a successful gag silently submits and reloads with
  no toast feedback. The spec asserts on the JSON API response
  instead. Page-tail JS migration to `window.SBPP.showToast` is a
  separate UX cleanup.
- **`page.commslist.php` and `page.banlist.php` still call legacy
  `ShowBox()` after a state mutation.** That helper went away with
  `sourcebans.js` (#1123 D1); it now throws `ReferenceError` in the
  inline `<script>` block but the server-side update has already
  committed by then, so the test is unaffected. Visible to users as
  a missing toast on `?a=ungag` / `?a=unmute` / `?a=delete`.

## Test plan

- [x] `./sbpp.sh phpstan` clean
- [x] `./sbpp.sh test` green (280 / 1189)
- [x] `./sbpp.sh ts-check` clean
- [x] `./sbpp.sh composer api-contract` zero diff
- [x] `./sbpp.sh e2e` green (existing specs unchanged)
- [x] `SCREENSHOTS=1 ./sbpp.sh e2e --grep '@screenshot flow-comms-gag-mute'` produces six PNGs (light/dark × add-form/list-active/list-unblocked, desktop only)
- [ ] CI passes
- [ ] Screenshots archived via `web/tests/e2e/scripts/upload-screenshots.sh` (posted as PR comment after creation)

### DIFF
--- a/web/tests/e2e/specs/_screenshots.spec.ts
+++ b/web/tests/e2e/specs/_screenshots.spec.ts
@@ -46,6 +46,7 @@ import {
     type SubmissionFixture,
 } from '../pages/SubmitBanFlow.ts';
 import { seedBanViaApi } from '../fixtures/seeds.ts';
+import { truncateE2eDb } from '../fixtures/db.ts';
 import { mkdir } from 'node:fs/promises';
 import { dirname, resolve } from 'node:path';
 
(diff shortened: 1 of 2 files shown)


rumblefrog:
## Screenshots (flow-comms-gag-mute)

| `login` |  |  |  |  |

PR #1162, opened 2026-05-05
tests/e2e: foundation (Playwright + axe + screenshot gallery harness)

rumblefrog:
Refs #1124

Foundation PR for the new Playwright E2E suite. Adds the harness, one
proof spec, the per-PR screenshot gallery flow, and a new CI gate. No
behaviour change to the panel itself — the only chrome-adjacent edits
are a new `Fixture::truncateOnly()` (so the cross-process e2e DB shim
doesn't race with parallel Playwright workers on DROP+CREATE) and a
parallel grant for the `sourcebans_e2e` schema in
`docker/db-init/00-render-schema.sh` (mirrors the existing
`sourcebans_test` grant).

## Acceptance criteria boxes ticked

- [x] `web/tests/e2e/` exists with Playwright config, TypeScript,
      page-object scaffold, axe integration.
- [x] `./sbpp.sh e2e` runs the suite locally against `./sbpp.sh up`'s
      stack.
- [x] `.github/workflows/e2e.yml` runs on every PR; failures upload
      Playwright trace + HTML report as an artifact.
- [x] `AGENTS.md` *Dev commands* + *Quality gates* tables updated;
      `ARCHITECTURE.md` *Quality gates* updated; `docker/README.md`
      documents the new sbpp.sh subcommand.
- [x] `web/tests/e2e/.gitignore` excludes `node_modules/`,
      `playwright-report/`, `test-results/`, `screenshots/`,
      `playwright/.auth/`.
- (Smoke per route, flows, responsive, a11y per page in light + dark
  — Slices 1–8.)

## Routes covered (initial smoke)

- `/login` (drives the form directly; the storage-state auth fixture
  in `fixtures/global-setup.ts` was validated against this same form
  to mint `playwright/.auth/admin.json`).

## Local commands run

- `./sbpp.sh up` (with worktree-local `docker-compose.override.yml`,
  project `sbpp-e2e-foundation`, web on :8190).
- `./sbpp.sh phpstan` — clean (zero new errors over baseline).
- `./sbpp.sh test` — 280 tests, 1189 assertions, all green.
- `./sbpp.sh ts-check` — clean (the new `web/tests/e2e/*.ts` files are
  also `tsc`-checked via the e2e tsconfig.json — separate config so
  the existing `web/scripts` tsc gate isn't widened).
- `./sbpp.sh composer api-contract` — zero diff.
- `./sbpp.sh e2e` — 4 tests pass (login × 2 projects × 2 cases),
[...]

PR #1175, opened 2026-05-05
tests/e2e: flow-public-submission (anonymous submit → admin approve → public list)

rumblefrog:
Refs #1124

Slice 3 of #1124 — first stateful Playwright spec covering one of the
"four critical flows" the parent issue asks for:

> At least the four critical flows above are covered end-to-end —
> partial (1/4: public ban submission).

The spec drives the full lifecycle via the live UI:

1. **Anonymous submit** — logged-out browser context (`browser.newContext({ storageState: empty, ...projectUse })`) hits `/index.php?p=submit`, fills every required field with deterministic-but-unique fixture data, picks the static "Other server" option (value=0 — passes server-side validation without depending on a `:prefix_servers` seed), and submits. We anchor the success-side-effect on `data-testid="submitban-reason"` round-tripping to empty (`page.submit.php` resets every captured POST value on success; a validation bounce keeps the original text intact).

2. **Admin queue** — switch back to the admin context (`page` fixture inherits Slice 0's `playwright/.auth/admin.json`). Goto `/index.php?p=admin&c=bans`, find the queued row by SteamID via `data-testid="submission-row"` filtered through `submission-row-steam`.

3. **Approve** — click `data-testid="row-action-ban"` on the row → `Actions.BansSetupBan` pre-fills the Add Ban form via `__sbppApplyBanFields`. Wait for `data-testid="addban-nickname"` to reflect the submission's player name (terminal state: setup-ban succeeded). Pick a static reason (`Inappropriate Language` from the legacy-locked optgroup), submit. `Actions.BansAdd` inserts the ban + flips every submission for that SteamID to `archiv=3`. Helper waits on `.toast[data-kind="success"]` filtered by "Ban added".

4. **Public banlist** — goto `/index.php?p=banlist`, assert `data-testid="ban-row"` filtered by SteamID is visible.

The shared interactions live in `web/tests/e2e/pages/SubmitBanFlow.ts` so the gate spec and the `@screenshot flow-public-submission` block in `_screenshots.spec.ts` drive identical UI.

## Acceptance criteria

- [x] `web/tests/e2e/specs/flows/public-ban-submission.spec.ts` — full lifecycle covered.
- [x] `web/tests/e2e/pages/SubmitBanFlow.ts` — page-object helpers for the three pages.
- [x] `_screenshots.spec.ts` — `@screenshot flow-public-submission` block appended (3 routes × 2 themes × 2 viewports = 12 PNGs, plus Slice 0's 4 = 16 total).
- [x] Selectors use #1123 testability hooks; no CSS class chains, no visible-text primaries, no `setTimeout` / `waitForTimeout`.
- [x] `docker-compose.override.yml` gitignored (`sbpp-slice3-pubsub`, ports 8225–8229).
- [x] Existing specs unchanged; Slice 0 contract preserved.
- [x] axe `critical` threshold not downgraded.
- [x] No new npm deps.

## Foundation gaps surfaced (and fixed inline; all small, all flagged)

Slice 0 (#1162) explicitly noted "Slices that mutate state need a worker-scoped serialisation strategy because both projects (chromium + mobile-chromium) share the DB". This is the first stateful spec; three additional foundation gaps surfaced once the spec actually drove the flow end-to-end. All three fixes are tiny, targeted, and documented inline at the call sites.

### 1. `web/includes/Database.php` — enable native prepares

PDO's emulated-prepares default literal-substitutes array-shortcut `execute([...])` params as strings, so `LIMIT ?,?` becomes `LIMIT '0','30'` — invalid syntax in MariaDB. This **500'd both `/index.php?p=banlist` and `/index.php?p=commslist`** whenever the relevant tables had any rows, a regression latent since #1092's ADOdb→PDO refactor (ADOdb did its own type detection on array binds; PDO does not). The fix is one line — `PDO::ATTR_EMULATE_PREPARES => false` in the constructor — with a comment explaining the regression and why native prepares are safe (existing call sites that go through `Database::bind()` already auto-detect `PARAM_INT` for ints, so native prepares are purely additive on the array-shortcut path). PHPUnit stays green at 280 / 1189.

### 2. `sbpp.sh` `e2e` command — swap `config.php`'s `DB_NAME`

`docker/php/web-entrypoint.sh` renders `config.php` once at container start with `define('DB_NAME', 'sourcebans')` (the dev DB). Apache+mod_php holds that constant for the life of the container, so the existing `dc exec -e DB_NAME=sourcebans_e2e` only affects the bash shell that drives `npx playwright test` (and the truncate shim it shells into). Without this swap the panel the test browser hits writes to `sourcebans` while the fixture truncates `sourcebans_e2e` — the two halves of "DB isolation" point at different schemas. Apache re-reads `config.php` on every request so the in-place `sed` takes effect immediately with no restart; the `trap` restores the dev-DB value so the developer's browser session is unaffected once the suite finishes.

CI inherits the swap because CI's `e2e` job goes through the same `./sbpp.sh e2e` entrypoint.

### 3. `web/themes/default/page_admin_bans_submissions.tpl` — `__sbppApplyBanFields`

The inline ban-from-submission handler called the legacy `window.applyBanFields` (deleted with `web/scripts/sourcebans.js` in #1123 D1); the actual sbpp2026 helper is `window.__sbppApplyBanFields` (defined in `admin.bans.php`'s tail script). Without the fix the BansSetupBan response round-trips but the Add Ban form stays empty — silently breaking the entire "Ban a submission" UX. The spec would have to reach past the broken UI by typing the steam/nickname/reason itself — the "drive the API, not the UI" anti-pattern this slice exists to avoid. Fallback to the legacy name kept so a third-party theme that still loads `sourcebans.js` keeps working.

### Why no `truncateE2eDb()` here (a fourth gap, sidestepped not fixed)

`Sbpp\Tests\Fixture::truncateAndReseed()` walks every table (`TRUNCATE :prefix_admins` included) and re-seeds the admin at the end. There's a small window between the truncate and the re-INSERT where the admin doesn't exist, and any spec running in parallel that authenticates against the panel (currently `smoke/login.spec.ts`) sees the seam and fails its login. We get hermeticity from a per-run unique SteamID seed (`STEAM_0:1:${Date.now()}${workerIndex}`) instead — the panel only rejects re-submitting the *same* SteamID with `already_banned`, so a unique seed sidesteps the guard without touching the admin row at all. The cost is leftover rows in `:prefix_bans` / `:prefix_submissions` between specs, fine for the e2e DB (`globalSetup` does a full install at the start of each `playwright test` invocation).

The "worker-scoped serialisation strategy" Slice 0 promised stays a follow-up. When that helper lands, the gate spec can flip back to `truncateE2eDb()` in a `beforeEach` and the comment in this PR's spec explains the migration path.

## Documented divergences from the issue text

- The issue describes the approve action as `data-testid="row-action-approve"`. The actual hook in `page_admin_bans_submissions.tpl` is `data-testid="row-action-ban"` — clicking it loads the submission's data into the Add Ban form via `Actions.BansSetupBan`, and the admin then submits *that* form via `Actions.BansAdd` with `fromsub=<subid>` to insert the ban + flip the submission to `archiv=3`. The spec drives that real flow end-to-end via the UI; it does not introduce a new `row-action-approve` testid.

- `web/pages/page.submit.php` still emits `print "<script>ShowBox('Successful', …)</script>"` on success, but `ShowBox` was deleted with `sourcebans.js` in #1123 D1, so no visible "thank you" toast lands in the sbpp2026 chrome on success. Filed as #1176 for a separate fix; the spec anchors on the deterministic form-reset behaviour (`BanReason` textarea round-trips to empty on success) instead, so this slice is not blocked.

## Local commands run

| Gate | Result |
| --- | --- |
| `./sbpp.sh phpstan` | clean (211 files, no errors over baseline) |
| `./sbpp.sh test` | 280 tests, 1189 assertions, all green |
| `./sbpp.sh e2e` | 5 passed, 17 skipped (16 `@screenshot` variants + 1 mobile flow) |
| `SCREENSHOTS=1 ./sbpp.sh e2e --grep @screenshot` | 16 PNGs produced (4 login + 3×2×2 flow) |

Worktree-local `docker-compose.override.yml`: project `sbpp-slice3-pubsub`, ports 8225–8229.

## Screenshots

(Comment posted by the upload-screenshots.sh script.)

### DIFF
--- a/sbpp.sh
+++ b/sbpp.sh
@@ -148,13 +148,44 @@ case "$cmd" in
             GRANT ALL PRIVILEGES ON \`sourcebans_e2e\`.* TO 'sourcebans'@'%';
             GRANT CREATE, DROP ON *.* TO 'sourcebans'@'%';
             FLUSH PRIVILEGES;" >/dev/null
+        # Pin the panel at the e2e DB for the duration of the run, then
+        # restore on exit. docker/php/web-entrypoint.sh renders config.php
+        # once at container start with `define('DB_NAME', 'sourcebans')`
+        # — the dev DB. Apache+mod_php holds that constant for the life
+        # of the container, so the `-e DB_NAME=sourcebans_e2e` below only
+        # affects the bash shell that drives `npx playwright test` (and
+        # the truncate shim it shells into). Without this swap the panel
+        # the test browser hits writes to `sourcebans` while the fixture
+        # truncates `sourcebans_e2e`, and any spec that mutates DB state
+        # is non-hermetic across runs (#1124 Slice 3 was the first slice
+        # to exercise the write path; Slice 0's smoke specs only login,
+        # which masked the gap). Apache re-reads config.php on every
+        # request so an in-place sed takes effect immediately with no
+        # restart; the trap restores the dev-DB value so the developer's
+        # browser session is unaffected once the suite finishes.
+        e2e_db="${E2E_DB_NAME:-sourcebans_e2e}"
+        dc exec -T web bash -c "
+            set -e
+            cfg=/var/www/html/web/config.php
+            stash=/var/www/html/web/config.php.e2e-stash
+            [ ! -f \$stash ] && cp \$cfg \$stash
+            sed -i \"s/define('DB_NAME', '[^']*');/define('DB_NAME', '${e2e_db}');/\" \$cfg
+        "
+        restore_panel_db() {
+            dc exec -T web bash -c '
+                cfg=/var/www/html/web/config.php
+                stash=/var/www/html/web/config.php.e2e-stash
+                if [ -f $stash ]; then mv $stash $cfg; fi
+            ' 2>/dev/null || true
+        }
+        trap restore_panel_db EXIT INT TERM
         dc exec -T \
             -e E2E_BASE_URL="${E2E_BASE_URL:-http://localhost}" \
             -e E2E_IN_CONTAINER=1 \
             -e SCREENSHOTS="${SCREENSHOTS:-}" \
             -e CI="${CI:-}" \
             -e DB_HOST=db -e DB_PORT=3306 \
-            -e DB_NAME="${E2E_DB_NAME:-sourcebans_e2e}" \
+            -e DB_NAME="${e2e_db}" \
             -e DB_USER=sourcebans -e DB_PASS=sourcebans \
             -e DB_PREFIX=sb -e DB_CHARSET=utf8mb4 \
             web bash -lc '
--- a/web/includes/Database.php
+++ b/web/includes/Database.php
@@ -26,7 +26,20 @@ class Database
         $this->prefix = $prefix;
         $dsn = 'mysql:host=' . $host . ';port=' . $port . ';dbname=' . $dbname . ';charset=' . $charset;
         $options = array(
-            PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION
+            PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION,
+            // Native (non-emulated) prepares: PDOStatement::execute([...])
+            // forwards values via MySQL's binary protocol with proper
+            // type metadata, instead of literal-substituting them as
+            // strings client-side. The latter breaks `LIMIT ?,?` (MariaDB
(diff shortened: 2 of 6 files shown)


rumblefrog:
## Screenshots (flow-public-submission)

| `login` |  |  |  |  |

Issue #1207, opened 2026-05-05
UI/UX audit: cross-cutting findings across public + admin flows on origin/main


rumblefrog:
## Summary

Walked the panel end-to-end across the public + admin surfaces in both
themes and both viewports against `origin/main` (HEAD `c796e98a`). Found
~20 distinct UI/UX issues that span the application — most are
cross-cutting (topbar, drawer, theme contrast) so they recur on every
page rather than belonging to a single template. This issue is the
umbrella; per-finding child tickets can be split off as follow-ups.

Severity legend: **P0** breaks the surface for the indicated viewport ·
**P1** clearly broken or confusing UX · **P2** rough edge / consistency
· **P3** polish.

Every finding below carries its evidence inline. The full screenshot
gallery (every captured route × theme × viewport) is at the bottom for
completeness.

## How this was produced

1. Brought up an isolated parallel docker stack from a fresh `origin/main`
   worktree so this audit could run alongside other developer stacks
   (per the "Parallel stacks" section of `AGENTS.md`).
2. Ran `SCREENSHOTS=1 ./sbpp.sh e2e --grep @screenshot` for both projects
   (`chromium` and `mobile-chromium`). This is the same harness used to
   gate visual changes per PR (#1124), so the captured surfaces are the
   ones the team already agreed to lock.
3. Added a small additional spec (`web/tests/e2e/specs/_ux_audit_extra.spec.ts`,
   not committed) to also cover `?p=admin&c=mods|servers|overrides|email`
   and `?p=lostpassword`, which the static gallery doesn't currently
   include.
4. Pushed all PNGs to the `screenshots-archive` orphan branch under
   `screenshots/issue-uxaudit-20260505-1912/` so this issue renders the
   evidence inline.

---

### CC-1 — Topbar search input collapses on mobile widths (P0)

**Affects**: every page when viewed at mobile width (`mobile-chromium`
project). Files: `web/themes/default/_partials/topbar.tpl` (and the
search input wrapper inside it), `web/themes/default/css/topbar.css`.

The topbar packs `[hamburger] [breadcrumb] [search input + ⌘K badge]
[theme toggle]` onto a single row that is too narrow for any
sub-480px viewport. The placeholder text "Search players, SteamIDs..."
wraps to **3 lines** and bleeds outside the input bounds, the input
visually overlaps the breadcrumb on the left, and the `⌘K` badge
collides with the sun/moon icon on the right.

[...]

PR #1208, opened 2026-05-06
fix(theme): mobile chrome rescue (#1207 CC-1, CC-2, ADM-8, DET-1)

rumblefrog:
## Summary

Slice 1 of #1207 — mobile chrome rescue. Addresses CC-1, CC-2, ADM-8, DET-1.

- **CC-1** — collapse the topbar `.topbar__search` palette trigger to an icon-only button at `<=768px` (label + Ctrl-K hint hidden via `.topbar__search-label` / `.topbar__search-kbd`, aria-label + `data-palette-open` testability hook preserved).
- **CC-2** — full-width drawer + scrollable id grid + scrollable tab strip + declarative page-scroll-lock (`html:has(#drawer-root[data-drawer-open="true"]) { overflow: hidden; }`) on mobile so SteamID copy buttons stay reachable, "Notes" tab no longer clips, and the body underneath the dimmed overlay locks.
- **ADM-8** — admin sub-tab strip (`.admin-tabs`) flips to `flex-wrap: nowrap; overflow-x: auto` with `scroll-snap-type: x mandatory` at `<=768px`, and the active tab gets a chip-style background (`var(--brand-600)`) so it reads at a glance even with the underline partially out of view.
- **DET-1** — `<meta name="format-detection" content="telephone=no,date=no,address=no,email=no">` + `x-apple-data-detectors=false` in `core/header.tpl` opt the document out of mobile auto-tap-to-dial detection (the source of the pinkish Steam3 highlight); CSS `.drawer a[href^="tel:"]` reset is the defensive belt-and-suspenders for Android variants that ignore the meta.

Side-effect: AGENTS.md's "Playwright E2E specifics" rule said animations shouldn't gate visibility, but the chrome's `slide-in` keyframe didn't honour `prefers-reduced-motion: reduce`. Added a global CSS guard that collapses every `animation-duration` / `transition-duration` to `0.001ms` for reduced-motion users; updated `AGENTS.md` accordingly. Without this guard a Playwright `boundingBox()` read right after `[data-drawer-open="true"]` settles can land mid-translateX.

## Test plan

- [x] PHPUnit clean (`./sbpp.sh test` — 285 tests, 1211 assertions)
- [x] API contract clean (`./sbpp.sh composer api-contract` — no diff)
- [x] Playwright E2E clean (`CI=1 ./sbpp.sh e2e` — 93 passed, 171 `@screenshot`-only skipped, both projects)
- [x] New mobile specs added under `web/tests/e2e/specs/responsive/`:
  - `topbar.spec.ts` — icon-only collapse at iPhone-13, tap-opens-palette, no horizontal page overflow.
  - `drawer.spec.ts` — extended with three new tests covering CC-2 (copy-button-fits-viewport, page-scroll-lock) and DET-1 (Steam3-no-auto-link).
  - `admin-tabs.spec.ts` — single-line scrollable strip + chip-style active background.
- [x] Screenshots regenerated (`SCREENSHOTS=1 ./sbpp.sh e2e --grep @screenshot`) — see screenshot gallery comment below.
- [x] Manual check at 375px and 768px viewports in both light + dark themes (drawer at full-width, topbar icon-only, admin tabs scroll horizontally with chip-style active state, Steam3 renders as plain text).

## Docs

- `AGENTS.md` — added a "Playwright E2E specifics" sub-bullet describing the new CSS-side `prefers-reduced-motion: reduce` guard, an "Anti-patterns" entry for the `format-detection` meta + `tel:` reset, and four new "Where to find what" rows (mobile chrome layout, mobile auto-link prevention, page scroll lock, animation opt-out).

## Coordination

Worker 2's dark-theme contrast pass also touches `theme.css` — confined my changes to layout selectors (`@media (max-width: 768px)` blocks for `.topbar__search`, `.admin-tabs`, `.drawer`) and left token redefinitions alone. The `html.dark .admin-tabs > [aria-current="page"]` rule I added at mobile is the only dark-mode token reference; if it conflicts with Worker 2's dark-mode chip refinement, prefer their definition and reapply the `border-radius: var(--radius-md)` shape locally.

Closes nothing (umbrella issue #1207 stays open until all 5 slices land).


rumblefrog:
## Screenshots (mobile-chrome)

| `admin-audit` |  |  |  |  |
| `admin-bans` |  |  |  |  |
| `admin-home` |  |  |  |  |
| `banlist` |  |  |  |  |
| `commslist` |  |  |  |  |
| `home` |  |  |  |  |
| `login` |  |  |  |  |
| `myaccount` |  |  |  |  |
| `protest` |  |  |  |  |
| `servers` |  |  |  |  |
| `submit` |  |  |  |  |

rumblefrog (commented):
## Self-review of slice 1 of #1207 (CC-1 / CC-2 / ADM-8 / DET-1)

(Posted as `--comment` because GitHub doesn't let an author
request-changes on their own PR; treat the findings below as the
canonical request-changes set — I'm addressing them in follow-up
commits before merging.)

Solid CSS-first slice. The reduced-motion guard, the declarative
`:has()` scroll lock, and the two-layer DET-1 fix (meta + CSS reset)
are the kind of changes that make later regressions cheap to chase.
Anti-patterns + Where-to-find updates in `AGENTS.md` are well placed.
Three findings to address before merge — one is the brief's explicit
touch-target spec, the other two are test-coverage gaps that leave
parts of the new contract unlocked.

### Findings (must address)

**Finding 1 — CC-1 touch target is 36×36 px (brief: ≥44×44 px) — major**
- Location: `web/themes/default/css/theme.css:165-171`
- What: at `<=768px` the icon-only palette trigger is `width: 2.25rem`
  and inherits `height: 2.25rem` from the unscoped `.topbar__search`
  rule on line 145 — that's 36 × 36 CSS px. The reviewer brief for
  this slice explicitly requires `>=44x44 px` (WCAG 2.1 AAA / Apple
  HIG / Material). The desktop labelled button is wide enough that
  the touch-target spec passes incidentally; the icon-only mobile
  collapse is the affordance the brief specifically called out.
- Why: 44 px is the documented contract for this slice. 36 px is in
  the "OK on stylus, awkward on thumb" zone, and the existing
  `topbar.spec.ts` upper bound is `<=48` so a future shrink wouldn't
  fail any test either. The very first user action on an iPhone-13
  panel (open palette → search) shouldn't be the smallest tap target
  in the chrome.
- Suggested fix: bump both axes to `2.75rem` (44 px) at the mobile
  breakpoint, and tighten `topbar.spec.ts` to lock it:

  ```css
  @media (max-width: 768px) {
    .topbar__search {
      flex: 0 0 auto; min-width: 0;
      width: 2.75rem; height: 2.75rem;
      padding: 0; justify-content: center; gap: 0;
    }
  }
  ```

  ```ts
  // topbar.spec.ts — replace the loose width range
  expect(box!.width).toBeGreaterThanOrEqual(44);
  expect(box!.height).toBeGreaterThanOrEqual(44);
  expect(box!.width).toBeLessThanOrEqual(48);
  ```

**Finding 2 — Page scroll lock applies on desktop but is documented as a mobile fix — minor**
- Location: `web/themes/default/css/theme.css:452`
- What: `html:has(#drawer-root[data-drawer-open="true"]) { overflow: hidden; }`
  fires at every viewport. The audit's CC-2 explicitly framed scroll
  lock as a mobile finding ("at <=768px ... the page underneath is
  not scroll-locked behind the dimmed overlay"), and the surrounding
  `.drawer` rules are gated on `@media (max-width: 768px)`. The
  global lock on desktop is a behavior change (35rem drawer leaves
  the underlying page partially visible — locking page scroll there
  changes how a desktop user interacts with the drawer-open state)
  that isn't called out in the inline comment, the commit body, or
  `AGENTS.md`.
- Why: Either it's intentional modal-style behavior (fine, but the
  comment should say so) or it overscoped the mobile fix. Right now
  a desktop reader of the rule has no way to tell which.
- Suggested fix: pick one. Either expand the comment to "this also
  locks desktop scroll as part of the modal contract — desktop users
  expect drawer-as-modal" and back it with an explicit assertion in
  the desktop drawer spec, OR wrap the rule in
  `@media (max-width: 768px)` to match the rest of the CC-2 block.
  The desktop-modal pattern (Linear/Vercel/Notion) is the consistent
  choice — the asymmetric "locked on mobile, scrollable on desktop"
  variant is harder to reason about — but pick one and document it.

**Finding 3 — admin-tabs.spec.ts doesn't exercise the dark-theme branch — minor**
- Location: `web/tests/e2e/specs/responsive/admin-tabs.spec.ts:99-134`
- What: the PR adds a dark-theme override for the active chip
  (`html.dark .admin-tabs > [aria-current="page"] { background: var(--brand-500); color: var(--zinc-950); }`,
  `theme.css:286-289`), but the test only navigates to
  `?p=admin&c=bans` in the default light theme. The dark variant's
  contract is unlocked — a future refactor that drops the
  `var(--brand-500)` reference (e.g. tokens get renamed in slice 2's
  contrast pass) wouldn't fail any spec.
- Why: Per `AGENTS.md` "Testability hooks ... `<html class="dark">`
  for resolved theme" — the resolved theme is part of the contract
  this PR is locking.
- Suggested fix: extend the test (or add a sibling) that toggles
  dark mode (the seeded `data-testid="theme-toggle"` does this,
  or `await page.addInitScript(() => localStorage.setItem('theme', 'dark'))`
  before the goto) and re-runs the same R/G/B heuristic. The brand
  family check (`r > 180`, `b < 60`) already tolerates both
  `--brand-600` (light) and `--brand-500` (dark), so the only
  addition is the toggle.

### Things you did well

- The reduced-motion CSS guard + the matching `AGENTS.md` rule under
  "Playwright E2E specifics" closes the gap between
  `playwright.config.ts`'s `reducedMotion: 'reduce'` and the chrome
  side. Future flakes on `boundingBox()` mid-animation are now a
  one-line fix.
- Two-layer DET-1 (`<meta name="format-detection">` + defensive
  `.drawer a[href^="tel:"]` reset) plus the spec that asserts both
  the meta presence AND no `<a>` descendants is the right shape.
  The decision to NOT inject a fake `tel:` anchor for the CSS branch
  is well-explained in the comment; agreed.
- Declarative `html:has(#drawer-root[data-drawer-open="true"])`
  scroll lock instead of a JS body-class hack — leaves a half-open
  drawer recoverable, unwinds automatically on close, no event
  listeners to leak. (Modulo Finding 2 about the scope.)
- Every test selector uses testability hooks
  (`data-testid="palette-trigger"`, `data-palette-open`,
  `[data-drawer-open="true"]`, `data-testid="drawer-tab-*"`,
  `[aria-current="page"]`); zero CSS class chains, zero
  visible-text-as-primary selectors. `hasText` filters only
  disambiguate when the primary selector matches multiple nodes.
- `AGENTS.md` updates land in the same PR (anti-patterns row for
  the format-detection meta, four "Where to find what" rows for
  the new layout/lock/animation hooks, the Playwright specifics
  paragraph). No follow-up doc PR needed.
- The spec waits on terminal attributes (`[data-drawer-open="true"]`,
  `not.toHaveAttribute('data-loading', 'true')`) — no
  `setTimeout`/`waitForTimeout` anywhere.

rumblefrog:
Addressed all three review findings in `e3836787`. CI re-running.

- **Finding 1 (major) — touch target** — bumped the icon-only mobile palette trigger from 2.25rem (36 px) to 2.75rem (44 px) on both axes (`web/themes/default/css/theme.css`). Tightened `topbar.spec.ts` from the loose `> 20` / `<= 48` width range to `>= 44` on both axes, with an upper bound of `<= 56` to catch accidental re-expansion via cascade drift.
- **Finding 2 (minor) — scroll-lock scope** — kept the rule universal (it is intentional modal-style behaviour, matching Linear/Vercel/Notion), but expanded the inline `theme.css` comment to call that out explicitly, updated the AGENTS.md "Where to find what" row, and added a desktop counterpart test `player drawer › opening the drawer locks page scroll on desktop too (#1207 CC-2)` in `flows/ui/player-drawer.spec.ts` so a viewport-conditional regression now lands on a different failing spec.
- **Finding 3 (minor) — dark-theme coverage** — refactored the rgb heuristic into a shared `assertActiveTabIsBrandOrange()` helper, kept the existing light test, and added a sibling that seeds `localStorage['sbpp-theme'] = 'dark'` via `page.addInitScript()` before the first navigation so theme.js's init-time `applyTheme(currentTheme())` resolves to `html.dark` on first paint without a click race.

Quality gates locally:
- PHPStan ✓
- PHPUnit ✓ (285 / 1211)
- ts-check ✓
- Playwright `CI=1` ✓ — 95 passed (was 93; +1 desktop scroll-lock spec, +1 dark-theme admin-tabs spec).

rumblefrog (commented):
## Review of slice 1 of #1207 (mobile chrome rescue)

All four findings (CC-1, CC-2, ADM-8, DET-1) are addressed against the
audit's "Suggested fix" lines. Comments and tests are unusually thorough
— good docs in the diff. CI gates green (PHPStan, PHPUnit, ts-check,
Playwright + axe); api-contract gate didn't run because the diff doesn't
touch handlers / contract / generator paths, which is correct.

Findings below are all `minor` / `nit` — none block, but I'll address
the minors before merge to keep the testability hooks consistent with
the project's #1123 contract.

### Findings

**Finding 1 — minor**
- Location: `web/tests/e2e/specs/responsive/topbar.spec.ts:105`
  (`page.locator('.topbar')`); `web/tests/e2e/specs/responsive/admin-tabs.spec.ts:45`
  (`page.locator('.admin-tabs').first()`).
- What: Both new specs use a bare CSS class chain (`.topbar`, `.admin-tabs`)
  as the **primary** selector. The `.admin-tabs > [aria-current="page"]`
  use a few lines down is fine because it pivots on the ARIA attribute,
  but the standalone class refs aren't disambiguated by an ARIA or
  `data-testid` hook.
- Why: AGENTS.md "Anti-patterns" → "CSS class chains or visible-text
  *primary* selectors in E2E specs → use `data-testid` / ARIA roles
  per #1123." Existing scoping uses (`#banlist-root .ban-cards
  [data-testid="…"]`) combine class scope with a `data-testid` leaf;
  here we have a class only.
- Suggested fix: add `data-testid="topbar"` to the `<header
  class="topbar">` in `web/themes/default/core/title.tpl`, and
  `data-testid="admin-tabs"` to the `<div class="admin-tabs …">` in
  `web/themes/default/core/admin_tabs.tpl`. Update the two spec
  references to use the testid. Ten extra characters and the contract
  matches the rest of the suite.

**Finding 2 — minor**
- Location: `web/themes/default/css/theme.css:431` —
  `.drawer .drawer__ids dd, .drawer .drawer__ban dd { min-width: 0;
  overflow-wrap: anywhere; word-break: break-word; }`
- What: `word-break: break-word` is non-standard / [deprecated by the
  CSS WG](https://drafts.csswg.org/css-text/#word-break-property)
  (the standard equivalent is `overflow-wrap: anywhere`, which the
  rule already declares). Keeping both is redundant and signals that
  the author was unsure which one to use.
- Why: The `overflow-wrap: anywhere` declaration on the same rule
  already gives us the desired wrap-on-any-character behaviour at
  every browser the suite targets (Chrome 105+, Safari 15.4+, Firefox
  121+ — the same baseline `:has()` requires below). Trimming the
  deprecated alias keeps the rule honest.
- Suggested fix: drop `word-break: break-word;` from the declaration.

**Finding 3 — minor**
- Location: `web/tests/e2e/specs/responsive/drawer.spec.ts:224-275`
  (`Steam3 value renders as plain text without an auto-link (DET-1)`).
- What: The test runs against `mobile-chromium`, which is Chrome
  devtools' mobile emulation — it doesn't replicate the iOS Safari
  / Android Chrome **heuristic** that injects `<a href="tel:…">` into
  digit-/colon-heavy strings. The
  `await expect(steam3Value.locator('a')).toHaveCount(0)` assertion
  therefore passes regardless of whether the meta tag opt-out is
  present, because the emulator never injects in the first place.
- Why: The test still locks the meta tag's presence (the regex
  `toHaveAttribute('content', /telephone=no/)` fails on removal), so
  it's a useful regression guard. But a future maintainer reading the
  test name "renders as plain text without an auto-link" might think
  it actually exercises the iOS Safari behaviour. The block comment
  could be more honest about what the assertion does and doesn't
  catch.
- Suggested fix: prepend a one-liner to the existing block comment:
  "Note: mobile-chromium doesn't reproduce the iOS Safari heuristic;
  the `.toHaveCount(0)` assertion only fails if some upstream change
  injects an anchor literally. The meta-tag regex above is the load-
  bearing assertion."

**Finding 4 — nit**
- Location: `web/themes/default/css/theme.css:472` (the `:has(...)`
  scroll-lock rule) and `:434` (the mobile `:has(…) body` companion).
- What: First usage of `:has()` in the codebase. Modern browsers
  support it (Chrome 105+ / Safari 15.4+ / Firefox 121+ — Web
  Baseline 2023), but the project doesn't have an explicit baseline
  statement and the rule is the first to require it.
- Why: A self-hoster on an older corporate-mandated browser would
  silently lose the scroll lock. The drawer still functions (the
  comment notes the failure mode is "lock-not-applied" rather than
  "drawer broken"), so this isn't a behavioural blocker — but worth
  acknowledging the baseline shift somewhere a casual reader will
  look.
- Suggested fix: extend the existing block comment at line 448 with
  one line: "Browser baseline: `:has()` is supported in Chrome 105+
  / Safari 15.4+ / Firefox 121+; older browsers silently lose the
  scroll lock (drawer still opens, page scrolls behind it)." Or
  drop a one-liner row in AGENTS.md "Where to find what" if you
  want it more discoverable. The existing comment is otherwise
  excellent.

### Things you did well

- **Self-review pass already addressed three findings.** The 44px
  touch-target floor (with both axes locked, plus a 56px upper bound
  to prevent cascade drift), the desktop scroll-lock counterpart spec,
  and the dark-theme admin-tabs sibling spec. Each was a real
  observation; the lock-spec and the dark sibling are exactly the
  kind of "regression in either direction lands on a different
  failing test" structure the project's testability hooks reward.
- The `prefers-reduced-motion: reduce` global guard (theme.css :398)
  is a quietly load-bearing addition — Playwright sets the media query
  globally per `playwright.config.ts`, but the chrome's keyframes
  weren't honouring it on the CSS side. Documenting it in AGENTS.md
  alongside the existing "Animations should never gate visibility"
  rule keeps the contract symmetric across CSS / JS.
- The `format-detection` opt-out via meta + the defensive
  `.drawer a[href^="tel:"]` reset is the right shape for DET-1: the
  meta is the canonical primary, the CSS reset is the belt-and-
  suspenders for Android variants. The "Anti-patterns" entry in
  AGENTS.md preventing future removal is appropriate.
- New "Where to find what" rows are specific enough to land an agent
  on the right `@media (max-width: 768px)` block in one read.
- Existing `data-palette-open` / `data-testid="palette-trigger"` /
  `data-drawer-open` hooks all preserved through the icon-only
  collapse — the trigger keeps working unchanged for both desktop
  click and mobile tap.

rumblefrog:
Addressed all four review findings in 6cb3cc2c:

- **Finding 1 (minor)** — Added `data-testid="topbar"` to
  `core/title.tpl` and `data-testid="admin-tabs"` to
  `core/admin_tabs.tpl`. Repointed `responsive/topbar.spec.ts` and
  `responsive/admin-tabs.spec.ts` to use the new testid hooks (the
  `.admin-tabs > [aria-current="page"]` chain becomes
  `[data-testid="admin-tabs"] > [aria-current="page"]`).
- **Finding 2 (minor)** — Dropped the deprecated `word-break:
  break-word` alias from `.drawer .drawer__ids dd, .drawer
  .drawer__ban dd`; the standard `overflow-wrap: anywhere` already
  on the rule does the work. Added a one-paragraph block comment
  explaining the load-bearing pair (`min-width: 0` + `overflow-wrap:
  anywhere`).
- **Finding 3 (minor)** — Extended the in-test block comment on
  `Steam3 value renders as plain text without an auto-link (DET-1)`
  with a "Coverage caveat" paragraph noting that mobile-chromium
  doesn't replicate the iOS Safari heuristic the fix addresses;
  the `toHaveAttribute('content', /telephone=no/)` regex match is
  the load-bearing assertion, the no-anchor check is a literal-
  injection guard.
- **Finding 4 (nit)** — Extended the existing block comment above
  the `:has(...)` rule in `theme.css` with a "Browser baseline
  note" paragraph: this is the codebase's first `:has()` usage
  (Web Baseline 2023 — Chrome 105+ / Safari 15.4+ / Firefox 121+);
  older browsers silently lose the scroll lock and the page
  scrolls behind the drawer.

Quality gates locally:
- PHPStan clean
- PHPUnit clean (285 tests, 1211 assertions)
- ts-check clean
- Playwright `CI=1` clean (95 passed across both projects under
  workers:1).

CI re-running.

PR #1209, opened 2026-05-06
fix(theme): dark-theme contrast pass (#1207 CC-4, AUTH-2)

### LINKED ISSUE #1207: UI/UX audit: cross-cutting findings across public + admin flows on origin/main

rumblefrog:
## Summary

Slice 2 of #1207 — dark-theme contrast pass. Addresses CC-4 + AUTH-2.

- **CC-4** — Sidebar nav, banlist filter chips, and admin/bans `Current / Archive` segmented chips now paint `var(--accent)` (`--brand-600`, the same orange the primary CTA uses) when active in dark mode. The original `bg: var(--zinc-100)` near-white pill blended with the panel-dark surface and read as "hovered" rather than "selected" in the audit screenshots. Light theme keeps the existing zinc-900 pill on zinc-50 page (which has plenty of contrast).
- **AUTH-2** — "Continue with Steam" picks up Steam's brand chrome (`#1b2838` / `#2a475e` hover) in dark mode so it stops reading as disabled next to the orange "Sign in" CTA. Light theme keeps the secondary surface, which already has 4.5:1+ against the white card.

CSS-only — no templates, JS, or PHP touched. Reuses the existing #1123 testability hooks (`[aria-current="page"]`, `[aria-pressed="true"]`, `[data-active="true"]`, `[data-testid="login-steam"]`) so the rules are naturally scoped without new selectors.

No new CSS variables introduced. The fallback in `var(--on-accent, #fff)` keeps the rule self-contained — a follow-up that introduces a real `--on-accent` token can drop the literal without changing the computed colour.

### Coordination

- Worker 1 (`fix/issue-1207-mobile-chrome`) owns layout `@media` rules — this slice doesn't touch any. The `prefers-reduced-motion` guard the audit gate relies on is theirs; my E2E spec uses `expect.toHaveCSS` (which polls) instead of a one-shot `getComputedStyle` read so it survives transitions even before that guard lands.
- Worker 4 (banlist + queue tables) — my filter-pill change touches `.chip[aria-pressed="true"]` / `.chip[data-active="true"]` *background-color and border-color only* in dark mode. Layout / column-width / chip-row geometry stays untouched.

## Test plan

- [x] PHPStan clean (no PHP changed)
- [x] PHPUnit clean — `285 tests, 1211 assertions`, no regressions
- [x] ts-check clean
- [x] API contract clean (regen produced no diff)
- [x] Playwright E2E clean in CI mode (`workers: 1`): `90 passed`, `168 skipped`. Local default (`workers: 8`) had pre-existing flakes in `flows/public-ban-submission.spec.ts` and `flows/ui/{command-palette,player-drawer}` that pass when re-run in isolation — those are the DB-race surfaces AGENTS.md calls out as the reason CI uses `workers: 1`.
- [x] axe gate stays green at 0 critical (the new colour is identical to the existing primary CTA's brand-600 / white pairing, which axe already accepts).
- [x] New spec `specs/a11y/dark-theme-contrast.spec.ts` locks the computed `background-color` via `expect.toHaveCSS` for: dark sidebar active nav, dark banlist `All` chip, dark admin/bans `Current` chip, dark Steam login button, and (regression guard) light sidebar active nav.
- [x] Screenshot gallery regenerated and pushed to `screenshots-archive` — see comment below for the per-slice gallery.

Does NOT close #1207 — that umbrella tracks the remaining slices.

### DIFF
--- a/web/tests/e2e/specs/a11y/dark-theme-contrast.spec.ts
+++ b/web/tests/e2e/specs/a11y/dark-theme-contrast.spec.ts
@@ -6,12 +6,19 @@
  * audit screenshots flagged ("hovered" rather than "selected"):
  *
  *   - CC-4: Sidebar nav, banlist filter chips, and admin/bans
- *     `Current / Archive` segmented chips all paint `var(--accent)`
- *     (the same `--brand-600` the primary CTA uses) when active.
+ *     `Current / Archive` segmented chips paint `--brand-700`
+ *     (`#c2410c`) when active in dark mode. We use brand-700 rather
+ *     than `--accent` (`--brand-600` = `#ea580c`) for accessibility:
+ *     `#ea580c` on white is ~3.56:1, which clears WCAG AA Large Text
+ *     and Non-text but FAILS AA Normal Text (4.5:1) for the 14px /
+ *     12px medium-weight nav + chip labels. `#c2410c` on white is
+ *     ~5.18:1 — clears AA Normal Text. See `theme.css` rule comments
+ *     for the full rationale.
  *   - AUTH-2: The "Continue with Steam" button paints Steam's brand
  *     chrome (`#1b2838`) in dark mode so it doesn't read as the
  *     `--bg-surface`-on-`--bg-page` near-disabled rectangle from the
- *     audit (~1.4:1 against the page background).
+ *     audit (~1.4:1 against the page background). White on `#1b2838`
+ *     ≈ 14.93:1, clears AAA.
  *
  * We assert via `expect(locator).toHaveCSS('background-color', …)`
  * rather than a one-shot `getComputedStyle()` read because theme.js
@@ -26,9 +33,10 @@
  * the DOM contract intact.
  *
  * Resolved colour values:
- *   - `--accent` = `--brand-600` = `#ea580c` = `rgb(234, 88, 12)`
+ *   - `--brand-700` = `#c2410c` = `rgb(194, 65, 12)` (active dark fill)
  *   - Steam dark chrome = `#1b2838` = `rgb(27, 40, 56)`
  *   - `--zinc-900` = `#18181b` = `rgb(24, 24, 27)` (light-theme guard)
+ *   - `--bg-surface` (light) = `#ffffff` = `rgb(255, 255, 255)`
  *
  * Theme pinning mirrors `_screenshots.spec.ts` / `a11y/routes.spec.ts`:
  * write `localStorage['sbpp-theme']` on `/`, then navigate to the
@@ -37,14 +45,15 @@
  *
  * Project filter: chromium-only. The mobile-chromium project shares
  * the same markup and computed-style computation; the rules under
- * test are token-driven (`var(--accent)`) and don't change with
- * viewport, so running on both would double the runtime without
- * revealing different findings.
+ * test are token-driven and don't change with viewport, so running
+ * on both would double the runtime without revealing different
+ * findings.
  */
 
 import { expect, test } from '../../fixtures/auth.ts';
+import type { Browser, BrowserContext, TestInfo } from '@playwright/test';
 
-const ACCENT_RGB = 'rgb(234, 88, 12)';
+const BRAND_700_RGB = 'rgb(194, 65, 12)';
 const STEAM_DARK_RGB = 'rgb(27, 40, 56)';
 const WHITE_RGB = 'rgb(255, 255, 255)';
 const ZINC_900_RGB = 'rgb(24, 24, 27)';
@@ -76,6 +85,29 @@ async function pinTheme(
     }, mode);
 }
 
+/**
+ * Anonymous context for tests that exercise logged-out chrome (the
+ * Steam login button is only rendered when no JWT cookie is present).
+ *
+ * We pass `reducedMotion: 'reduce'` explicitly because
+ * `playwright.config.ts` only sets it on the top-level `use`, not on
+ * any individual project — `testInfo.project.use` does not surface
+ * top-level `contextOptions`. Without this, a fresh anonymous context
+ * runs with motion enabled and silently drops the contract AGENTS.md
+ * "Playwright E2E specifics" calls out by name. `toHaveCSS` polls so
+ * the assertion survives either way, but the contract should hold.
+ */
+async function newAnonContext(
+    browser: Browser,
+    testInfo: TestInfo,
+): Promise<BrowserContext> {
+    return browser.newContext({
+        ...testInfo.project.use,
+        storageState: { cookies: [], origins: [] },
+        reducedMotion: 'reduce',
+    });
+}
+
 test.describe('#1207 dark-theme contrast', () => {
     test.beforeEach(({}, testInfo) => {
         test.skip(
@@ -84,7 +116,7 @@ test.describe('#1207 dark-theme contrast', () => {
         );
     });
 
-    test('CC-4 sidebar active link paints --accent', async ({ page }) => {
+    test('CC-4 sidebar active link paints --brand-700 in dark mode', async ({ page }) => {
         await pinTheme(page, 'dark', '/');
 
         // The active marker is set by navbar.tpl on whichever
@@ -94,11 +126,11 @@ test.describe('#1207 dark-theme contrast', () => {
         // attribute is in the DOM before the assertion runs.
         const active = page.locator('.sidebar__link[aria-current="page"]').first();
         await expect(active).toBeVisible();
-        await expect(active).toHaveCSS('background-color', ACCENT_RGB);
+        await expect(active).toHaveCSS('background-color', BRAND_700_RGB);
         await expect(active).toHaveCSS('color', WHITE_RGB);
     });
 
-    test('CC-4 banlist filter chip "All" paints --accent when pressed', async ({ page }) => {
+    test('CC-4 banlist filter chip "All" paints --brand-700 when pressed', async ({ page }) => {
         await pinTheme(page, 'dark', '/index.php?p=banlist');
 
         // The "All" chip is the default-selected filter in
@@ -109,11 +141,11 @@ test.describe('#1207 dark-theme contrast', () => {
         // assuming the first chip is the active one.
         const active = page.locator('.chip[aria-pressed="true"]').first();
         await expect(active).toBeVisible();
-        await expect(active).toHaveCSS('background-color', ACCENT_RGB);
+        await expect(active).toHaveCSS('background-color', BRAND_700_RGB);
         await expect(active).toHaveCSS('color', WHITE_RGB);
     });
 
-    test('CC-4 admin/bans Current sub-tab paints --accent', async ({ page }) => {
+    test('CC-4 admin/bans Current sub-tab paints --brand-700 in dark mode', async ({ page }) => {
         await pinTheme(page, 'dark', '/index.php?p=admin&c=bans');
 
         // The admin/bans page emits two `chip-row` segmented groups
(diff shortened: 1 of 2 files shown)


rumblefrog:
## Screenshots (dark-theme-contrast)

Per-PR gallery for slice 2 of #1207. The two surfaces flagged by the audit:

### CC-4 — sidebar nav + filter pills (dark theme)

| Route | Light desktop (reference) | Dark desktop | Light mobile | Dark mobile |
| `home` (sidebar active) |  |  |  |  |
| `banlist` (`All / Permanent / Active / Expired / Unbanned` chips) |  |  |  |  |
| `admin-bans` (`Current / Archive` sub-tabs) |  |  |  |  |

### AUTH-2 — Steam login button (dark theme)

| `login` |  |  |  |  |

### Full gallery (every captured route × theme × viewport)

<details>
<summary>Click to expand</summary>

| `admin-audit` |  |  |  |  |
| `admin-bans` |  |  |  |  |
| `admin-home` |  |  |  |  |
| `banlist` |  |  |  |  |
| `commslist` |  |  |  |  |
| `home` |  |  |  |  |
| `login` |  |  |  |  |
| `myaccount` |  |  |  |  |
| `protest` |  |  |  |  |
| `servers` |  |  |  |  |
| `submit` |  |  |  |  |

</details>

rumblefrog (commented):
## Self-review pass on slice 2 of #1207 (CC-4 + AUTH-2)

GitHub blocks `--request-changes` on your own PR; posting as a comment review with the same severity bands. Will fix Findings 1–3 in a follow-up commit on this branch before merging. Finding 4 is a declinable nit; see disposition below.

### Findings (must address)

**Finding 1 — major — Active sidebar/chip text fails WCAG AA contrast in dark mode**
- Location: `web/themes/default/css/theme.css:137`, `:330`
- What: The new dark-mode active rule paints `--accent` (`--brand-600` = `#ea580c`) with `#fff` text. The computed contrast is **3.56:1**.
  - Sidebar nav text: `font-size: 0.875rem` (14px), `font-weight: 500` → normal text by WCAG → AA requires 4.5:1.
  - Chip text: `font-size: 0.75rem` (12px), `font-weight: 500` → normal text → AA requires 4.5:1.
  - 3.56:1 only clears AA Large Text (≥18pt or ≥14pt bold) and AA Non-text (3:1). It fails AA Normal Text (4.5:1).
- Why this slipped past CI: `fixtures/axe.ts` thresholds at `critical` (per #1124). axe's `color-contrast` rule reports at `serious`, so a 3.56:1 hit is silently accepted. The reviewer prompt asks to compute the actual contrast — the math here doesn't pass.
- Why I'm not just accepting it as "consistent with the existing primary CTA": the audit's stated goal for CC-4 is improved contrast / improved selected affordance. Cloning the CTA's 3.56:1 perpetuates a known subthreshold pairing across **three more surfaces** (sidebar nav, banlist filter chips, admin/bans Current/Archive), each carrying user-readable text. The `.btn--primary` itself is out of scope for this slice, but propagating its colour without a contrast-fixed variant means we now owe a follow-up to fix four surfaces instead of one.
- Suggested fix: paint the dark-mode active state with `--brand-700` (`#c2410c`) instead of `--accent`. Computed contrast against white is **5.18:1** — clears AA Normal Text. Visually it's a half-step darker than the CTA, which actually helps "selected" read distinctly from "the orange Save button next to me". Keep the `var(--on-accent, #fff)` fallback shape for the text colour so a future `--accent-strong` token swap is a one-liner. Update the spec's `ACCENT_RGB` constant accordingly.

**Finding 2 — minor — Light-theme regression guard only covers sidebar; chip + Steam rules are also `html.dark`-scoped**
- Location: `web/tests/e2e/specs/a11y/dark-theme-contrast.spec.ts:160-173`
- What: The light-theme assertion only locks the sidebar pill (`zinc-900` / white). The two other dark-only rules (`html.dark .chip[aria-pressed="true"]` and `html.dark .btn[data-testid="login-steam"]`) also rely on the `html.dark` scoping for the light theme to stay correct. If a future change accidentally drops the `html.dark` prefix (or moves the rule outside its block), only the sidebar regression guard fires.
- Suggested fix: extend the regression case to assert (a) the active chip in light mode resolves `background-color: rgb(24, 24, 27)` (zinc-900), and (b) the Steam login button in light mode keeps `background-color: rgb(255, 255, 255)` (`var(--bg-surface)`). Cheap, ~10 extra lines, removes the asymmetric coverage.

**Finding 3 — minor — Steam button context spread can drop the `reducedMotion` contract**
- Location: `web/tests/e2e/specs/a11y/dark-theme-contrast.spec.ts:139-142`
- What: `browser.newContext({ ...testInfo.project.use, … })` only inherits the project-level `use`. The `contextOptions: { reducedMotion: 'reduce' }` flag from `playwright.config.ts` is set at the top level (`use`), not per-project, and isn't included in `testInfo.project.use`. The new context can run with motion enabled.
- Why this isn't a major: `expect.toHaveCSS` polls, so the assertion still lands on the settled colour. But the spec implicitly drops a contract that AGENTS.md "Playwright E2E specifics" calls out by name ("`prefers-reduced-motion: reduce` is set globally").
- Suggested fix: pass `reducedMotion: 'reduce'` explicitly on `browser.newContext`. Same shape as `a11y/routes.spec.ts`'s anonymous-context branch will need when its `route.auth = false` cell runs into the same gap (filed mentally as a follow-up — not in this PR's scope).

**Finding 4 — nit (declinable) — `[data-testid="login-steam"]` used as a CSS selector**
- Location: `web/themes/default/css/theme.css:234`
- What: AGENTS.md frames `data-testid` as a testability hook (consumed by Playwright + the screenshot harness), not a styling hook. The PR's choice to scope the dark-only Steam paint via the testid couples the visual contract to the test contract — a future testid rename would silently flip the dark Steam button back to the unreadable secondary chrome.
- Disposition (declined for this slice, with rationale): the testid is the only stable, unique hook on that button today, and adding a `.btn--steam` modifier requires a template touch which the audit's CSS-only suggestion deliberately avoided. No other rule in `theme.css` uses `data-testid` as a selector — this is the first. I'll leave the rule as-is and call it out below for AGENTS.md follow-up; if the next reviewer wants a `.btn--steam` modifier, that's a one-line template change + one-line CSS tweak in a separate PR. Will add a comment block in `theme.css` documenting the rationale so the convention is at least findable.

### Things this PR did well

- CSS-only blast radius. No template / JS / PHP touched, so the parallel slices (#1210/#1211/#1212) can't collide.
- Reused `[aria-current="page"]`, `[aria-pressed="true"]`, `[data-active="true"]`, and the `data-testid="login-steam"` hook — every selector under test was already in the DOM. The visual contract is the only thing changing.
- `var(--on-accent, #fff)` fallback shape future-proofs the rule for a real `--on-accent` token without re-editing this CSS later.
- Comments at every change point say *why* the original colour was wrong and *why* this paint fixes it.
- Spec uses `expect.toHaveCSS` (polls) instead of a one-shot `getComputedStyle()` read — survives theme.js's tail-of-document boot timing without `setTimeout`.
- Spec correctly spins up an anonymous context for the Steam button (the button only renders when logged out) and tears it down in a `try/finally`.
- `_screenshots.spec.ts` parity for the theme-pinning helper (`localStorage` → `goto` → `waitForFunction` on `<html.dark>`).
- Steam dark chrome `#1b2838` on white = **14.93:1** — clears AAA. AUTH-2 is genuinely fixed.
- Light theme is unchanged on every surface (verified by reading the diff and the regression guard).

rumblefrog:
Addressed self-review findings on `683cf196`. CI re-running.

- **Finding 1 (major) — addressed.** `html.dark .sidebar__link[aria-current="page"]` and `html.dark .chip[aria-pressed=true|data-active=true]` now paint `--brand-700` (`#c2410c`, the existing `--accent-hover` token) instead of `--accent` (`--brand-600`). White on `#c2410c` is ~5.18:1 — clears WCAG AA Normal Text. White on `#ea580c` was ~3.56:1 (cleared AA Large but not AA Normal at 14px / 12px medium-weight). The CSS comment block documents the contrast math inline so the next reader sees why we deviated from the audit's literal `bg: var(--accent)` suggestion. The `var(--on-accent, #fff)` fallback shape is preserved so a future `--on-accent` token swap is a one-liner. AUTH-2's Steam paint (`#1b2838`) is unchanged (already ~14.93:1).
- **Finding 2 (minor) — addressed.** Light-theme regression guard now covers all three `html.dark`-scoped rules: sidebar (zinc-900 background, white text), chip (zinc-900 background, white text), and Steam button (`--bg-surface` = white). One test per surface so an un-scoping of any single rule attributes the failure to the right `theme.css` line.
- **Finding 3 (minor) — addressed.** Factored `newAnonContext(browser, testInfo)` helper that passes `reducedMotion: 'reduce'` explicitly to `browser.newContext`. `testInfo.project.use` doesn't surface the top-level `contextOptions` from `playwright.config.ts`, so the previous spread silently dropped the global motion-reduce contract. `toHaveCSS` polls so the assertions weren't broken, but the contract now holds.
- **Finding 4 (nit) — declined with documentation.** Kept `[data-testid="login-steam"]` as the CSS selector for the Steam rule (the audit's CSS-only-suggested fix didn't want a template touch, the testid is unique to that button, and no other rule in `theme.css` uses the testid pattern). Added an inline comment block in the rule explaining the trade-off and the swap path if a future PR introduces a `.btn--steam` modifier.

Local gates re-ran clean against the parallel `sbpp-1209-rev` stack:
- PHPStan: no errors (`web/phpstan.neon`, level 5 + dba)
- PHPUnit: 285 tests / 1211 assertions, no regressions
- ts-check: clean
- API contract regen: no diff
- Playwright `specs/a11y` (chromium): all 35 axe-routes + 7 new contrast tests pass; mobile-chromium skips per the project filter.

The screenshot gallery in the comment above was captured against the original `--accent` paint. The visual delta from `#ea580c` → `#c2410c` is one step darker on the same hue — the "selected vs hovered" affordance the audit asked for is still clearly intact. Will regenerate the gallery and post a follow-up comment if it produces visibly different shots.

PR #1216, opened 2026-05-06
fix(theme): comms list affordances (#1207 slice 6)

### LINKED ISSUE #1207: UI/UX audit: cross-cutting findings across public + admin flows on origin/main

rumblefrog:
## Summary

Slice 6 of #1207 — comms list affordances. Addresses ADM-5, ADM-6.

- **ADM-5** — Edit / Unmute / Remove on the comms list are visible by default at every viewport. The `.row-actions { opacity:0 → 1 on hover }` trick is gone (it never worked on touch, and was hostile to reduced-motion / keyboard / screen-reader users). The desktop `<td class="col-actions">` and the new mobile `.ban-card__actions` row both expose explicit text-labelled buttons.
- **ADM-6** — Lifting an active block updates the row in place: `data-state` → `unmuted`, status pill swaps `pill--active` → `pill--unmuted` and reads "Unmuted", the `Unmute` button is replaced by a `Re-apply` anchor (threaded through the existing `comms.prepare_reblock` flow), and a `success` toast confirms via `window.SBPP.showToast`. New JSON twins `comms.unblock` + `comms.delete` drive the click handlers; the legacy `?p=commslist&a=ungag|unmute|delete&id=…&key=…` GET URL ships as `data-fallback-href` so no-JS callers + third-party themes that strip api.js still work.

Mobile cards were split from a single card-wrapping `<a>` into a `.ban-card` div with a clickable `.ban-card__summary` anchor + a sibling `.ban-card__actions` row, so each action is its own `<button>` / `<a>` instead of nested-interactive HTML.

## Test plan

- [x] PHPUnit clean (`./sbpp.sh test` — 325 tests, 1332 assertions)
- [x] API contract clean (`./sbpp.sh composer api-contract`; the regenerated `web/scripts/api-contract.js` is in this PR)
- [x] Playwright E2E clean (`CI=1 ./sbpp.sh e2e` — 138 passed; new `flows/comms-affordances.spec.ts` covers desktop visible-actions + in-place flip + Re-apply, mobile-card mirror, and Remove)
- [x] Screenshot gallery regenerated and pushed (see comment with the per-PR table; `flow-comms-gag-mute-list-active`/`-list-unblocked` now show the visible action buttons + the Re-apply swap)
- [x] Manual: comms list row Edit / Unmute / Remove buttons visible without hover; clicking Unmute updates row in-place + toast; row's `unmuted` state hides Unmute and surfaces Re-apply

Does NOT close #1207 — umbrella tracks remaining slices.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -522,6 +522,15 @@ audit (#1207) locked in. New CTAs:
   → use `data-testid` / ARIA roles per #1123. `hasText` filters for
   disambiguation are fine; "find element by its label text" as the
   whole selector is not.
+- Hover-only row-action affordances (`.row-actions { opacity: 0 }`
+  flipped to `1` on `tbody tr:hover`) → row-level Edit / Unmute /
+  Remove (and equivalent) buttons must be visible by default at
+  every viewport. The opacity-on-hover trick was removed from the
+  comms list in #1207 ADM-5; it never worked on touch viewports
+  (no hover state) and silently regressed discoverability for
+  every keyboard / screen-reader user. New surfaces add visible
+  buttons in the same shape as `.queue-row` (admin moderation
+  queue) or the comms-list desktop table (`web/themes/default/page_comms.tpl`).
 - Removing `<meta name="format-detection" content="telephone=no…">`
   from `core/header.tpl` (or the defensive `.drawer a[href^="tel:"]`
   reset in `theme.css`) → mobile Safari + some Android Chromes
@@ -542,6 +551,7 @@ audit (#1207) locked in. New CTAs:
 | Render a page                          | `web/pages/<page>.php` + `web/includes/View/*View.php`   |
 | Edit a template                        | `web/themes/default/*.tpl`                               |
 | Reuse the moderation-queue card layout (admin submissions / protests, mobile-stacked summary rows) | `web/themes/default/css/theme.css` (`.queue-row`, `.queue-row__body`, `.queue-row__date` — #1207 PUB-2). Apply by adding `class="queue-row …"` to the outer `<details>` and dropping the inline `flex` / `flex-shrink:0` styles from the summary children. |
+| Add visible row actions to a table-rendered admin list (Edit / Unmute / Remove buttons + responsive mobile-card mirror) | `web/themes/default/page_comms.tpl` (#1207 ADM-5) is the canonical reference: `<button class="btn btn--secondary btn--sm">` / `<a class="btn btn--ghost btn--sm">` inside a `.row-actions` cell, plus `.ban-card__actions` row of identical-data-action buttons in the mobile card. Wire destructive / state-changing buttons via `data-action="…"` + `data-bid` + `data-fallback-href`; the inline page-tail JS calls `sb.api.call(Actions.PascalName)` and falls back to the GET URL if the JSON dispatcher is absent. |
 | Edit the player-detail drawer (open trigger, tabs, panes, lazy loaders) | `web/themes/default/js/theme.js` (`renderDrawerBody` / `loadPaneIfNeeded`) |
 | Add admin-only per-player notes | `web/api/handlers/notes.php` (CRUD) — Notes tab is gated by `bans.detail`'s `notes_visible` flag |
 | Render admin-authored Markdown to safe HTML | `web/includes/Markup/IntroRenderer.php` (`Sbpp\Markup`) |
--- a/web/api/handlers/_register.php
+++ b/web/api/handlers/_register.php
@@ -86,6 +86,11 @@ Api::register('comms.add',                    'api_comms_add',                  
 Api::register('comms.prepare_reblock',        'api_comms_prepare_reblock',        0, true);
 Api::register('comms.paste',                  'api_comms_paste',                  ADMIN_OWNER | ADMIN_ADD_BAN);
 Api::register('comms.prepare_block_from_ban', 'api_comms_prepare_block_from_ban', 0, true);
+// comms.unblock dispatcher gate matches the legacy GET handler: any of the
+// four "lift-a-block" flags lets the request through; the handler then
+// does the precise per-row check (own/group bans). #1207 ADM-5/ADM-6.
+Api::register('comms.unblock', 'api_comms_unblock', ADMIN_OWNER | ADMIN_UNBAN | ADMIN_UNBAN_OWN_BANS | ADMIN_UNBAN_GROUP_BANS);
+Api::register('comms.delete',  'api_comms_delete',  ADMIN_OWNER | ADMIN_DELETE_BAN);
 
 // ---- groups -----------------------------------------------------------
 Api::register('groups.add',                   'api_groups_add',                   ADMIN_OWNER | ADMIN_ADD_GROUP);
--- a/web/api/handlers/comms.php
+++ b/web/api/handlers/comms.php
@@ -107,6 +107,161 @@ function api_comms_prepare_reblock(array $params): array
     ];
 }
 
+/**
+ * Lift an active gag/mute on a comm row (#1207 ADM-5/ADM-6).
+ *
+ * Modern JSON twin of the legacy `?p=commslist&a=ungag|unmute&id=…&key=…`
+ * GET handlers in `page.commslist.php`. The legacy GET path stays put
+ * (no-JS fallback for the icon-only theme leg + third-party themes that
+ * still ship the v1.x action links), but the comms list's visible-action
+ * affordance now wires through this action so the row can update
+ * in-place + show a toast without a full page reload. Permission gate
+ * mirrors the legacy GET handler exactly:
+ *
+ *   ADMIN_OWNER | ADMIN_UNBAN     — unconditional, lift any row.
+ *   ADMIN_UNBAN_OWN_BANS          — only the row's own admin (`aid`).
+ *   ADMIN_UNBAN_GROUP_BANS        — only rows where `gid` matches the
+ *                                   caller's `gid`.
+ *
+ * The dispatcher gate is `ADMIN_OWNER | ADMIN_UNBAN |
+ * ADMIN_UNBAN_OWN_BANS | ADMIN_UNBAN_GROUP_BANS` — the broadest "any
+ * unban-ish flag" match — and the per-row precision check happens
+ * inside the handler, since the dispatcher can't see which row the
+ * caller wants to act on.
+ *
+ * Inputs:
+ *   - `bid`     (int, required) — the comm-block id.
+ *   - `ureason` (string, optional) — admin-supplied unblock reason; we
+ *     trim and store as-is. Stored raw in `ureason` (per the
+ *     "store raw, escape on display" anti-pattern); the column lives
+ *     behind the same Smarty auto-escape pipeline as `reason`.
+ *
+ * @return array{ bid: int, state: string, type: int }
+ */
+function api_comms_unblock(array $params): array
+{
+    global $userbank;
+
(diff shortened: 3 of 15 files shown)


rumblefrog:
## Screenshots (comms-affordances)

| `admin-audit` |  |  |  |  |
| `admin-bans` |  |  |  |  |
| `admin-home` |  |  |  |  |
| `banlist` |  |  |  |  |
| `commslist` |  |  |  |  |
| `home` |  |  |  |  |
| `login` |  |  |  |  |
| `myaccount` |  |  |  |  |
| `protest` |  |  |  |  |
| `servers` |  |  |  |  |
| `submit` |  |  |  |  |

rumblefrog (commented):
## Review of slice 6 of #1207 — comms list affordances (ADM-5 / ADM-6)

Walked the diff end-to-end against `AGENTS.md` (Frontend / JSON API /
CSRF / Permissions / nofilter / Anti-patterns) and the legacy GET
handlers in `web/pages/page.commslist.php` lines 62–209. Everything
material checks out; one minor test-coverage suggestion below, no
blockers / majors.

### Things I verified

**ADM-5 (visible row actions)**
- `web/themes/default/css/theme.css` drops the `.row-actions { opacity: 0 } / tr:hover { opacity: 1 }` rule and replaces it with `display: flex` only — comment cites #1207 ADM-5.
- Desktop `<td class="col-actions">` and mobile `.ban-card__actions` both render explicit text-labelled `<button>` / `<a>` (Edit / Unmute / Ungag / Lift block / Re-apply / Remove) at every viewport. The `data-testid` set is the same one the screenshot harness already grabs (`row-action-edit`, `row-action-unmute`, `row-action-delete`) plus parallel `-mobile` testids on the card siblings and a new `row-action-reapply` for ADM-6.
- Mobile card was correctly split out of the previous single card-wrapping `<a>` so the action buttons aren't nested-interactive HTML inside the summary anchor. The summary anchor's filter-by-SteamID navigation is preserved on `data-testid="comm-card-link"`.

**ADM-6 (in-place flip + toast + Re-apply)**
- The inline page-tail JS at `page_comms.tpl:498–662` calls `sb.api.call(Actions.CommsUnblock, …)`, flips `data-state` and the `pill--<state>` classes, swaps the Unmute `<button>` for a Re-apply `<a href="…&rebanid=…">`, and fires `window.SBPP.showToast({ kind: 'success', title: 'Block lifted' })`. The Re-apply target threads through the existing `comms.prepare_reblock` flow — no new "re-block" handler needed.
- `flipRowToUnmuted()` (lines 547–589) hits both the desktop `<tr>` and the mobile `<div>` mirrors via `rowsForBid()` so a viewport switch mid-test sees a consistent state.

**New JSON handlers (`comms.unblock`, `comms.delete`)**
- Registered in `web/api/handlers/_register.php:92–93` with the right gates: `comms.unblock` is `ADMIN_OWNER | ADMIN_UNBAN | ADMIN_UNBAN_OWN_BANS | ADMIN_UNBAN_GROUP_BANS` (broadest "any unban-ish flag" union), `comms.delete` is `ADMIN_OWNER | ADMIN_DELETE_BAN`. Both match the legacy GET semantics line-for-line.
- The per-row precision check inside `api_comms_unblock` (`web/api/handlers/comms.php:162–174`) mirrors `page.commslist.php:71` exactly via De Morgan: `OWNER|UNBAN` OR (`UNBAN_OWN_BANS && rowAid === GetAid()`) OR (`UNBAN_GROUP_BANS && rowGid === callerGid`). No widening / no escalation. The dispatcher's `is_logged_in()` gate also makes the JSON path slightly *stricter* than the public legacy GET path (anonymous callers can hit `?p=commslist` but can't reach the JSON action) — that's a hardening, not a regression.
- `comms.delete` correctly only fires `sc_fw_un{mute,gag}` when the row was still active (`empty(RemoveType) && (length === 0 || ends > now)`), matching `page.commslist.php:197`.
- `PermissionMatrixTest::matrix()` extended with both new actions; snapshots (`unblock_success`, `unblock_forbidden`, `delete_success`, `delete_forbidden`) present and shape-match the handler return values. `web/scripts/api-contract.js` regenerated with `Actions.CommsUnblock` / `Actions.CommsDelete` constants.

**Anti-patterns / conventions**
- `git diff origin/main..HEAD` greps clean for `htmlspecialchars_decode` / `html_entity_decode` / `xajax` / `sb-callback` / `ShowBox(` / `MooTools` / `ADOdb`. The two `sourcebans.js` mentions in the diff are both comments explaining its prior removal.
- No new `{$foo nofilter}` (the diff is empty for `nofilter`).
- No schema or `data.sql` changes — no `web/updater/data/<N>.php` migration needed.
- `data-action`-driven click delegation goes through `sb.api.call` (which sets `X-CSRF-Token` from the meta tag automatically per `web/scripts/api.js:42`) — CSRF coverage is correct and uses `Actions.PascalName`, never raw integers / string literals.
- `ureason` is stored raw with `trim()` (per AGENTS.md "store raw, escape on display" — slight divergence from the legacy GET path which still `htmlspecialchars()`s it, but that's the *legacy* path being wrong; the new JSON path is following project policy).

**E2E spec quality (`flows/comms-affordances.spec.ts`)**
- Anchored on `data-testid` / `data-state` / `data-kind` per #1123. No CSS class chains as primary selectors. `hasText` filters only used for toast disambiguation.
- No `setTimeout` / `waitForTimeout` — every assertion waits on Playwright's auto-retrying `expect(...)`. Reads `getComputedStyle(el).opacity === '1'` *without* hovering as the deterministic regression guard for the hover-only treatment, which is exactly the right shape.
- `test.describe.configure({ mode: 'serial' })` with the right rationale comment about the truncate-and-reseed window.
- Existing `flow-comms-gag-mute.spec.ts` + `_screenshots.spec.ts` correctly re-aimed at `data-fallback-href` since the legacy GET URL moved off the `<a>`'s `href`.

**CI** — All five gates green on the latest commit (PHPStan, PHPUnit, ts-check, API contract, Playwright + axe). No need to re-run locally.

**Docs sync** — `AGENTS.md` Anti-patterns section grows the hover-only-row-actions entry, and the "Where to find what" table gets a "Add visible row actions to a table-rendered admin list" row pointing at this PR's template as the canonical reference. `ARCHITECTURE.md`'s "Handler registration" section doesn't list individual actions (only topic files), so no update needed there since `api/handlers/comms.php` already existed.

### Findings

**Finding 1 — Minor — No regression test exercising the per-row OWN/GROUP precision check on `comms.unblock`.**
- Location: `web/tests/api/CommsTest.php` (lines 166–238 cover `comms.unblock`).
- What: The existing tests (`testUnblockRejectsAnonymous`, `testUnblockBadRequestOnMissingBid`, `testUnblockNotFoundForUnknownBid`, `testUnblockLiftsActiveGagAndPersistsState`, `testUnblockRejectsAlreadyLiftedRow`) all run as `loginAsAdmin()`, which has `ADMIN_OWNER` and short-circuits the per-row precision check via the first OR-branch (`HasAccess(ADMIN_OWNER | ADMIN_UNBAN)`). The "load-bearing security gate" the handler comment explicitly calls out (`web/api/handlers/comms.php:162–174`) — that a caller with only `ADMIN_UNBAN_OWN_BANS` can unblock their own rows but not someone else's, and the same for `ADMIN_UNBAN_GROUP_BANS` — has zero direct test coverage. The legacy GET path at `page.commslist.php:71` also doesn't have a test, so this isn't a regression introduced *by this PR*, but the JSON twin makes the surface bigger and the absence of a test makes a future refactor that drops the per-row branch silently undetectable.
- Why minor (not major / blocker): The handler logic itself is correct and matches the legacy GET path verbatim under De Morgan. The dispatcher gate fails closed for users with no unban flag at all, and the snapshot tests confirm the forbidden envelope shape. So the security boundary is sound at HEAD; this is a "lock the contract" follow-up.
- Suggested fix (out-of-scope for this PR; happy to merge as-is): a follow-up that exposes a `loginAsAdminWith($flags, $gid)` helper in `ApiTestCase` and adds three cases — `OWN_BANS hits own row → success`, `OWN_BANS hits other admin's row → forbidden`, `GROUP_BANS hits same-group row → success`, `GROUP_BANS hits cross-group row → forbidden`. Same shape would be useful for the mirroring `bans.*` handlers too.

### Things you did well

- The dispatcher gate decision (broadest "any unban flag" union, with the precision check inside the handler) is the correct shape, and the docblock above the handler explains *why* the dispatcher can't do the precision check (it doesn't know which row the caller wants to act on). That makes future maintainers far less likely to "tighten" the gate and accidentally lock out OWN_BANS / GROUP_BANS callers.
- `data-fallback-href` for the legacy GET URL + a JS path that follows it when `window.sb.api` / `window.Actions` aren't defined is a clean degradation contract — third-party themes that strip `api.js` continue working without a JS-error noise floor, and the existing `flow-comms-gag-mute` screenshot harness pivoted to `data-fallback-href` cleanly.
- Splitting the mobile card from a single wrapping `<a>` into a `.ban-card__summary` anchor + sibling `.ban-card__actions` row is the right HTML — nested-interactive-content is a real a11y / serialisation hazard that the previous shape was carrying. The `comm-card-link` testid on the inner anchor lets the spec independently assert the navigate-to-search affordance still works.
- `flipRowToUnmuted` hitting both the desktop and mobile mirrors via `rowsForBid()` (with the comment explaining why — `display:none` keeps both in DOM) catches a class of "switch viewport mid-spec, see stale state" bug that's easy to write into existence.
- AGENTS.md additions are prescriptive and include a concrete reference to `page_comms.tpl` as the canonical pattern. That's exactly the right shape for a "introduce a new pattern" change per the "Keep the docs in sync" table.
- The PHPDoc on `api_comms_unblock` quotes the permission-flag matrix verbatim, which means the api-contract.js regen carries it to JS callers as a `@typedef`/JSDoc — anyone reading the JS contract sees the same security note.

Issue #1184, opened 2026-05-05
[#1124 followup] ⌘K palette hint renders as tofu on non-Mac platforms (no font has U+2318)


rumblefrog:
User-flagged during the #1124 screenshot gallery review:

> The Command K symbol looks weird.

The topbar's "Search players, SteamIDs…" trigger ends in a small `<kbd>` hint that the issue's intent was to show as `⌘K`. On Linux/Windows Chromium (and the Playwright default test runners), the `⌘` (U+2318 PLACE OF INTEREST SIGN) character has no glyph in any of the available fonts and renders as a placeholder box (tofu).

## Repro

1. Load any panel page in Linux or Windows Chromium.
2. Look at the small `<kbd>` widget on the right of the topbar search trigger — instead of `⌘K` you see something like `□K` or `⩁K`.

In the gallery, every desktop screenshot shows it:

![home light topbar](https://raw.githubusercontent.com/sbpp/sourcebans-pp/screenshots-archive/screenshots/issue-1124-final/light/desktop/home.png)

## Cause

`web/themes/default/core/title.tpl` line 51:

\`\`\`html
<kbd>&#8984;K</kbd>
\`\`\`

`&#8984;` is U+2318 ⌘. The vendored `JetBrains Mono` (`web/themes/default/fonts/JetBrainsMono-*`) does not include this glyph, and the CSS fallback `'JetBrains Mono', ui-monospace, SFMono-Regular, Menlo, monospace` only finds a font with the glyph on macOS.

The shortcut handler in `web/themes/default/js/theme.js` line 112 already supports both: `(e.metaKey || e.ctrlKey)`, so the actual binding works on every platform — only the visual hint is platform-confused.

## Fix sketch

Detect platform once at boot and rewrite the kbd label:

\`\`\`js
const isMac = navigator.platform.toUpperCase().includes('MAC') ||
              /Mac|iPhone|iPad/.test(navigator.userAgent);
document.querySelectorAll('.topbar__search kbd').forEach((el) => {
  el.textContent = isMac ? '⌘K' : 'Ctrl K';
});
\`\`\`

Or render a static `Ctrl K` and accept that Mac users see a Windows-flavoured hint; the binding still works.

Other places may carry similar `⌘`/`⌥`/`⌃` hints — grep `&#8984;` / `&#8997;` / `&#8963;` / `⌘` / `⌥` / `⌃` to find them.

## Files involved

- `web/themes/default/core/title.tpl` (line 51)
[...]

PR #1217, opened 2026-05-06
fix(theme): topbar palette desktop polish (#1207 slice 9)

### LINKED ISSUE #1207: UI/UX audit: cross-cutting findings across public + admin flows on origin/main

rumblefrog:
## Summary

Slice 9 of #1207 — topbar/palette desktop polish. Addresses CC-3, DET-2.

- **CC-3** — extend slice 1 (#1208) mobile icon-only collapse of `.topbar__search` to desktop too. The labelled "search input + Ctrl K hint" was a duplicate affordance for the same `<dialog id="palette-root">` the ⌘K shortcut already opens; both surfaces competed for attention. The trigger is now a ghost icon button at every viewport (matching the sibling theme-toggle's chrome). The `.topbar__search-label` / `.topbar__search-kbd` spans stay in the DOM so SR users still hear the existing `aria-label` and `applyPlatformHints` can still rewrite the kbd text on Mac without re-rendering.
- **DET-2** — each player result row in the palette grows a kbd hint group at the right edge surfacing two interactions: bare `Enter` opens the player drawer for the ban (rows now carry `data-drawer-bid` so theme.js's existing drawer click delegate handles the path; the palette closes itself first so the drawer isn't stacked behind it), `Ctrl/Cmd+Enter` copies the row's SteamID via `navigator.clipboard.writeText` + a `showToast` confirmation. Glyphs are server-rendered in non-Mac form (`Enter`, `Ctrl`); `applyPlatformHints` picks up a new `[data-enterkey]` attribute and swaps to ⏎ alongside the existing `[data-modkey]` → ⌘ swap (#1184).

Docs paired in commit 2:

- `AGENTS.md` "Anti-patterns" — entry warning future agents not to re-add the labelled topbar search.
- `AGENTS.md` "Where to find what" — row pointing at `theme.js`'s palette functions, the `.topbar__search` shape in `core/title.tpl`, and the `.palette__row*` rules in `theme.css`.
- `ARCHITECTURE.md` "Smarty templates + View DTOs" — descriptive blurb on the palette being the only search affordance in the chrome and the topbar trigger being icon-only at every viewport, plus the per-row data-attribute / glyph-swap contract.

## Test plan

- [x] PHPStan: no PHP changed in this slice; the only failures on this branch (`Sbpp\View\YourAccountView::$web_permissions_grouped`) reproduce on `origin/main` and belong to slice 8 territory.
- [x] PHPUnit clean (315 tests, 1291 assertions; nothing under `web/tests/Unit` or `web/tests/api` changed).
- [x] ts-check clean (`./sbpp.sh ts-check`).
- [x] API contract clean (no handler / perm / `@param` / `@return` touched; `composer api-contract` re-emits the same `scripts/api-contract.js`).
- [x] Playwright E2E clean — `./sbpp.sh e2e --workers=1` (mirroring CI's `workers: 1`): 141 passed, 193 skipped (project filtering), 0 failed.
  - New coverage in `command-palette.spec.ts`:
    - `topbar palette trigger renders icon-only at every viewport (#1207 CC-3)` — chromium + mobile-chromium, locks the icon-only shape (square button, no visible label / kbd, `aria-label` intact).
    - `clicking the icon trigger opens the palette (#1207 CC-3)` — chromium + mobile-chromium.
    - `player result row carries kbd hint group with Enter and Ctrl+Enter (#1207 DET-2)` — chromium-only (mobile skips on the Linux runner per platform-glyph default).
    - `Ctrl+Enter on a focused player row copies the SteamID + toasts (#1207 DET-2)` — chromium-only, grants `clipboard-read`/`clipboard-write`, asserts `navigator.clipboard.readText()` matches the row's SteamID + the success toast surfaces.
    - `focused result + Enter opens the player drawer (#1207 DET-2)` — replaces the previous "Enter navigates to ban-list" assertion now that the click delegate hands off to `loadDrawer` and closes the palette.
- [x] Screenshot gallery regenerated (96 PNGs; topbar desktop look changed because the `.topbar__search` shape is now icon-only). Will run `web/tests/e2e/scripts/upload-screenshots.sh <pr> palette-desktop` once the PR exists and comment the markdown table.
- [x] Manual: desktop topbar renders the palette trigger icon-only at every width; palette opens via icon click AND `Meta+k`; player rows carry `Enter` / `Ctrl+Enter` kbd hints with `applyPlatformHints` swapping to ⏎ / ⌘ on Mac; `Ctrl+Enter` on a focused row copies SteamID and surfaces the success toast; bare `Enter` opens the drawer and dismisses the palette in one beat.

Does NOT close #1207.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -531,6 +531,16 @@ audit (#1207) locked in. New CTAs:
   iOS dark, blueish on Android). The chrome doesn't have a single
   phone number on it; the meta is the canonical opt-out and the
   CSS reset is the belt-and-suspenders for variants that ignore it.
+- Re-adding a labelled search input or a visible `Ctrl K` / `⌘K` hint
+  to the topbar palette trigger (the `.topbar__search` button in
+  `core/title.tpl`) → the labelled shape was a duplicate affordance
+  for the same `<dialog id="palette-root">` the ⌘K shortcut already
+  opens, and on mobile it broke the topbar (#1207 CC-1, slice 1) and
+  on desktop it visibly competed with the palette itself (#1207 CC-3,
+  slice 9). The trigger is now icon-only at every viewport, matching
+  the sibling theme-toggle's chrome. The `.topbar__search-label` /
+  `.topbar__search-kbd` spans stay in the DOM for SR users + the
+  Mac glyph swap, but `display: none` everywhere — don't unhide them.
 
 ## Where to find what
 
@@ -543,6 +553,7 @@ audit (#1207) locked in. New CTAs:
 | Edit a template                        | `web/themes/default/*.tpl`                               |
 | Reuse the moderation-queue card layout (admin submissions / protests, mobile-stacked summary rows) | `web/themes/default/css/theme.css` (`.queue-row`, `.queue-row__body`, `.queue-row__date` — #1207 PUB-2). Apply by adding `class="queue-row …"` to the outer `<details>` and dropping the inline `flex` / `flex-shrink:0` styles from the summary children. |
 | Edit the player-detail drawer (open trigger, tabs, panes, lazy loaders) | `web/themes/default/js/theme.js` (`renderDrawerBody` / `loadPaneIfNeeded`) |
+| Edit the command palette (icon-only trigger, ⌘K binding, result rows, kbd hints, Ctrl+Enter copy) | `web/themes/default/js/theme.js` (`openPalette` / `closePalette` / `renderPaletteResults` / `applyPlatformHints` / `handlePaletteCopyShortcut`) + `core/title.tpl` (the `.topbar__search` icon button) + the `.palette__row*` rules in `web/themes/default/css/theme.css`. Player rows carry `data-drawer-bid="<bid>"` (bare Enter / click → `loadDrawer`, palette closes itself) + `data-steamid="<steam>"` (`Ctrl/Cmd+Enter` → `navigator.clipboard.writeText` + `showToast`). The kbd glyphs are server-rendered in non-Mac form (`Enter`, `Ctrl`); `applyPlatformHints` swaps `[data-enterkey]` → ⏎ and `[data-modkey]` → ⌘ on Mac/iOS at boot and after every render (#1184, #1207 DET-2). |
 | Add admin-only per-player notes | `web/api/handlers/notes.php` (CRUD) — Notes tab is gated by `bans.detail`'s `notes_visible` flag |
 | Render admin-authored Markdown to safe HTML | `web/includes/Markup/IntroRenderer.php` (`Sbpp\Markup`) |
 | Live-preview Markdown in a settings textarea | `system.preview_intro_text` JSON action + `web/themes/default/page_admin_settings_settings.tpl` (`.dash-intro-editor` / `.dash-intro-preview`) |
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -336,6 +336,25 @@ controls, light/dark/system theming. Custom themes ship their own
 `theme.conf.php` with `theme_name` / `theme_author` / `theme_version` /
 `theme_link` / `theme_screenshot`.
 
+The command palette (`#palette-root` `<dialog>` rendered by
+`themes/default/js/theme.js`) is the only search affordance in the
+chrome. The topbar carries an icon-only ghost button
+(`.topbar__search` in `core/title.tpl`) that opens the same dialog as
+the `Meta+k` keybinding — the pre-v2.0.0 inline search input was
+dropped at #1207 CC-1 (mobile, slice 1) + CC-3 (desktop, slice 9)
+because the labelled "search input + Ctrl K hint" was a duplicate
+affordance for the same dialog. Player result rows in the palette
+carry `data-drawer-bid` (bare Enter / click hands off to the existing
+`[data-drawer-bid]` click delegate, which closes the palette and
+opens the player drawer) and `data-steamid` (the `Ctrl/Cmd+Enter`
+handler in `theme.js`'s `handlePaletteCopyShortcut` reads it and
+copies via `navigator.clipboard.writeText` + `showToast`). Keyboard
+glyphs in the row's `.palette__row-hints` group are server-rendered
+in non-Mac form (`Enter`, `Ctrl`); `applyPlatformHints` rewrites
+`[data-enterkey]` → ⏎ and `[data-modkey]` → ⌘ on Mac/iOS clients at
+boot and after every render so glyph swaps don't require re-fetching
+results (#1184, #1207 DET-2).
+
 The preferred way to render is via typed view-model DTOs:
 
 ```php
--- a/web/tests/e2e/specs/flows/ui/command-palette.spec.ts
+++ b/web/tests/e2e/specs/flows/ui/command-palette.spec.ts
@@ -1,5 +1,5 @@
 /**
- * Command palette (#1124, Slice 6).
+ * Command palette (#1124, Slice 6 + #1207 CC-3, DET-2).
  *
  * Acceptance criteria from #1124:
  *   "⌘K opens, typing a SteamID surfaces the player, Enter navigates
@@ -15,9 +15,22 @@
  *     [data-result-kind="nav"|"ban"]>` once `q.length >= 2`, with a
  *     200ms input debounce.
  *
- * The third subtest (Enter navigates) follows the focused result's
- * native `<a>` activation rather than a custom keydown handler —
- * `theme.js` doesn't bind Enter on the palette input itself.
+ * #1207 CC-3 (this slice) extends slice 1's mobile icon-only collapse
+ * to desktop too: the topbar's palette trigger is icon-only at every
+ * viewport, so the topbar reads as `[hamburger] [breadcrumb] [spacer]
+ * [palette icon] [theme toggle]` consistently — the palette dialog
+ * itself owns the search affordance.
+ *
+ * #1207 DET-2 (this slice) layers two interactions onto each player
+ * result row, advertised by a kbd hint group at the right edge:
+ *   - bare Enter → opens the player drawer for the ban,
+ *   - Ctrl/Cmd+Enter → copies the row's SteamID via
+ *     navigator.clipboard.writeText + surfaces a toast.
+ * The kbds are server-rendered in non-Mac form ("Enter", "Ctrl");
+ * theme.js's applyPlatformHints swaps `[data-enterkey]` → ⏎ and
+ * `[data-modkey]` → ⌘ on Mac at boot and on every render. The Linux
+ * test runner sees the non-Mac form by default — that's what the
+ * assertions below pin.
  *
  * == Seeding strategy: unique-per-(test × project), no truncate ==
  *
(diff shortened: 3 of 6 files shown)


rumblefrog:
## Screenshots (palette-desktop)

| `admin-audit` |  |  |  |  |
| `admin-bans` |  |  |  |  |
| `admin-home` |  |  |  |  |
| `banlist` |  |  |  |  |
| `commslist` |  |  |  |  |
| `home` |  |  |  |  |
| `login` |  |  |  |  |
| `myaccount` |  |  |  |  |
| `protest` |  |  |  |  |
| `servers` |  |  |  |  |
| `submit` |  |  |  |  |

rumblefrog (commented):
## Review of slice 9 of #1207 — topbar palette desktop polish (CC-3, DET-2)

CI is green across all gates, the deviation from CC-3/DET-2's stated suggested fix (Enter→drawer instead of Enter→banlist) is justified and well-tested, and the issue's exact fix shape is delivered. Findings below are minor — none block merge. Reviewer 9 will push fixes for findings 1+2 in a follow-up commit on this branch.

### Findings

**Finding 1 — major (docs accuracy)**
- Location: `web/themes/default/js/theme.js:337-339`, `web/tests/e2e/specs/flows/ui/command-palette.spec.ts:425-428`, commit message of `5540ea02`
- What: comments claim "middle-click / Cmd+click still navigates to a name-filtered banlist" via the preserved `href`. Only **middle-click** actually does — Cmd+click fires a regular `click` event with `metaKey: true`, the existing drawer click delegate at `theme.js:919` doesn't guard modifiers and unconditionally calls `e.preventDefault()`, suppressing the browser's "open in new tab" default action. So Cmd+click on a palette row opens the drawer in the current tab, not a name-filtered banlist in a new tab.
- Why: middle-click works because browsers fire `auxclick` (not `click`) for non-primary buttons, so the JS delegate doesn't intercept. Cmd/Ctrl/Shift+left-click fires `click` with the modifier flag set, which the delegate intercepts indistinguishably from a bare left-click.
- Suggested fix: add `if (e.metaKey || e.ctrlKey || e.shiftKey || e.button !== 0) return;` at the top of the `if (trigger)` branch in the drawer click delegate (theme.js:920ish). This delivers the documented behaviour for **all** `[data-drawer-bid]` / `[data-drawer-href]` triggers — palette rows fall back to the name-filtered banlist URL on Cmd+click; banlist rows fall back to `?p=banlist&id=N` (a generic banlist URL — slightly less useful than current behaviour, but consistent with native anchor semantics that the user opted into via the modifier). Add a chromium-only spec asserting `dialog[data-palette-open]` stays open and `#drawer-root[data-drawer-open]` stays `false` after a `Control+click` on a focused result row.

**Finding 2 — minor (UX inconsistency for IP-only bans)**
- Location: `web/themes/default/js/theme.js:344` (`const steamid = b.steam || b.ip || ''`) and `205-208` (toast title `'SteamID copied'`)
- What: when an IP-based ban (`type === 1`, empty `authid`) shows up in palette results, the row's `data-steamid` actually holds the IP, the kbd hint says "Ctrl+Enter to copy steamid", and the toast title says "SteamID copied" with the IP as the body. All three labels are misleading.
- Why: `bans.search` returns `steam` (= `authid`) and `ip` separately; for IP-only bans `authid` is `''` and the fallback chain falls through to the IP. The current code treats the SteamID and IP as interchangeable copy targets but advertises them as "SteamID".
- Suggested fix: smallest patch — make the toast title conditional in `handlePaletteCopyShortcut`, e.g. read `row.dataset.steam` (a separate explicit attribute) and show `'SteamID copied'` vs `'IP copied'`. Or downgrade the title to the existing generic `'Copied to clipboard'` (matches the `[data-copy]` handler at `theme.js:1058-1066`) so the title is true regardless of the underlying ban shape. The kbd hint label is a copy edit but can stay as "copy steamid" if we treat that label as referring to the affordance (IP-only bans are rare in palette results in practice).

**Finding 3 — nit (redundant attribute)**
- Location: `web/themes/default/js/theme.js:349-350`
- What: every player result row carries both `data-id="<bid>"` and `data-drawer-bid="<bid>"` with the same value. The `data-id` predates this slice; `data-drawer-bid` is the new attribute the click delegate consumes. The existing `data-id` isn't read anywhere on palette rows — search-confirmed (other tables use `[data-testid="ban-row"][data-id="…"]` but palette rows don't have a `ban-row` testid).
- Suggested fix: drop `data-id` from the palette row's attribute list — `data-drawer-bid` is the single source of truth. Or document the intentional duplication if there's a future plan for `data-id` on palette rows.

**Finding 4 — nit (a11y improvement)**
- Location: `web/themes/default/js/theme.js:358` (`<div class="palette__row-hints" data-testid="palette-row-hints" aria-hidden="true">`)
- What: the kbd hint group is `aria-hidden="true"`, so SR users don't hear "Enter to open drawer / Ctrl+Enter to copy steamid". The aria-hidden is the right call for the decorative hint glyphs, but the row itself doesn't otherwise advertise the keyboard interactions.
- Suggested fix (follow-up): add `aria-keyshortcuts="Enter Control+Enter"` to the `<a>` so AT users hear the row's keyboard contract via the standard ARIA shortcut hint. Out of scope for this slice if the existing `[data-drawer-bid]` triggers in the codebase don't carry it either; tracker for the wider "advertise drawer-trigger keyboard contract to AT" rework.

### Things you did well

- **Comprehensive E2E coverage**: viewport-locked icon-only shape (both projects, with a square-button + bounding-box assertion that catches cascade drift), click-opens-palette behavioural counterpart (both projects), kbd hint group contract (chromium), Ctrl+Enter copy + toast + clipboard `readText` (chromium with explicit `clipboard-read`/`clipboard-write` permission grant on the live origin), Enter→drawer with palette-close mirror.
- **`context.grantPermissions` placement**: granted AFTER `page.goto('/')` so the permission applies to the live test origin, not `about:blank`. Loud at the call site, portable across Chromium versions.
- **`applyPlatformHints` re-run after every `renderPaletteResults`**: dynamically-injected kbds inside `.palette__row-hints` get the platform glyph swap on every render, no re-fetching needed. The `[data-enterkey]` attribute extension is a clean reuse of the #1184 `[data-modkey]` mechanism.
- **`closePalette` scope on the drawer click delegate**: `target.closest('.palette')` confines the new behaviour to in-palette drawer triggers — banlist rows, history list items, and other `[data-drawer-bid]` callers keep their existing behaviour. Worker comment makes the scoping intent explicit.
- **Behaviour deviation justification**: Enter→drawer matches #1124's original acceptance criteria ("Enter navigates to their drawer"); the previous Enter→banlist was a Slice 6 implementation artifact. Spec-rewrite is in-place (the previous "Enter navigates" assertion is replaced, not duplicated). No other E2E spec or page-tail JS depended on the old behaviour — `_screenshots.spec.ts:616` only asserts the result row is visible, doesn't navigate; `responsive/drawer.spec.ts` and `flows/ui/player-drawer.spec.ts` are scoped to non-palette drawer triggers.
- **Handler scope guards** in `handlePaletteCopyShortcut`: `!palette || palette.hidden` short-circuit, `closest('[data-result-kind="ban"]')` (not `nav`), `if (!steamid) return`. The chord won't hijack Ctrl+Enter on a notes textarea or edit-ban form because the `data-result-kind="ban"` ancestor lookup only succeeds inside an open palette result row.
- **Docs paired in commit 2**: AGENTS.md anti-patterns + Where-to-find-what + ARCHITECTURE.md narrative all land with the code change per "Keep the docs in sync". The new "anti-pattern" entry will save a future agent from re-introducing the labelled topbar.
- **Selector hygiene**: every E2E selector is `data-testid` / `data-palette-open` / `data-loading` (terminal-attribute waits, no `setTimeout`/`waitForTimeout`). `hasText` is only used as a `.filter()` disambiguator alongside the testid primary, per #1123 "Testability hooks" rule.
- **Function declaration of `handlePaletteCopyShortcut`** (hoisted) lets it be called from the keydown handler defined earlier in the IIFE without extra plumbing — a nice match for the file's existing hoist-vs-position pattern (`applyPlatformHints` does the same dance for the `renderPaletteResults` callsite).

rumblefrog:
Addressed review findings 1 + 2 in commit `6258dbb7`:

- **Finding 1 (modifier-click graceful-degradation)**: added `if (e.metaKey || e.ctrlKey || e.shiftKey || e.button !== 0) return;` at the top of the `[data-drawer-bid] / [data-drawer-href]` click delegate's `if (trigger)` branch in `web/themes/default/js/theme.js`. Cmd/Ctrl/Shift+left-click now bails before `e.preventDefault()`, so the browser's native "open in new tab/window" takes over and the row's preserved `href` (`?p=banlist&advType=name&advSearch=<name>` for palette rows) lands in the new tab. Locked with a chromium-only spec that captures the new page via `context.waitForEvent('page')`, waits on `waitForURL(/[?&]p=banlist/)` (handles the `about:blank` → navigation-target race), asserts the URL shape, and verifies the original tab's drawer stays closed and the palette stays open.

- **Finding 2 (toast title for IP-only bans)**: in `handlePaletteCopyShortcut`, detect Steam2/Steam3 shape via `/^(STEAM_|\[U:)/` and title the toast as either "SteamID copied" (default) or "IP copied" (IP fallback) so the surface is honest for IP-only bans where `b.steam === ''` and the renderer falls back to `b.ip`. Existing seed (STEAM_0:1:N) still exercises the SteamID branch.

Findings 3 (`data-id` redundancy) and 4 (`aria-keyshortcuts`) are nits — leaving as follow-ups, since they touch the wider drawer-trigger pattern (`data-id` is read by other tables; `aria-keyshortcuts` should land on every drawer trigger, not just palette rows).

CI re-running. Local gates green:

- PHPStan OK (216 files)
- PHPUnit 315 tests, 1291 assertions
- API contract no-diff
- ts-check OK
- E2E `--workers=1` 142 passed, 194 skipped (project filtering), 0 failed

PR #1219, opened 2026-05-06
fix(admin/admins): density rework — sticky ToC + single-submit search (#1207 slice 7)

### LINKED ISSUE #1207: UI/UX audit: cross-cutting findings across public + admin flows on origin/main

rumblefrog:
## Summary

Slice 7 of #1207 — admins page density rework. Addresses ADM-3 (admin-admins only — admin-bans and myaccount are out of scope), ADM-4.

- **ADM-3** — Wrap admin/admins in a sticky page-level table of contents: anchor sidebar at `>=1024px`, accordion (`<details open>`) at `<1024px`. Each of the five surfaces (Search, Admins, Add admin, Overrides, Add override) becomes an anchored `<section id="…" class="admin-admins-section" data-testid="admin-admins-section-…">` with `scroll-margin-top: 4rem` so jumps clear the sticky topbar (3.5rem).
- **ADM-4** — Collapse the eight per-row `[Search]` buttons in `box_admin_admins_search.tpl` to a single submit + a "Clear filters" reset link. Server-side, `admin.admins.php` ANDs every populated filter; pagination uses `http_build_query($activeFilters)` to round-trip filters across page jumps. Legacy `?advType=name&advSearch=foo` URLs still narrow correctly via a backward-compat shim that translates the legacy shape into the modern field names on entry — bookmarks and external links keep working.

Locked with a new PHPUnit integration test (renders the page handler with a real Smarty bootstrap and asserts the row count + content for the AND matrix and the legacy compat shim) and a new Playwright flow spec (desktop sidebar visibility + click-scrolls-to-section, mobile accordion, single-submit network count, URL pre-fill round-trip, Clear-filters reset).

Convention captured in `AGENTS.md` so the next dense admin page (admin-bans, myaccount) can reuse the ToC pattern verbatim. The "Where to find what" table grows two rows pointing at the partial + the four-file admin-admins search filter path.

## Test plan

- [x] PHPStan clean (`./sbpp.sh phpstan` — 216 files, 0 errors)
- [x] PHPUnit clean (`./sbpp.sh test` — 323 tests, 1312 assertions, includes the new 8-case `AdminAdminsSearchTest`)
- [x] API contract clean (`./sbpp.sh composer api-contract` — no diff)
- [x] Playwright E2E clean (`./sbpp.sh e2e --workers=1` — 141 passed, 0 flakes; the new flow spec contributes 6 chromium + 1 mobile-chromium tests)
- [x] Screenshot gallery regenerated (admin-admins light/dark on chromium + mobile-chromium)
- [x] Manual: anchor sidebar visible at >= 1024px on admin-admins; accordion ToC at < 1024px; advanced-search single submit AND-combines filters; legacy `advType=…&advSearch=…` URLs still narrow

Does NOT close #1207.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -428,6 +428,50 @@ bearing assertion that the value is already safe HTML, so:
   Reuse this pattern — never call a third-party Markdown renderer
   client-side, as it would diverge from the safe-on-render contract.
 
+### Page-level table of contents (dense admin pages)
+
+Some admin routes (`admin-admins`, the audit pegged `admin-bans` and
+`myaccount` for the same shape) cram several semi-related surfaces
+onto one long scroll. The pattern locked in by #1207 ADM-3 is a
+sticky page-level ToC: anchor sidebar at `>=1024px`, accordion
+(`<details open>`) at `<1024px`, with each section anchored via a
+unique `id="…"` + `scroll-margin-top: 4rem` so jumps clear the
+sticky topbar (3.5rem).
+
+`admin-admins` is the reference shape (see `admin.admins.toc.tpl`,
+the `.admin-admins-shell` / `.admin-admins-toc` / `.admin-admins-section`
+rules in `theme.css`, and the `<section id="…" class="admin-admins-section"
+data-testid="admin-admins-section-…">` wrapping in
+`page_admin_admins_list.tpl` / `page_admin_admins_add.tpl` /
+`page_admin_overrides.tpl`):
+
+- The ToC partial is included **once**, from the first template
+  rendered for the route (here `page_admin_admins_list.tpl`). The
+  outer `.admin-admins-shell` wrapper opens there and closes at the
+  bottom of the **last** template (`page_admin_overrides.tpl`).
+  Document the tag pairing with a Smarty comment at each end so
+  edits don't silently break the layout.
+- Each section gets `<section id="…" class="<page>-section"
+  data-testid="<page>-section-…" aria-labelledby="…-heading">` plus
+  an `<h2 id="…-heading">` so screen-readers can navigate by
+  landmark name. Reuse the `*-section` class so all sections share
+  the `scroll-margin-top` rule.
+- Permission gates on ToC entries mirror what the dispatcher would
+  render anyway (e.g. `$can_add_admins` in `admin-admins`); a ToC
+  entry for a section the dispatcher wouldn't paint is a dead link.
+- Selectors for E2E specs are `[data-testid="<page>-toc"]` (outer
+  aside), `[data-testid="<page>-toc-link-<slug>"]` (per anchor),
+  and `[data-testid="<page>-section-<slug>"]` (per section). Never
+  CSS class chains; never visible-text-as-primary. The desktop
+  contract is "scrolls to the right section" — assert via
+  `expect(section).toBeInViewport()` after the click, not against
+  a fixed `boundingBox.y` because the **last** section can never
+  reach the top of the viewport (no document below it).
+
+When the next dense page (admin-bans, myaccount) adopts this
+pattern, lift the ToC partial into a parameterized shared template
+rather than duplicating `admin.admins.toc.tpl`.
+
 ### Empty states (`first-run` vs `filtered`)
 
 Empty surfaces follow one of two shapes; pick by whether the empty
@@ -547,6 +591,8 @@ audit (#1207) locked in. New CTAs:
 | Render admin-authored Markdown to safe HTML | `web/includes/Markup/IntroRenderer.php` (`Sbpp\Markup`) |
 | Live-preview Markdown in a settings textarea | `system.preview_intro_text` JSON action + `web/themes/default/page_admin_settings_settings.tpl` (`.dash-intro-editor` / `.dash-intro-preview`) |
 | Build an empty-state surface (first-run vs filtered, primary/secondary CTAs) | `.empty-state` rules in `web/themes/default/css/theme.css` + reference shapes in `page_servers.tpl`, `page_dashboard.tpl`, `page_bans.tpl`, `page_comms.tpl`, `page_admin_audit.tpl`, `page_admin_bans_protests.tpl`, `page_admin_bans_submissions.tpl` |
+| Add a sticky page-level ToC to a dense admin route (anchor sidebar at desktop, accordion at mobile) | `web/themes/default/admin.admins.toc.tpl` is the reference partial; `.admin-admins-shell` / `.admin-admins-toc` / `.admin-admins-section` rules in `web/themes/default/css/theme.css` carry the responsive layout. See the Conventions section "Page-level table of contents (dense admin pages)" (#1207 ADM-3). |
+| Add or rename an admin-admins advanced-search filter | `web/pages/admin.admins.php` (filter-building loop + active-filter map for pagination) + `web/pages/admin.admins.search.php` (DTO population) + `web/includes/View/AdminAdminsSearchView.php` (`active_filter_*` properties) + `web/themes/default/box_admin_admins_search.tpl` (input + pre-fill). The form is single-submit AND-semantics with a backward-compat shim for legacy `advType=…&advSearch=…` URLs (#1207 ADM-4); cover new filters in `web/tests/integration/AdminAdminsSearchTest.php`. |
 | Add a shared "1 of these required" badge for an either/or input pair | `web/themes/default/page_submitban.tpl` (`data-required-group="…"` + the inline guard script — vanilla JS `// @ts-check`, blocks submit when both are empty) |
 | Bootstrap (paths, autoload, theme)     | `web/init.php`                                           |
 | Routing (`?p=…&c=…&o=…`)               | `web/includes/page-builder.php` — unrecognised admin `c=…` returns the 404 page slot via `web/pages/page.404.php` + `Sbpp\View\NotFoundView` (#1207 ADM-1) |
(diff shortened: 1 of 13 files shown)


rumblefrog:
## Screenshots (admins-density)

rumblefrog (commented):
## Review of slice 7 of #1207 (ADM-3 + ADM-4)

(Self-review — GitHub blocks `--request-changes` on own PRs; treating these as findings to address before merging.)

Solid execution overall. Single-submit + AND-semantics is correctly wired; backward-compat shim narrows-only (never broadens for legitimate users); the new admemail gate closes a URL-forgery bypass that legacy code shipped with. ToC + sticky-sidebar shape matches the slicing brief, and the AGENTS.md "page-level ToC" convention is well-written so the next dense page (admin-bans, myaccount) can copy verbatim.

Two findings to address before merge.

### Findings (must address)

**Finding 1 — Minor — Stale `nofilter` safety annotation**
- Location: `web/themes/default/page_admin_admins_list.tpl:75`
- What: The annotation reads `advSearch/advType (the only $_GET inputs) are htmlspecialchars(addslashes(...))'d before interpolation in admin.admins.php — same escape pipeline as the legacy theme.` That assertion is no longer true after this PR — `admin.admins.php` switched the pagination to `htmlspecialchars(http_build_query($activeFilters))` for the `<select onchange>` base URL (line 360) and `addslashes` is gone (`rg addslashes web/pages/admin.admins.php` returns nothing). The "only `$_GET` inputs" claim is also stale: `name`, `steamid`, `webgroup`, `srvadmgroup`, `srvgroup`, `admwebflag[]`, `admsrvflag[]`, `server` are all primary inputs now.
- Why: Per AGENTS.md "nofilter discipline" — every annotation is a load-bearing assertion that the value is safe HTML. A comment that describes the wrong escape pipeline is worse than no comment because future readers (and security reviewers) will trust the wrong story. The output IS in fact safe (`http_build_query` percent-encodes filter values; `htmlspecialchars` runs on the select's base URL; `CreateLinkR` interpolates into single-quoted `href` and percent-encoding keeps single quotes out), but the comment must reflect the current chain.
- Suggested fix: Update line 75 to describe the actual escape pipeline, e.g.:
```smarty
{* nofilter: server-built pagination HTML. Filter values flow through `http_build_query($activeFilters)` (percent-encoded, single-quote-safe) into prev/next `<a href='…'>`, and the page-jump `<select onchange>` additionally `htmlspecialchars()`-escapes the base URL before interpolation. Loop counters are integers. No raw user input reaches the rendered string. *}
```

**Finding 2 — Minor — ToC dead-link risk for users with `ADD_ADMINS` but not `LIST_ADMINS`**
- Location: `web/themes/default/admin.admins.toc.tpl:44-45`
- What: The `Search` and `Admins` ToC links render unconditionally, but `<section id="search">` and `<section id="admins">` only render inside the `{else}` arm of `{if !$can_list_admins}` in `page_admin_admins_list.tpl:45-66`. A user with `ADMIN_ADD_ADMINS` but not `ADMIN_LIST_ADMINS` (rare but valid — the perms are independent) sees the ToC's Search/Admins entries linking to non-existent anchors. The `Add admin` / `Overrides` / `Add override` entries already gate on `$can_add_admins`; this is the symmetric gate the dispatcher uses.
- Why: The slicing brief explicitly calls this out: *"The ToC entries are gated on the same permissions the dispatcher uses (no dead links)."* The toc.tpl's own header comment also says *"entries for sections the dispatcher would elide become dead links otherwise"* — but only honours that for the gated entries, not Search/Admins.
- Suggested fix: Wrap the first two `<li>` entries in `{if $can_list_admins}` (the variable is already in scope from `AdminAdminsListView`, since `admin.admins.toc.tpl` is included from inside `page_admin_admins_list.tpl`):
```smarty
{if $can_list_admins}
    <li><a href="#search" class="admin-admins-toc__link" data-testid="admin-admins-toc-link-search">Search</a></li>
    <li><a href="#admins" class="admin-admins-toc__link" data-testid="admin-admins-toc-link-admins">Admins</a></li>
{/if}
```
Add a regression case in `admin-admins-density.spec.ts` (or an integration test) that asserts the ToC links match the rendered sections — easiest shape is "for each `[data-testid^='admin-admins-toc-link-']`, the corresponding `[data-testid^='admin-admins-section-']` exists in the DOM."

### Things you did well

- **Backward-compat shim is narrowing-only.** Verified the matrix: legacy `?advType=…&advSearch=…` (single filter) still produces the same result set as pre-PR for any user with `ADMIN_LIST_ADMINS`; mixed legacy+modern URLs only ever AND-narrow (never broaden). The `if (!isset($_GET[$legacyType]) || $_GET[$legacyType] === '')` guard correctly defers to the modern field when both are present.
- **`admemail` gate is a real security improvement.** Pre-PR, anyone could URL-forge `?advType=admemail&advSearch=…` to substring-match the email column even without `ADMIN_EDIT_ADMINS`. Post-PR, the filter is dropped server-side for unprivileged users; the search box also hides the input (`{if $can_editadmin}`). Comment at `admin.admins.php:122-123` documents the intent.
- **Constant-name validation.** `preg_match('/^ADMIN_[A-Z_]+$/', $candidate) && defined($candidate)` is materially safer than the legacy `constant($flag)` direct call (which would warn-and-return-`null` on bogus input → `0|0` flag string → `HasAccess(0, …)` likely true for everyone).
- **Cross-template shell discipline.** The `.admin-admins-shell` open in `page_admin_admins_list.tpl:39` and close in `page_admin_overrides.tpl:159` are paired with explicit Smarty comments at each end documenting the boundary — exactly the shape the new AGENTS.md convention prescribes.
- **CSS rationale is well-commented.** Each new rule (`.admin-admins-section`, sticky-sidebar `top: 4rem`, `<details>` desktop force-visible override) carries a comment explaining the topbar offset, the why-not-`display:none`, and the future-shared-partial follow-up.
- **Integration test boots Smarty in-process.** Cheaper than spinning a real HTTP request; it matches `init.php`'s wiring closely (`setEscapeHtml(true)`, the same plugins) so the rendered HTML mirrors prod. Locks the AND matrix end-to-end with row-content assertions.
- **Playwright spec uses `toBeInViewport()` correctly.** The comment at `admin-admins-density.spec.ts:51-56` explicitly explains why `boundingBox.y === 64` would be a false contract for the last section — and the new "page-level ToC" AGENTS.md section codifies that as the project convention so the next dense page reuses the right assertion shape.
- **Header CTA `#add-admin` matches the new section id.** Old `#Add%20new%20admin` (with spaces!) is gone; the testable hook works out of the box.
- **Docs in lockstep.** New "Page-level table of contents (dense admin pages)" Conventions section + two "Where to find what" rows land in the same PR per `AGENTS.md`'s "Keep the docs in sync" rule.

rumblefrog:
Addressed review findings in 9a5bd3f6:

- **Finding 1** — Rewrote the stale `nofilter` annotation in `web/themes/default/page_admin_admins_list.tpl:75` to describe the actual `http_build_query()` + `htmlspecialchars()` pipeline (the legacy `addslashes(...)` call is gone after ADM-4).
- **Finding 2** — Wrapped the `Search` / `Admins` ToC entries in `{if $can_list_admins}` so a user with `ADMIN_ADD_ADMINS` but not `ADMIN_LIST_ADMINS` no longer sees dead links to elided sections. Updated the toc partial's permissions docblock to reflect the matched gates.
- Added two regression cases in `web/tests/integration/AdminAdminsSearchTest.php`:
  - `testTocLinksMatchRenderedSectionsForOwner` — structural invariant: every `admin-admins-toc-link-*` testid has a matching `admin-admins-section-*`.
  - `testTocElidesSearchAndAdminsForAddOnlyAdmin` — gated dead-link contract: a fresh ADD-only admin (gid=-1, extraflags=ADMIN_ADD_ADMINS) sees only the Add admin / Overrides / Add override entries.

Local gates green:
- PHPStan: 216 files, 0 errors
- PHPUnit: 325 tests / 1318 assertions (10 in `AdminAdminsSearchTest`, +2 from the previous run)
- ts-check clean
- API contract unchanged (no handler edits)
- Playwright `admin-admins-density.spec.ts`: 6 passed (chromium + mobile-chromium projects)

CI re-running on the new commit.

PR #1254, opened 2026-05-07
fix(banlist+commslist): empty pagination, parity filter bar, pressed Hide-inactive (#1225, #1226, #1230)

### LINKED ISSUE #1225: Banlist + commslist render an empty pagination shell ('displaying 0 - 0 of 0 results') on first-run
## Surface

`banlist` and `commslist` first-run / empty state — the public ban list and the public comms list both render a pagination shell containing the literal string "displaying 0 - 0 of 0 results" / "Showing 0-0 of 0 + Prev/Next" inside a card *below* the empty state. The empty state already tells the user there's nothing here; the pagination card is dead chrome that adds vertical weight and reads as a placeholder bug on a fresh install.

| Light | Dark |
| --- | --- |
| banlist light | banlist dark |
| commslist light | commslist dark |

## Current behavior

[`web/pages/page.banlist.php:884`](web/pages/page.banlist.php#L884) builds `$ban_nav` unconditionally:

```php
$ban_nav = 'displaying&nbsp;' . $BansStart . '&nbsp;-&nbsp;' . $BansEnd . '&nbsp;of&nbsp;' . $BanCount . '&nbsp;results';
```

[`web/themes/default/page_bans.tpl:342-351`](web/themes/default/page_bans.tpl#L342-L351) only checks `{if !empty($ban_nav)}`, which is always true since the string is built even when `$BanCount === 0`. Result: a card titled "displaying 0 - 0 of 0 results" renders below the empty state on every fresh install. `commslist` is the same shape (its `$comm_nav` equivalent in [`web/pages/page.commslist.php`](web/pages/page.commslist.php)).

## Expected behavior

When the result count is zero, the pagination card should not render at all. The empty state already owns the "this list is empty" message and the (gated) primary CTA per the `.empty-state` first-run pattern documented in `AGENTS.md` ("Empty states").

## Suggested fix

In [`web/pages/page.banlist.php`](web/pages/page.banlist.php#L884) and the equivalent block in [`web/pages/page.commslist.php`](web/pages/page.commslist.php), short-circuit when `$BanCount === 0` (or its commslist equivalent):

```php
if ($BanCount === 0) {
    $ban_nav = '';
} else {
    $ban_nav = 'displaying&nbsp;' . $BansStart . '…';
    // …existing prev/next/page-jump assembly…
}
```

The template's existing `{if !empty($ban_nav)}` guard then collapses the card cleanly. No template change required.

### LINKED ISSUE #1226: Public banlist filter bar lacks the server + time selects the public commslist has
## Surface

Public `banlist` vs public `commslist` filter bars. The two surfaces moderate the same family of records (per-server enforcement actions over time), but the banlist filter UX is a strict subset of the commslist one — there's no way to narrow the public ban list by server or by time on the marquee page.

| Light | Dark |
| --- | --- |
| banlist light | banlist dark |
| commslist light | commslist dark |

## Current behavior

[`web/themes/default/page_bans.tpl:87-113`](web/themes/default/page_bans.tpl#L87-L113) ships only:

- A `searchText` text input (player / SteamID / IP)
- Five chips: `All / Permanent / Active / Expired / Unbanned`

[`web/themes/default/page_comms.tpl:69-83`](web/themes/default/page_comms.tpl#L69-L83) ships:

- A search input
- A `server` `<select>` (`All servers` + one `<option>` per server)
- A `time` `<select>` (`All time / Today / Last 7 days / Last 30 days`)
- An `Apply` submit button
- Filter chips (`All / Active / Mute / Gag / Silence`)

Operators trying to answer "what bans hit our prophunt server last week?" can do it on `commslist` in two clicks; on `banlist` they have to drop into the `box_admin_bans_search.tpl` advanced-search box (`?advType=where_banned`/`?advType=date`), which most public visitors will never find.

## Expected behavior

The banlist filter bar should match the commslist contract: search + `server` select + `time` select + chips, all submitting via a single `Apply` button (or via the chip-state JS sync `banlist.js` already does for the chips). The advanced search box stays as the kitchen-sink power-user tool; the inline bar covers the 80% case.

## Suggested fix

Lift the `server` + `time` selects + `Apply` button from [`web/themes/default/page_comms.tpl:71-83`](web/themes/default/page_comms.tpl#L71-L83) into [`web/themes/default/page_bans.tpl`](web/themes/default/page_bans.tpl#L87-L113) under the existing `<form id="banlist-filters">`. Wire the `server` value into `box_admin_bans_search.tpl`'s `?advType=where_banned&advSearch=<sid>` shape and the `time` value into the existing date predicate in [`web/pages/page.banlist.php`](web/pages/page.banlist.php) (the comms page's "Last N days" mapping is the reference). The `BanListView` DTO already carries enough state for the filter bar; add `server_list` and `filters.{server,time}` props mirroring `CommsListView`.

Pair with a paired update to [`web/includes/View/BanListView.php`](web/includes/View/BanListView.php) and a row in `web/tests/e2e/specs/responsive/` if the new selects affect the mobile chrome layout.

### LINKED ISSUE #1230: 'Hide inactive' toggle on banlist + commslist reads as a one-shot action, not a stateful toggle
## Surface

`banlist` and `commslist` page header. The "Hide inactive" / "Show inactive" toggle is rendered as an outline button visually identical to the action buttons sitting next to it (`Export Steam`, `Export IP`, `Add ban`, `Add comm block`). The user can't tell from the chrome alone whether the toggle is currently on or off — the icon does swap (`eye` vs `eye-off`), but the button's outer shape, color, and weight don't communicate the binary state.

| Light | Dark |
| --- | --- |
| banlist light | banlist dark |

## Current behavior

[`web/themes/default/page_bans.tpl:74`](web/themes/default/page_bans.tpl#L74):

```smarty
<a class="btn btn--secondary btn--sm" href="index.php?p=banlist&hideinactive={if $hidetext == 'Hide'}true{else}false{/if}{$searchlink}">{$hidetext} inactive</a>
```

[`web/themes/default/page_comms.tpl:38-44`](web/themes/default/page_comms.tpl#L38-L44):

```smarty
<a class="btn btn--secondary btn--sm" href="…hideinactive={if $hide_inactive}false{else}true{/if}…"
   data-testid="toggle-hide-inactive">
    <i data-lucide="{if $hide_inactive}eye{else}eye-off{/if}"></i>
    {if $hide_inactive}Show inactive{else}Hide inactive{/if}
</a>
```

The icon + label *do* update with state, but the outer button class is `.btn--secondary btn--sm` — same as `Export Steam`, `Export IP`, etc. There's no `aria-pressed`, no fill-vs-outline state distinction, no underline. A user landing on the page after a deep-link or after another admin set the toggle has no visual cue that the list is currently filtered.

## Expected behavior

The control should read as a binary state, not as a one-shot action. Two reasonable shapes:

- **Switch / chip with pressed state.** Add `aria-pressed="{if $hide_inactive}true{else}false{/if}"` and a CSS rule that fills the chip (e.g. dark background + white text, mirroring the active filter chip below) when `aria-pressed="true"`. Easy lift, lives in the existing button class system.
- **Segmented control.** `[All bans] [Hide expired]` two-button group, one of which is always pressed. More chrome, but unambiguous about "what is this list currently showing".

Either way, the visual state must be readable from a glance at the chrome — without reading the button label.

## Suggested fix

Smallest change that ships value: add `aria-pressed` to both call sites and a CSS rule keyed off it.

[`web/themes/default/page_bans.tpl:74`](web/themes/default/page_bans.tpl#L74):

```smarty
<a class="btn btn--secondary btn--sm"
   aria-pressed="{if $hidetext == 'Show'}true{else}false{/if}"
   href="…">{$hidetext} inactive</a>
```

[`web/themes/default/page_comms.tpl:38`](web/themes/default/page_comms.tpl#L38) — same shape.

[`web/themes/default/css/theme.css`](web/themes/default/css/theme.css) — add a single rule:

```css
.btn--secondary[aria-pressed="true"] {
    background: var(--accent-surface);
    color: var(--accent-on);
    border-color: var(--accent);
}
```

(Use whatever variable names the rest of the theme uses; the dark banlist's "Bans" sidebar item is a good colour reference.) Cover the new contract with an E2E spec under `web/tests/e2e/specs/smoke/` that asserts `aria-pressed` flips after the Hide/Show round-trip.

rumblefrog:
## Summary

Three fixes on the public banlist + commslist surfaces:

- **#1225**: Drop the empty pagination shell on first-run. `page.banlist.php` and `page.commslist.php` now short-circuit `$ban_nav` / `$comm_nav` to `''` when the count is 0, so the template's existing `{if !empty(...)}` guard collapses the card.
- **#1226**: Public banlist filter bar gains the `server` + `time` selects + `Apply` button the public commslist already had. Wires `BanListView` with `server_list` + `filters.{server,time}` props mirroring `CommsListView`. Existing `?advType=where_banned&advSearch=<sid>` URLs still resolve via the legacy shim.
- **#1230**: `Hide / Show inactive` button on both surfaces now sets `aria-pressed`, and a single `.btn--secondary[aria-pressed="true"]` rule in `theme.css` paints the chip when active.

Closes #1225
Closes #1226
Closes #1230

## Test plan

- [ ] Fresh-install banlist (no bans) → empty state alone, no pagination card.
- [ ] Fresh-install commslist (no comms) → same.
- [ ] Public banlist: pick a server in the new select, hit Apply → list filtered, URL preserves the filter, paging links keep the filter.
- [ ] Public banlist: pick "Last 7 days" → date-filtered correctly.
- [ ] Hide-inactive button reads as filled/pressed when filtering, outline when not, in both light and dark.
- [ ] PHPStan + PHPUnit pass on CI (esp. SmartyTemplateRule on the new BanListView props).

### DIFF
--- a/web/includes/View/BanListView.php
+++ b/web/includes/View/BanListView.php
@@ -35,6 +35,8 @@ final class BanListView extends View
      * @param int|false                           $comment        Bid being commented on, or false when not in comment-edit mode.
      * @param int                                 $page           Active pagination page (or -1 when not paginated).
      * @param array<int, array<string,mixed>>|string $othercomments  Sibling comments shown beneath the editor; "None" string when the ban has no other comments.
+     * @param list<array{sid: int, name: string}> $server_list    Enabled servers for the public filter bar's `<select name="server">` (#1226).
+     * @param array{search: string, server: string, time: string} $filters Current filter state — drives the sticky filter bar's pre-fill + active selected `<option>` (#1226).
      */
     public function __construct(
         public readonly array $ban_list,
@@ -65,10 +67,19 @@ final class BanListView extends View
         // sees the body copy without the link). Splatted from
         // `Perms::for($userbank)` in `web/pages/page.banlist.php`.
         public readonly bool $can_add_ban,
-        // #1207: detects whether the current request applied any filter
-        // (search text / advSearch / hide-inactive). Drives the
-        // first-run-vs-filtered split in the empty-state shape.
+        // #1207 + #1226: detects whether the current request applied any
+        // filter (search text / advSearch / hide-inactive / server /
+        // time). Drives the first-run-vs-filtered split in the empty-
+        // state shape.
         public readonly bool $is_filtered,
+        // #1226: public filter parity with `CommsListView`. The
+        // server list mirrors `CommsListView::$servers`; it's named
+        // `server_list` here to match the existing
+        // `box_admin_bans_search.tpl` convention so a future
+        // consolidation between the inline filter bar and the
+        // advanced-search box doesn't have to rename the property.
+        public readonly array $server_list,
+        public readonly array $filters,
     ) {
     }
 }
--- a/web/pages/page.banlist.php
+++ b/web/pages/page.banlist.php
@@ -231,6 +231,46 @@ if (isset($_SESSION["hideinactive"])) {
     $hideinactiven = "";
 }
 
+// #1226: public ?server=<sid> + ?time=<1d|7d|30d> filter parity with
+// the commslist URL surface. Both are applied as additional AND
+// predicates on top of whichever upper branch (searchText / no
+// filter / advSearch) is active, so they compose with the existing
+// search box AND with the legacy ?advType=where_banned shim.
+//
+// Mirrors the $hideinactive / $hideinactiven pattern: build two
+// fragments — `$publicFilterAnd` (` AND a = ? AND b = ?`, appended
+// to a query that already opens a WHERE) and `$publicFilterWheren`
+// (` WHERE a = ? AND b = ?`, used when no other clause has opened
+// one yet).
+$serverFilter = isset($_GET['server']) ? trim((string) $_GET['server']) : '';
+$timeFilter   = isset($_GET['time'])   ? trim((string) $_GET['time'])   : '';
+$publicFilterTimeMap = [
+    '1d'  => 86400,
+    '7d'  => 7 * 86400,
+    '30d' => 30 * 86400,
+];
+
+$publicFilterClauses = [];
+$publicFilterArgs    = [];
+$publicFilterLink    = '';
+if ($serverFilter !== '' && ctype_digit($serverFilter)) {
+    $publicFilterClauses[] = 'BA.sid = ?';
+    $publicFilterArgs[]    = (int) $serverFilter;
+    $publicFilterLink     .= '&server=' . urlencode($serverFilter);
+}
+if ($timeFilter !== '' && isset($publicFilterTimeMap[$timeFilter])) {
+    $publicFilterClauses[] = 'BA.created >= ?';
+    $publicFilterArgs[]    = time() - $publicFilterTimeMap[$timeFilter];
+    $publicFilterLink     .= '&time=' . urlencode($timeFilter);
+}
+
+$publicFilterAnd     = '';
+$publicFilterWheren  = '';
+if ($publicFilterClauses) {
+    $joined             = implode(' AND ', $publicFilterClauses);
+    $publicFilterAnd    = ' AND ' . $joined;
+    $publicFilterWheren = ' WHERE ' . $joined;
+}
 
 if (isset($_GET['searchText'])) {
     $searchText = trim($_GET['searchText']);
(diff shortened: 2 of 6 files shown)


rumblefrog:
Post-merge review (subagent reviewer): all three issues are fixed correctly and the CI gate set was green at merge. Two follow-up notes from a thorough read of the diff:

### 1. `page.banlist.php` branch 1: operator-precedence bug exposed by the new `?server` / `?time` filters

The `searchText` branch's `WHERE` is built as a flat OR/AND chain:

```php
"WHERE " . $search_ips . $authidClause . " or BA.name LIKE ? or BA.reason LIKE ?" . $hideinactive . $publicFilterAnd
```

which renders to e.g.

```sql
WHERE BA.ip LIKE ? OR BA.authid REGEXP ? OR BA.name LIKE ? OR BA.reason LIKE ?
      AND RemoveType IS NULL AND BA.sid = ? AND BA.created >= ?
```

MySQL precedence (`AND` binds tighter than `OR`) parses that as:

```sql
WHERE BA.ip LIKE ? OR BA.authid REGEXP ? OR BA.name LIKE ?
   OR (BA.reason LIKE ? AND RemoveType IS NULL AND BA.sid = ? AND BA.created >= ?)
```

So `?searchText=fishy&server=42&time=7d` returns every \"fishy\" hit (ip / authid / name) on **all** servers, not just server 42 within the last 7 days — the filter only constrains the reason-match arm. This was already latent for `$hideinactive` (#1225 era), but the new `?server` / `?time` make it observably wrong because users *expect* a visible select to filter the whole list.

Cheap follow-up: wrap the OR group in parens so the trailing AND fragments apply globally:

```php
"WHERE (" . $search_ips . $authidClause . " OR BA.name LIKE ? OR BA.reason LIKE ?)" . $hideinactive . $publicFilterAnd
```

Branches 2 and 3 are AND-only so they're fine.

### 2. `page.commslist.php` `?server` / `?time` are still cosmetic

The PR's stated goal is \"banlist filter bar parity with commslist\" (#1226), but the public commslist's `?server` / `?time` GET params only flow into the `$filters` array consumed by the View for the `<select>` pre-fill — they never enter any SQL `WHERE`. So the banlist's filter (post-PR) actually works *better* than the commslist's: picking \"Last 7 days\" in the comms `<select data-testid=\"comms-time-filter\">` round-trips the URL but doesn't change the result set.

That's a separate pre-existing bug, not a regression — but worth a follow-up issue so the parity goes both ways.

### Otherwise LGTM

- `:prefix_` discipline: the new `SELECT … FROM \`:prefix_servers\`` uses the placeholder, no `DB_PREFIX` inlines in the diff. ✓
- `BanListView` props (`server_list`, `filters`) match the template's reads, SmartyTemplateRule is satisfied. ✓
- `aria-pressed` + `role=\"button\"` + the `.btn--secondary[aria-pressed=\"true\"]` rule: vars all resolve in light + dark (`--zinc-900` / `--brand-700` / `--on-accent`); axe `aria-allowed-attr` is happy with the `role` follow-up commit.
- Side-effect on `page_admin_settings_themes.tpl:146` (\"In use\" disabled button): strict UX improvement — the chip-style fill matches the sidebar's `aria-current=\"page\"` treatment so the active selection is clearly readable instead of competing with neighbouring outline buttons.
- `$banlistIsFiltered` correctly extends to `$publicFilterClauses !== []` so the filtered empty-state shape kicks in when only the new filters are active. ✓
- Conventional Commits ✓, all five gates green ✓.

cc reviewer-loop for the operator-precedence follow-up.

PR #1277, opened 2026-05-07
fix(admin-groups): coerce flag bitmask to unsigned 32-bit + widen schema (#1272)

### LINKED ISSUE #1272: Web admin groups bitmask preview goes negative when high-bit flags toggled — JS `|=` returns Int32
## Problem

On `?p=admin&c=groups&section=list`, toggling `ADMIN_UNBAN_GROUP_BANS` (or any combination that includes flags at bit 31 or above) makes the live "bitmask" preview render a **negative** number, and the same negative value is what the form submits to `groups.edit` on Save. The high-bit web permission flags can't round-trip through JS bitwise math without sign-bit corruption.

## Reproduce

1. `./sbpp.sh up && ./sbpp.sh db-seed`.
2. Sign in as `admin/admin`. Visit `?p=admin&c=groups&section=list`.
3. Pick any web admin group on the left rail (or create one via "Add a group" if the dev seed is empty).
4. Tick **Unban Group Bans** (`ADMIN_UNBAN_GROUP_BANS`, value `2147483648` = 2^31). Optionally tick **All Web Permissions** (`ALL_WEB`, value `4294966783`) by ticking everything.
5. Watch the bitmask badge top-right of the flag grid (`<span data-testid="flag-bitmask">`):
   - With only `ADMIN_UNBAN_GROUP_BANS` checked: shows `-2147483648 bitmask`.
   - With many flags including bit-31: shows a negative seven-or-ten-digit number.
6. Click Save. The negative value is sent in `web_flags` to `Actions.GroupsEdit`.

Expected: the badge mirrors the unsigned 32-bit OR-sum (e.g. `2147483648`, not `-2147483648`), and the API receives an unsigned value.

## Root cause

JS bitwise operators (`|=`, `&`, `^`, `~`, `<<`, `>>`) coerce both operands to **signed 32-bit integers** via `ToInt32` (ECMAScript spec) before operating, then return another Int32. The instant any operand has bit 31 set, the result is interpreted as negative.

Both compute paths in `page_admin_groups_list.tpl` hit this:

```363:367:web/themes/default/page_admin_groups_list.tpl
    var bitmask = 0;
    var checks = form.querySelectorAll('input[name="flags[]"]:checked');
    for (var i = 0; i < checks.length; i++) {
        bitmask |= Number(checks[i].value);
    }
```

```412:418:web/themes/default/page_admin_groups_list.tpl
        var bitmask = 0;
        var checks = grid.querySelectorAll('input[name="flags[]"]:checked');
        for (var i = 0; i < checks.length; i++) {
            var input = /** @type {HTMLInputElement} */ (checks[i]);
            bitmask |= Number(input.dataset.flagValue || input.value);
        }
        preview.textContent = bitmask + ' bitmask';
```

The flag values that trigger this:

```87:95:web/configs/permissions/web.json
    "ADMIN_UNBAN_OWN_BANS": {
        "value": 1073741824,
        "display": "Unban Own Bans"
    },

    "ADMIN_UNBAN_GROUP_BANS": {
        "value": 2147483648,
        "display": "Unban Group Bans"
    },
```

```157:160:web/configs/permissions/web.json
    "ALL_WEB": {
        "value": 4294966783,
        "display": "All Web Permissions"
    }
```

`ADMIN_UNBAN_GROUP_BANS = 2^31` is exactly the boundary that flips the sign in Int32. `ALL_WEB ≈ 2^32` is unrepresentable in Int32 entirely (it folds back to a small negative under `ToInt32`).

The schema is the second half of the trap:

```91:97:web/install/includes/sql/struc.sql
CREATE TABLE IF NOT EXISTS `{prefix}_groups` (
  `gid` int(6) NOT NULL auto_increment,
  `type` smallint(6) NOT NULL default '0',
  `name` varchar(128) character set {charset} NOT NULL default 'unnamed',
  `flags` int(10) NOT NULL,
  PRIMARY KEY  (`gid`)
) ENGINE=InnoDB DEFAULT CHARSET={charset};
```

`flags` is **`int(10)` signed** in MariaDB — display width 10, but the actual storage is a signed 32-bit `INT`. Range: `-2147483648` to `2147483647`. So `ADMIN_UNBAN_GROUP_BANS = 2147483648` is exactly **one beyond** signed-INT max and `ALL_WEB = 4294966783` is well past it. Even if the JS bug were fixed, the column can't represent these values as positive integers.

The chain works "accidentally" today because PHP's 64-bit signed int and MariaDB's 32-bit signed int conspire to round-trip the bit pattern (the high bit gets stored as the sign, and PHP-side `&` still produces the right boolean for the `{if ($flags & $value) == $value}` checkbox restore). But the visible badge shows a negative number, the API receives a negative number, and any tooling that introspects the stored value (audit log, third-party query) sees a confusing negative.

## Suggested fix

Two layers — ship the JS fix first (resolves the user-visible badge regression), then the schema/API fix (resolves the round-trip semantics):

### 1. JS — coerce to unsigned 32-bit at display + send (small, ships now)

`>>> 0` (unsigned-right-shift by zero) is the standard idiom to convert a signed Int32 back to its unsigned interpretation. After the OR loop:

```js
var bitmask = 0;
var checks = grid.querySelectorAll('input[name=\"flags[]\"]:checked');
for (var i = 0; i < checks.length; i++) {
    bitmask |= Number(checks[i].value);
}
bitmask = bitmask >>> 0;   // <- convert to unsigned 32-bit before display / send
```

Apply to both compute sites — `SbppGroupsSave` (the actual save path) and the `[data-testid="flag-grid"] change` listener (the preview). Pull the OR-fold into a small helper inside the inline `<script>` block so the two call sites stay in sync.

(Bonus: PHPStan-style assertion in a comment near the helper — "JS Int32 coercion would render bit-31 flags negative without the `>>> 0` wrap; see `web/configs/permissions/web.json` for the values that hit this.")

### 2. Schema + API — store the value as unsigned (paired migration)

- Add a `web/updater/data/<N>.php` migration that runs `ALTER TABLE \`:prefix_groups\` MODIFY COLUMN \`flags\` INT UNSIGNED NOT NULL`. Register in `web/updater/store.json`. Pair the change with a `struc.sql` edit so fresh installs land on the same column type. Per AGENTS.md ("Updater migrations"), the migration must be idempotent — guard with a `SHOW COLUMNS` check or run a no-op `ALTER` that's safe to repeat.
- The `:prefix_admins.extraflags` column has the same shape problem and the same `bit-31 / ALL_WEB` bug surface (set via `web/pages/admin.edit.adminperms.php`); audit it in the same migration. Same data-correction story.
- API handler `web/api/handlers/groups.php` already does `(int)($params['web_flags'] ?? 0)`. With unsigned storage, the cast still works on 64-bit PHP (which treats it as a 64-bit signed int and stores fine in `INT UNSIGNED`); on 32-bit PHP, audit `Database::query` / `bindParam` to ensure the binding doesn't drop the high bit. Add a unit test in `web/tests/api/GroupsTest.php` that round-trips `ALL_WEB` and `ADMIN_UNBAN_GROUP_BANS` as positive integers.
- Existing data correction: any group that was saved post-bug with a negative `flags` value needs to be coerced back to the unsigned interpretation. The migration can do this in-place: `UPDATE \`:prefix_groups\` SET \`flags\` = CAST(\`flags\` AS UNSIGNED) WHERE \`flags\` < 0` (run BEFORE the column type change, while the column is still signed). On `INT UNSIGNED` post-migration, no rows can be negative, so this becomes a no-op for fresh installs.

### 3. Regression test

Add a Playwright spec that:

1. Creates a group with `ADMIN_UNBAN_GROUP_BANS` toggled.
2. Asserts the badge shows `2147483648 bitmask` (no minus sign).
3. Saves, reloads, and asserts the checkbox is still checked AND the badge still reads `2147483648`.

Anchor on `[data-testid="flag-bitmask"]` per AGENTS.md's "Selectors must use #1123's testability hooks" rule (the testid was added in #1258 specifically for this).

## Acceptance criteria

- [ ] Toggling `ADMIN_UNBAN_GROUP_BANS` on the web-groups detail editor shows `2147483648 bitmask` (no minus sign) in the live preview.
- [ ] Toggling all flags shows the correct unsigned OR-sum (matches `ALL_WEB = 4294966783` when every flag is on).
- [ ] Save round-trip: pick a flag combination that includes bit 31, Save, reload — the same flags re-render checked AND the badge reads the same unsigned number.
- [ ] `:prefix_groups.flags` widened to `INT UNSIGNED` via paired `struc.sql` edit + `web/updater/data/<N>.php` migration registered in `web/updater/store.json`. Same audit applied to `:prefix_admins.extraflags`. Both migrations idempotent.
- [ ] Playwright spec asserts the badge contract using `[data-testid="flag-bitmask"]`.
- [ ] PHPUnit case in `GroupsTest` round-trips `ALL_WEB` and `ADMIN_UNBAN_GROUP_BANS` through `groups.edit` → DB → `groups.list` as positive integers.

## Notes

- The `flag-bitmask` testid was introduced in #1258 ("admin-groups: drop nested chip borders, live-update bitmask preview") for exactly this kind of contract anchoring; #1258 wired the live preview but didn't catch the Int32 coercion. Cross-reference in the fix-PR.
- Adjacent: `web/pages/admin.edit.adminperms.php` (per-admin permission editor) builds `web_flags` PHP-side, so it doesn't have the JS Int32 bug — but the SCHEMA fix from layer 2 still applies to `:prefix_admins.extraflags`. Bundle in the same migration.
- Out of scope: rethinking the permission flag scheme to fit in 31 bits (e.g. dropping `ALL_WEB` as a stored value and computing it dynamically). The flag values are part of the on-disk + on-wire contract — third-party themes and SourceMod plugins read them.

rumblefrog:
Closes #1272.

## Summary

Fixes the `?p=admin&c=groups&section=list` flag editor showing (and
saving) bit-31 web permission flags as negative numbers — the
`ADMIN_UNBAN_GROUP_BANS = 2^31` toggle rendered `-2147483648 bitmask`,
`ALL_WEB = 4294966783` rendered a small negative, and the same
negative flowed through `Actions.GroupsEdit` to the DB. Three
layers, one PR.

### 1. JS — unsigned coercion at both compute sites

`web/themes/default/page_admin_groups_list.tpl`: pull the OR-fold
into a `SbppFoldFlags(root)` helper that applies `>>> 0`
(unsigned-right-shift by zero — the standard JS idiom for
projecting a signed Int32 onto its unsigned 32-bit
interpretation). Used by both `SbppGroupsSave` (the actual save
path — was the bug source) AND the grid's `change` listener (live
preview) so they can never drift on the coercion rule. JSDoc on
the helper points at the issue.

### 2. Schema — `INT UNSIGNED` end-to-end

- `web/install/includes/sql/struc.sql`: both `:prefix_groups.flags`
  and `:prefix_admins.extraflags` widened to `INT(10) UNSIGNED`
  for fresh installs.
- `web/updater/data/806.php`: paired migration for existing
  installs. 3-step dance per column:
  1. Widen to `BIGINT NOT NULL` (preserves every legacy value
     verbatim, including the negative bit-patterns).
  2. `UPDATE ... SET col = col + 4294967296 WHERE col < 0` —
     reinterprets each pre-fix negative bit-pattern as its
     unsigned 32-bit equivalent (`-2147483648 → 2147483648`).
  3. Narrow to `INT UNSIGNED NOT NULL`.

  We can't do the simpler `UPDATE ... CAST(neg AS UNSIGNED)` while
  the column is still signed: under `STRICT_TRANS_TABLES` (the
  MariaDB 10.x default) assigning a value past the signed-INT max
  to a signed-INT column raises `1264 Out of range value`. The
  BIGINT widening is what makes the assignment legal.

  **Idempotent.** Re-running on an already-migrated DB:
  - Step 1: shape change `INT UNSIGNED → BIGINT SIGNED` is legal
    (every existing value fits in BIGINT SIGNED) and reversible.
  - Step 2: `WHERE col < 0` matches no rows (no negatives left).
  - Step 3: shape change `BIGINT → INT UNSIGNED` is a no-op when
    every value is already in the unsigned 32-bit range.

  Verified by running the migration twice against the dev DB with
  an injected negative test row — the second pass converged on
  the same shape with the same data.

- `web/updater/store.json`: registers `806.php`.

### 3. Tests

- **PHPUnit** (`web/tests/api/GroupsTest.php`): data-provider'd
  round-trip through `groups.edit` for `ADMIN_UNBAN_GROUP_BANS
  (2^31)`, `ALL_WEB (4294966783)`, and a high+low mix
  (`2^31 + 1`). Asserts the stored value equals the input AND is
  non-negative. Companion `groups.add` test locks in that the
  schema widening covers the create path too.
- **Playwright** (`web/tests/e2e/specs/flows/admin-groups-bitmask.spec.ts`):
  seeds an empty web admin group via `Actions.GroupsAdd`, opens
  `?p=admin&c=groups&section=list`, ticks the
  `[data-testid="flag-unban_group_bans"]` checkbox, asserts
  `[data-testid="flag-bitmask"]` reads `2147483648 bitmask` (and
  explicitly does NOT contain a `-`), Saves, reloads, asserts
  badge + checkbox state both persist. Anchored on
  `data-testid` per AGENTS.md "Selectors must use #1123's
  testability hooks". Pinned to chromium (mobile gives no
  additional coverage; flow specs that mutate the shared
  `sourcebans_e2e` DB race across projects).

## Doc-sync

- `ARCHITECTURE.md` schema table is at table-level only and
  doesn't reference column types for `sb_groups.flags` /
  `sb_admins.extraflags`, so no edit needed there per the
  AGENTS.md "Keep the docs in sync" rule.
- `struc.sql` + paired updater + `store.json` all aligned per
  AGENTS.md "Updater migrations".

## Test plan

- [x] `./sbpp.sh phpstan` — clean
- [x] `./sbpp.sh test --filter=GroupsTest` — 21/21 pass (including
      3 new high-bit cases + 1 add-path case)
- [x] `./sbpp.sh test` — full suite 365/365 pass
- [x] `./sbpp.sh composer api-contract` — no diff (`groups.edit`
      shape unchanged; integers in / integers out)
- [x] `./sbpp.sh e2e --grep "groups bitmask"` — 1 passes, 1
      skipped (mobile gating)
- [x] Manual idempotency verification: ran migration twice
      against the dev DB with an injected `-2147483648` test row;
      first pass corrected to `2147483648` and widened both
      columns, second pass was a no-op on shape and matched no
      rows on the negative-rewrite.


rumblefrog (commented):
**Self-review (subagent reviewer): LGTM, all gates green.**

(GitHub blocks self-approval, so this is a comment-style review with the audit trail. Verdict is approve.)

Verified end-to-end against the issue's acceptance criteria + AGENTS.md "Updater migrations" / "Database" / "Frontend" rules.

## JS fix (`page_admin_groups_list.tpl`)

- `SbppFoldFlags(root)` helper at lines 359-387 wraps the OR-fold and applies `>>> 0` at the return — the canonical signed-Int32 → uint32 idiom.
- Both compute sites go through the helper:
  - `SbppGroupsSave` (line 394) — the save path that posts `web_flags` to `Actions.GroupsEdit`.
  - The `[data-testid="flag-grid"] change` listener (line 439) — the live preview.
- JSDoc on the helper points back to #1272 and explains why bare `|=` is a trap (ECMAScript `ToInt32`).

## Schema (`struc.sql` + migration `806.php`)

- `web/install/includes/sql/struc.sql` widens both `:prefix_admins.extraflags` (line 11) and `:prefix_groups.flags` (line 95) to `int(10) UNSIGNED NOT NULL`. Fresh installs converge to the same shape.
- `web/updater/data/806.php` is the paired migration. The 3-step BIGINT widen → negative-rewrite → INT UNSIGNED narrow is necessary because under `STRICT_TRANS_TABLES` (MariaDB 10.x default) you can't `CAST(neg AS UNSIGNED)` and assign back into a still-signed column without `1264 Out of range value`. The BIGINT widening is what makes the assignment legal. Comment block at the top of the file documents this trade-off.
- Migration **idempotency walked through**:
  - **First run on legacy 1.x install with negatives** (e.g. `-2147483648`, `-513`):
    1. ALTER MODIFY ... BIGINT — values preserved verbatim (BIGINT SIGNED holds the 32-bit range).
    2. UPDATE ... SET col = col + 4294967296 WHERE col < 0 — `-2^31 → 2^31 = ADMIN_UNBAN_GROUP_BANS`, `-513 → 4294966783 = ALL_WEB`.
    3. ALTER MODIFY ... INT UNSIGNED — every value now in `[0, 2^32-1]`.
  - **Second run** (or fresh install + run): values already in `[0, 2^32-1]`. Step 1's BIGINT widen is reversible (every uint32 value fits in BIGINT SIGNED). Step 2's `WHERE col < 0` matches no rows. Step 3 is a no-op shape change. Worker also verified by running twice against the dev DB with an injected `-2147483648` test row.
- Uses `:prefix_` placeholder per "Database" rule (lines 63, 72, 80, 89, 96, 103). Each `$this->dbs->query` / `->execute()` call has the `// @phpstan-ignore variable.undefined` annotation per "Updater migrations" canonical shape.
- `web/updater/store.json` registers `806`. Sequence is correct (next integer above `805`).
- No FK / index references on `flags` or `extraflags` — `MODIFY COLUMN` is shape-only and there are no constraints to preserve.

## Tests

- `GroupsTest::testEditRoundTripsHighBitFlagsAsPositiveIntegers` (line 209) is data-provider'd over `2^31`, `4294966783`, and `2^31 + 1`. Asserts `assertSame($flags, $stored)` AND `assertGreaterThanOrEqual(0, $stored)` so a bit-pattern-coincidence regression still trips the second assertion.
- `GroupsTest::testAddStoresHighBitFlagAsPositiveInteger` (line 252) locks in the `groups.add` create-path coverage.
- Playwright spec (`admin-groups-bitmask.spec.ts`):
  - Anchors on `[data-testid="flag-bitmask"]` per "Selectors must use #1123's testability hooks".
  - `await expect(bitmaskBadge).toHaveText(\`${'\$'}{FIXTURE.unbanGroupBansValue} bitmask\`)` + the defensive `not.toContainText('-')` so a numeric-match regression with a leading `-` still fails on the sign separately.
  - Save → reload round-trip is asserted at lines 178-188 (re-loads the route, re-locates the badge + checkbox, asserts both persist).

## CI

All four applicable gates SUCCESS — Static analysis (PHPStan), PHPUnit, Playwright + axe, tsc --checkJs. The api-contract workflow correctly didn't run (path-filtered to `web/api/handlers/**` / `web/configs/permissions/**` / etc.; this PR touches neither, which is itself proof the JSON wire format is unchanged).

## Doc-sync

`ARCHITECTURE.md` schema table is table-level only (no per-column types for `sb_groups.flags` / `sb_admins.extraflags`), so the worker correctly skipped the doc edit per AGENTS.md "Keep the docs in sync" rule. `struc.sql` + paired updater + `store.json` all aligned per "Updater migrations".

Issue #1259, opened 2026-05-07
admin chrome: adopt Settings-style vertical sidebar for sub-paged admin routes (servers/mods/groups/comms) instead of the horizontal AdminTabs strip


rumblefrog:
## Surface

Admin Settings (`?p=admin&c=settings`) renders its sub-section nav as a 14rem vertical sidebar of `.sidebar__link` anchors (`page_admin_settings_*.tpl`'s `<nav aria-label="Settings sections">` block). Every other sub-paged admin route — `admin-servers`, `admin-mods`, `admin-groups`, `admin-comms` — surfaces the same `?section=…` architectural pattern (locked in by #1239) but renders it as the horizontal `core/admin_tabs.tpl` strip ("List servers · Add new server").

Two patterns, same routing shape, completely different chrome.

## Current state

- **Settings** (`page_admin_settings_*.tpl`): vertical sidebar in a `grid-template-columns:14rem 1fr` two-column layout. Active section carries `aria-current="page"`. Lucide icon per row.
- **Servers / Mods / Groups / Comms**: horizontal pill strip rendered by `web/includes/AdminTabs.php` → `web/themes/default/core/admin_tabs.tpl`. Active tab carries `aria-current="page"`. No icons; mobile shape is a scrollable chip strip (#1207 ADM-8).
- **Edit-\* pages** (`admin.edit.*.php`): same `AdminTabs` partial called with `[]` so only the trailing Back link renders.

The horizontal strip:

- Visually competes with the page heading on the same row (the heading sits directly below the strip, often without padding — see the sibling padding bug for `admin-admins`).
- Reads as "tabs into one document" rather than "navigation between sibling pages" — even after #1239 routed each section to its own URL.
- Less scannable than a labeled vertical sidebar for routes that have 3+ sections (mods routes Add / List / Game type / Web type, and the strip is already a chip cluster).

## Proposal

Standardise on the Settings-style vertical sidebar across every sub-paged admin route.

Two pieces of work:

1. **Lift the Settings sidebar markup into a parameterized partial** (e.g. `core/admin_sidebar.tpl`) keyed on the same `tabs` + `active_tab` shape `AdminTabs.php` already builds. Each entry needs `name` + `slug` + `url` + optional `icon` (Lucide name); the page handlers add the `icon` field to their existing `$sections` arrays.
[...]

PR #1278, opened 2026-05-07
fix(chrome): pin main sidebar with align-self: flex-start (#1271)

### LINKED ISSUE #1271: Main sidebar (`.sidebar`) scrolls with page content — sticky breaks because flex `align-items: stretch` overrides `height: 100vh`
## Problem

On long pages, the main app sidebar (`<aside class="sidebar">` — the public + admin nav rail rendered by `web/themes/default/core/navbar.tpl`) scrolls down with the page content and goes out of view, instead of staying pinned at the top of the viewport. On a tall surface like `?p=admin&c=bans` (multi-section ToC layout) or `?p=admin&c=audit` (long log scroll), the user reaches the bottom of the page with the sidebar nowhere on screen.

## Reproduce

1. `./sbpp.sh up && ./sbpp.sh db-seed` (medium scale gives enough rows).
2. Sign in as `admin/admin`.
3. Visit any tall page — `?p=admin&c=bans`, `?p=admin&c=audit`, or a fully-populated `?p=banlist`.
4. Scroll to the bottom. The sidebar leaves the viewport.

Expected: the sidebar stays pinned at the top of the viewport while the page content scrolls beneath it (the `position: sticky` contract that the sub-paged admin route's inner sidebar already gets right — see `.admin-sidebar` below).

## Likely root cause

`web/themes/default/css/theme.css` declares the sidebar like this:

```111:116:web/themes/default/css/theme.css
.app { min-height: 100vh; display: flex; }
.sidebar {
  width: 15rem; flex-shrink: 0; height: 100vh; position: sticky; top: 0;
  background: var(--bg-surface); border-right: 1px solid var(--border);
  display: flex; flex-direction: column;
}
```

The intent is correct (sidebar is `position: sticky; top: 0; height: 100vh` inside a `.app` flex parent that grows past the viewport). The bug is the well-known `position: sticky` + flex `align-items: stretch` interaction:

- `.app` is `display: flex` with default `align-items: stretch`. On a long page, `.app`'s height is determined by `.main` (the right column), which is taller than the viewport.
- `align-items: stretch` resizes flex items to match the cross-axis of the container — so the sidebar's *computed* cross-size is the full page height, **not** `100vh`. The `height: 100vh` declaration is "preferred", but `align-items: stretch` overrides it for the actual computed layout box.
- A sticky element whose computed height equals its containing block's height has no room to "stick" — it stays in flow and effectively behaves like `position: static`. As you scroll, it scrolls with the page.

The Pattern A inner admin sidebar (the 14rem rail next to the right content column on `?p=admin&c=servers`/`mods`/`groups`/`comms`/`settings`) gets this right, and the contrast is the smoking gun:

```535:540:web/themes/default/css/theme.css
  .admin-sidebar {
    position: sticky;
    top: 4rem;
    align-self: start;
    margin-bottom: 0;
  }
```

`align-self: start` opts the sticky element OUT of the parent's `align-items: stretch`, so its computed height is its content height (or its declared `height`) and sticky has room to work.

## Investigation plan

1. **Reproduce in the dev stack.** Confirm the symptom against a tall surface (`?p=admin&c=audit` or `?p=admin&c=bans`). Capture before/after screenshots at desktop width (>=1025px, where `.sidebar` is `display: flex`).
2. **Verify the `align-items: stretch` hypothesis.** In DevTools, inspect `.sidebar`'s computed height. If it matches the document height (not 100vh), the hypothesis is correct. If something else is broken (e.g. an ancestor with `transform` or `overflow` creating a new containing block), pivot the fix there.
3. **Confirm parity with `.admin-sidebar`.** The inner admin rail uses `align-self: start` and stays pinned correctly on the same long pages — the fix is to bring `.sidebar` to the same shape.

## Suggested fix

Add `align-self: flex-start` (or `align-self: start`, equivalent in flex) to `.sidebar`, matching what `.admin-sidebar` already does:

```css
.sidebar {
  width: 15rem; flex-shrink: 0;
  height: 100vh;
  position: sticky; top: 0;
  align-self: flex-start;       /* opt out of `.app`'s align-items: stretch */
  background: var(--bg-surface); border-right: 1px solid var(--border);
  display: flex; flex-direction: column;
}
```

Or, equivalently, set `align-items: flex-start` on `.app`. Prefer `align-self` on the child — `.main` is happy to stretch (it has its own internal column-flex layout that fills the row), and changing the parent's default would require auditing every flex sibling.

The fix is one line. Add an inline CSS comment pointing to this issue (and to the Pattern A `.admin-sidebar` precedent) so the next author who refactors `.app` doesn't reintroduce the bug.

## Acceptance criteria

- [ ] On a tall page (`?p=admin&c=audit` with a fully-populated audit log, or `?p=admin&c=bans`), the sidebar stays pinned at the top of the viewport while the right column scrolls. Verified at desktop widths >= 1025px (the breakpoint at which `.sidebar` is `display: flex`).
- [ ] The sidebar's internal scroll (`.sidebar__nav { overflow-y: auto }`) still kicks in correctly when the nav itself overflows 100vh — `align-self: flex-start` doesn't break the inner overflow contract.
- [ ] Mobile drawer behaviour at <=1024px (the `.sidebar.is-open { position: fixed; inset: 0 auto 0 0 }` shape) is unchanged. The fix is a desktop-only concern; the mobile breakpoint already replaces sticky with fixed.
- [ ] An E2E spec under `web/tests/e2e/specs/responsive/` (or similar) anchors on `[data-testid^="nav-"]` being `inViewport()` after scrolling to the bottom of a tall page. Without the spec, this regresses silently the next time someone refactors `.app`.

## Notes

- Adjacent: #1233/#1234/#1237 (the most recent `.sidebar` work — focused on per-row chrome, not layout); #1259 (introduced `.admin-sidebar` with the correct `align-self: start` shape — direct precedent).
- Out of scope: rethinking the broader `.app` flex layout (e.g. moving to a CSS grid shell). This issue is the one-line `align-self` fix that brings `.sidebar` to parity with `.admin-sidebar`. If profiling finds adjacent layout fragility, file follow-ups.

rumblefrog:
## Summary

Closes #1271.

- Adds `align-self: flex-start` to `.sidebar` in `web/themes/default/css/theme.css` so the main app shell sidebar (`<aside class="sidebar" id="sidebar">` from `core/navbar.tpl`) shares the same defensive shape as the Pattern A inner rail (`.admin-sidebar`, which has `align-self: start` per #1259). The explicit `height: 100vh` on `.sidebar` already keeps the cross-axis pinned to the viewport on current evergreen engines (flex `align-items: stretch` only operates on `auto` cross-sizes), but the parity guards against future refactors that drop the explicit height — or that wrap `.app` in a transformed / `overflow: hidden` ancestor — from silently regressing the sticky containing block.
- Adds `web/tests/e2e/specs/responsive/sidebar-sticky.spec.ts` as the regression guard. At desktop (chromium project, 1280x720) it visits `?p=admin&c=bans` (deterministically taller than the viewport on the e2e seed), asserts the sidebar is sticky-pinned at viewport y=0 mid-scroll (the tight regression guard), and confirms it stays in viewport at the bottom of the page. The spec is gated to chromium only — at `<=1024px` the sidebar swaps to `position: fixed` drawer mode (covered by the existing `responsive/sidebar.spec.ts`).

The CSS comment above `.sidebar` documents the constraint for future readers and points back to the `.admin-sidebar` precedent so a future `.app` refactor doesn't silently reintroduce the bug.

## Test plan

- [x] `./sbpp.sh phpstan` — clean
- [x] `./sbpp.sh test` — 361 tests, 1488 assertions
- [x] `./sbpp.sh e2e --grep "sidebar sticky"` — new spec passes on chromium, skipped on mobile-chromium per the project gate
- [x] `./sbpp.sh e2e specs/responsive/sidebar.spec.ts` — mobile sidebar drawer specs still green (no regression to the `position: fixed` drawer shape)
- [x] Manually verified mid-scroll behaviour against the dev stack (`?p=admin&c=bans`) — sidebar bounding-box `top` stays at viewport `y=0` until `.app`'s containing block runs out of room near the bottom of the page.

## Out of scope

- Rethinking the broader `.app` flex layout (e.g. moving to a CSS grid shell). The issue's "Out of scope" note flags this; if profiling finds adjacent layout fragility, a follow-up issue is the right surface.

### DIFF
--- a/web/tests/e2e/specs/responsive/sidebar-sticky.spec.ts
+++ b/web/tests/e2e/specs/responsive/sidebar-sticky.spec.ts
@@ -1,0 +1,130 @@
+/**
+ * Responsive: main sidebar stays sticky-pinned at desktop (#1271).
+ *
+ * Desktop viewport (>=1025px) contract for the main app shell sidebar
+ * (`<aside class="sidebar" id="sidebar">` rendered by
+ * `web/themes/default/core/navbar.tpl`):
+ *
+ *   - The sidebar declares `position: sticky; top: 0; height: 100vh;
+ *     align-self: flex-start;` inside `.app` (`display: flex`). The
+ *     explicit `height: 100vh` keeps the sidebar's cross-axis size at
+ *     the viewport so sticky has room to operate; `align-self:
+ *     flex-start` is defensive parity with the Pattern A inner rail
+ *     (`.admin-sidebar`, which has `align-self: start` for the same
+ *     reason — see #1259). #1271 added the parity so a future refactor
+ *     that drops the explicit height (or wraps `.app` in a transformed
+ *     / overflow-hidden ancestor that breaks the sticky containing
+ *     block) cannot regress the chrome silently.
+ *   - The contract is desktop-only. At <=1024px the sidebar swaps to
+ *     `position: fixed` drawer mode (`responsive/sidebar.spec.ts`
+ *     covers that shape on iPhone-13).
+ *
+ * Choice of test page: `?p=admin&c=bans` rides the multi-section
+ * page-level ToC layout (Add a ban form + Ban protests + Ban
+ * submissions + Import bans + Group ban) and is reliably taller than
+ * the 720px desktop viewport even on the bare e2e seed (the forms
+ * and section chrome are structural, not row-count-driven). The issue
+ * also calls out `?p=admin&c=audit`, but the audit log on
+ * `sourcebans_e2e` is empty by default and would not necessarily
+ * exceed the viewport without extra seeding — the bans route is the
+ * deterministic-tall surface for this assertion.
+ *
+ * Project gating: this whole describe runs only on `chromium`
+ * (Desktop Chrome at 1280x720). The `mobile-chromium` project
+ * (iPhone-13, viewport <1024px) renders the sidebar as a hidden
+ * drawer and the assertion would not apply.
+ */
+
+import { expect, test } from '../../fixtures/auth.ts';
+import { AdminBansPage } from '../../pages/admin/AdminBans.ts';
+
+test.describe('responsive: sidebar sticky at desktop', () => {
+    test.beforeEach(({}, testInfo) => {
+        test.skip(
+            testInfo.project.name !== 'chromium',
+            'Desktop-only contract — see file-level comment.',
+        );
+    });
+
+    test('sidebar stays pinned to the viewport top after scrolling to the bottom of admin-bans', async ({ page }) => {
+        const p = new AdminBansPage(page);
+        await p.goto();
+        await expect(p.pageMounted).toBeVisible();
+
+        // Sanity: the page has to be taller than the viewport for the
+        // sticky test to be meaningful. If the page fits in the
+        // viewport there's nothing to scroll past, and the assertion
+        // would silently pass even on the broken pre-#1271 layout.
+        const { docHeight, viewport } = await page.evaluate(() => ({
+            docHeight: document.documentElement.scrollHeight,
+            viewport: window.innerHeight,
+        }));
+        expect(
+            docHeight,
+            `admin-bans must be taller than the viewport (got docHeight=${docHeight}, viewport=${viewport})`,
+        ).toBeGreaterThan(viewport + 200);
+
+        // The main sidebar's <nav> is uniquely identified by its ARIA
+        // role + name: the topbar's breadcrumb nav uses
+        // aria-label="Breadcrumb" and the inner Pattern A admin
+        // sidebar uses aria-label="<page> sections", so "Primary" is
+        // unambiguous (see core/navbar.tpl). Anchoring on the role +
+        // name pair satisfies AGENTS.md's "ARIA roles" testability
+        // hook contract — never CSS class chains as the primary
+        // selector.
+        const sidebarNav = page.getByRole('navigation', { name: 'Primary' });
+        await expect(sidebarNav).toBeVisible();
+        await expect(sidebarNav).toBeInViewport();
+
+        // Mid-scroll, well clear of the page top, the sticky contract
+        // says the sidebar's bounding-box top must be at viewport y=0
+        // (sticky-pinned to the viewport, not scrolling with the
+        // document). The `aside.sidebar` chrome is the sticky element;
+        // measure its top directly via the live DOM rect. This is the
+        // tight regression guard — `toBeInViewport()` only requires
+        // partial overlap and would tolerate a sidebar drifting tens
+        // of pixels off-screen, but a fully-broken sticky layout
+        // (sidebar scrolls with the page) drives the rect's top to
+        // -scrollY which is unmistakable.
+        await page.evaluate(() => window.scrollTo(0, 500));
+        const midScrollTop = await page.evaluate(() => {
+            const el = document.getElementById('sidebar');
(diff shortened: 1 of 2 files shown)


rumblefrog:
## Reviewer subagent: APPROVE (posted as comment — `gh` blocks self-approval on author-owned PRs)

LGTM — minimal, well-scoped fix.

### CSS fix (`web/themes/default/css/theme.css:111-130`)

- Single line added (`align-self: flex-start;` on `.sidebar`, line 127), matching the `.admin-sidebar` precedent at line 552 (`align-self: start`).
- 13-line comment block above the rule explains the constraint, references #1271 + the `.admin-sidebar` parity, and notes the desktop-only nature of the cross-axis interaction. Future refactor of `.app` won't silently regress this.
- `.admin-sidebar` (line 549-554) is **untouched** — the precedent is preserved as-is.
- Mobile drawer block at `@media (max-width: 1024px)` (line 967-969) is **untouched** — and `align-self` has no effect on `position: fixed` elements anyway, so the drawer mode contract is structurally safe.
- Issue's "Out of scope" note (rethinking `.app` to a CSS grid shell) was respected — diff stat is +144/-0 across exactly the two reported files.

### E2E spec (`web/tests/e2e/specs/responsive/sidebar-sticky.spec.ts`)

- Anchored on AGENTS.md-blessed testability hooks: `getByRole('navigation', { name: 'Primary' })` (matches `<nav role="navigation" aria-label="Primary">` in `core/navbar.tpl:27`) and `[data-testid="nav-home"]` (always rendered for logged-in users — `navbar.php` declares `'home'` with `permission => true`). No CSS class chains as primary selectors.
- Gated correctly to `chromium` only via `test.beforeEach` skip-guard (Desktop Chrome 1280×720 → sidebar in sticky mode); the iPhone-13 `mobile-chromium` project's drawer-mode contract is covered by the existing `responsive/sidebar.spec.ts`.
- Three layered assertions:
  1. **Sanity guard**: `docHeight > viewport + 200` so the test fails loud if `?p=admin&c=bans` ever stops being deterministically tall (avoids silent passes on the broken layout).
  2. **Tight regression guard at mid-scroll**: `expect(Math.round(midScrollTop ?? -1)).toBe(0)` — strict assertion (not `expect.soft`), `?? -1` defensive coercion ensures null can't accidentally pass. This is the precise "sticky pinned to viewport y=0" contract.
  3. **Bottom-of-page**: `toBeInViewport()` — relaxed because at the very bottom of `.app` sticky correctly releases as the containing block runs out of room (correctly explained in the inline comment).
- Choice of `?p=admin&c=bans` over `?p=admin&c=audit` is justified inline: bans is structurally tall on the bare e2e seed (multi-section page-level ToC), audit log is empty on `sourcebans_e2e`.

### CI

All four gates pass: PHPStan, PHPUnit (361 tests / 1488 assertions), ts-check, Playwright + axe.

### AGENTS.md adherence

- Testability hooks: ARIA role + name pair plus `data-testid` — both blessed.
- No `setTimeout` / `waitForTimeout` waits.
- Spec uses the `auth.ts` storage-state fixture (admin/admin minted by `global-setup.ts`).
- `failOnFlakyTests` / `workers: 1` constraints aren't touched.
- No new tag/grep convention introduced — minimal doc-sync correctly assessed.

No blocking findings, no nits. Ready to merge.

PR #1146, opened 2026-05-05
[#1123 B10] Redesign admin home (card grid landing)

rumblefrog:
## Summary

Phase B10 of the v2.0.0 theme rollout (#1123) replaces the Phase A1 stub
for `?p=admin` with the canonical **8-card landing grid** from the
handoff `AdminPanelView` mockup, gates each card on the union of
permission flags the legacy router enforces for that sub-route, and
switches `web/pages/page.admin.php` from the legacy `$theme->assign`
chain to the typed `Renderer::render(new AdminHomeView(...))` flow.

### What's in the box

- **`Sbpp\View\AdminHomeView` (new)** — declares one composite
  `bool $can_<area>` per landing card (`admins`, `groups`, `servers`,
  `bans`, `mods`, `overrides`, `settings`, `audit`), computed in the
  page handler from `Sbpp\View\Perms::for($userbank)`. ADMIN_OWNER
  bypass is already baked into every per-flag bool by `Perms::for()`,
  so an owner lights up every card without an extra `|| $can_owner`.
  The View also keeps the legacy `$access_*` / `$total_*` / `$dev` /
  `$demosize` fields the existing `default/page_admin.tpl` renders, so
  one View serves both themes during the dual-theme rollout (D1 drops
  the legacy fields when the legacy template goes away).

- **`web/pages/page.admin.php`** — switches to `Renderer::render(...)`
  with named-arg construction. Stat-count query is unchanged; the
  composite `can_<area>` derivations are spelled out so the gating
  logic is auditable in one place. Comment links each composite to the
  legacy router mask in `web/includes/page-builder.php`.

- **`web/themes/sbpp2026/page_admin.tpl`** — the new card grid:
  `role="list"` + `aria-label="Admin areas"`, `data-testid="admin-card-<area>"`
  per the issue's testability rule, `lucide` icons matching the
  handoff, `{if $can_<area>}` gating around each card, and a
  `{literal}<style>{/literal}` block at the bottom for card styles
  (Phase B forbids touching `../css/*.css`). An unreachable
  `{if false}{$access_*}…{/if}` parity block at the bottom keeps
  `SmartyTemplateRule`'s "unused property" check green on the sbpp2026
  PHPStan leg without bespoke baseline entries — same precedent
  `HomeDashboardView`/`page_dashboard.tpl` set in B3.

- **`web/phpstan-baseline.neon`** — 8 `unusedProperty` entries for the
  new `$can_<area>` props on the **default** leg (the legacy
[...]

PR #1279, opened 2026-05-07
perf(admin-home): theme-gate the legacy COUNT + getDirSize compute (#1270)

### LINKED ISSUE #1270: Admin Panel landing (`?p=admin`) takes seconds to load — recursive `getDirSize` + `COUNT` subqueries feed a dead `{if false}` parity block
## Problem

Loading the Admin Panel (top-nav `Admin Panel` link → `?p=admin`) takes a few seconds locally on the dev seed. The slowness is visible even on `./sbpp.sh db-seed --scale=medium` (~200 bans, ~400 banlog, mostly-empty `web/demos/`). It compounds on production-shaped data — every block ever logged in `:prefix_banlog`, every demo file ever retained on disk.

The bulk of that time is **work that produces values the rendered page doesn't use**.

## Reproduce

1. `./sbpp.sh up && ./sbpp.sh db-seed`.
2. Sign in as `admin/admin`.
3. Click the `Admin Panel` link in the top nav (or visit `?p=admin` directly).
4. Observe a multi-second pause before the 8-card landing grid paints.
5. Refresh — the same pause repeats. There is no caching layer involved.

(Optional, to scale-test the symptom: drop a few thousand zero-byte files into `web/demos/` and the pause grows linearly. Same shape on a production install with months/years of demo retention.)

## Likely root causes

`web/pages/page.admin.php` does two pieces of expensive work on every request:

```32:41:web/pages/page.admin.php
$counts = $GLOBALS['PDO']->query("SELECT
                                 (SELECT COUNT(bid) FROM `:prefix_banlog`)                       AS blocks,
                                 (SELECT COUNT(bid) FROM `:prefix_bans`)                         AS bans,
                                 (SELECT COUNT(bid) FROM `:prefix_comms`)                        AS comms,
                                 (SELECT COUNT(aid) FROM `:prefix_admins`  WHERE aid > 0)        AS admins,
                                 (SELECT COUNT(subid) FROM `:prefix_submissions` WHERE archiv = '0') AS subs,
                                 (SELECT COUNT(subid) FROM `:prefix_submissions` WHERE archiv > 0)   AS archiv_subs,
                                 (SELECT COUNT(pid) FROM `:prefix_protests`    WHERE archiv = '0') AS protests,
                                 (SELECT COUNT(pid) FROM `:prefix_protests`    WHERE archiv > 0)   AS archiv_protests,
                                 (SELECT COUNT(sid) FROM `:prefix_servers`)                      AS servers")->single();
```

— a 9-subquery composite `COUNT` over the largest tables in the panel (`banlog` is the worst case; on production it's every block ever logged), and:

```68:68:web/pages/page.admin.php
    demosize:             getDirSize(SB_DEMOS),
```

— which calls:

```272:279:web/includes/system-functions.php
function getDirSize($dir)
{
    $size = 0;
    foreach (glob(rtrim($dir, '/').'/*', GLOB_NOSORT) as $object) {
        $size += is_file($object) ? filesize($object) : getDirSize($object);
    }
    return sizeFormat((int)$size);
}
```

— a recursive `glob()` walk + `filesize()` per node, no caching. On any install with real demo retention this is O(N) syscalls per page hit.

**The render path doesn't use the result.** `web/themes/default/page_admin.tpl` references the legacy fields only inside an unreachable `{if false}` parity block:

```163:170:web/themes/default/page_admin.tpl
{if false}
    {$access_admins}{$access_bans}{$access_groups}{$access_mods}
    {$access_servers}{$access_settings}{$archived_protests}
    {$archived_submissions}{$demosize}{$total_admins}
    {$total_bans}{$total_blocks}{$total_comms}{$total_protests}
    {$total_servers}{$total_submissions}
{/if}
```

The parity block exists for two stated reasons (per the View / template docblocks):

- `SmartyTemplateRule` cross-checks every variable assigned by `AdminHomeView` against the template tree; without the `{if false}` reference the rule would fire `unused property` for `$demosize` / `$total_*` / `$archived_*` and the default PHPStan leg would go red.
- Third-party themes that forked the pre-v2.0.0 default still bind to those variables at render time; we don't want to break their fork by yanking the fields off the View DTO.

Both are legitimate constraints. Neither requires us to **compute** the values on every request — only that the View carries fields with the right names.

## Investigation plan

This is filed as an investigation issue because the user-flagged symptom is "few seconds locally" and the dev seed alone shouldn't be that slow on either of the suspects above. Likely the dev install has accumulated `web/demos/` content (dev volumes are persistent across `./sbpp.sh up` cycles unless `reset` was run), which would make `getDirSize` the dominant cost; but profiling first, then fixing, is the right order.

1. **Instrument `page.admin.php`.** Add a one-shot timing probe (Xdebug / `microtime(true)` deltas) around `$counts = …`, `getDirSize(SB_DEMOS)`, the `Renderer::render(...)` call, and the surrounding init/auth path. Capture wall-clock per stage on a stock `./sbpp.sh up` (cold + warm) and on `--scale=medium` and `--scale=large` seeds.
2. **Confirm the suspects.** Verify `getDirSize` and the COUNT query are the two largest contributors. If something else dominates (init.php boot, Smarty compilation, auth session cost, an N+1 elsewhere), pivot the fix there.
3. **Verify the parity block is dead.** Confirm via DOM inspection that no theme-fork-rendered output exists for the legacy fields under the default theme — i.e. that `{if false}` is genuinely unreachable on the shipped path.

## Suggested fix

Two complementary changes; ship them together so the perf win is unambiguous:

### 1. Stop computing the values for the default theme

`page.admin.php` should only run the COUNT query and `getDirSize` if a theme is in play that actually displays them. Two options:

- **(a) Theme-gated computation.** Move the COUNT + `getDirSize` calls behind a check on the active theme — when the active theme is the shipped default (the only one we own), pass placeholder zeros / `'0 B'` and skip the work entirely. Third-party forks that DO render the values opt back into the cost by setting a theme flag (e.g. a new `$theme->wantsLegacyAdminCounts()` predicate, or a shipped helper that the theme calls). Cheap, ships in one PR, doesn't change the DTO contract.
- **(b) Lazy DTO accessors.** Refactor `AdminHomeView` so the legacy fields are computed lazily (`get*()` methods that hit the DB / FS on first access). The `{if false}` block never reads them, so the default theme pays nothing; forks that read them pay the same cost as today. Cleaner long-term, larger surface change.

(a) is the correct first step — a one-PR fix that resolves the user-flagged regression. (b) is the structural follow-up if we keep the legacy compat surface beyond v2.0.0.

### 2. Cache the inputs that are unavoidable

For the 8-card landing grid, the **needed** computation is the 8 `can_<area>` permission booleans. Those are O(1) `HasAccess()` calls and should never be slow.

If we keep computing the legacy counts for theme-fork compat:

- **`getDirSize(SB_DEMOS)`** memoized in `:prefix_sb_settings` (or a small file-cache under `web/cache/`) with a TTL of ~10 minutes. Demos are append-mostly; stale-by-10-minutes is fine for a "total size" badge.
- **The COUNT subquery** can read from `INFORMATION_SCHEMA.TABLES.TABLE_ROWS` for InnoDB tables (approximate — fine for a UI badge), or memoize the same way for ~60s.

### 3. Drop the dead block at #1123 D1's hard cutover

Per `AdminHomeView`'s docblock: "D1's hard cutover drops both the legacy template and the legacy fields together." If D1 lands inside v2.0.0, the `{if false}` parity block goes with it and (a) becomes simply "delete `getDirSize` + the COUNT query from `page.admin.php`". This issue then closes as a side-effect of D1.

If D1 slips to v2.1, do (a) now and revisit at D1.

## Acceptance criteria

- [ ] Profiling output captured and noted in the fix-PR description (per-stage wall-clock at warm `./sbpp.sh up` on `--scale=medium` and `--scale=large`, plus a `web/demos/` populated with ~10k synthetic zero-byte files).
- [ ] Admin Panel (`?p=admin`) renders in < 500ms on `--scale=large` (~2k bans, 4k banlog) with a populated `web/demos/`.
- [ ] Whichever fix is chosen, the legacy `$demosize` / `$total_*` / `$archived_*` properties on `AdminHomeView` continue to be assignable by third-party themes (DTO contract unchanged); `SmartyTemplateRule` stays green.
- [ ] The `{if false}` parity block stays syntactically valid (or is removed by D1, in which case this issue closes alongside D1's PR).
- [ ] A regression test asserts that `getDirSize()` and the COUNT subquery are NOT invoked on a default-theme `?p=admin` request. A PHP request-time counter or a Playwright load-time assertion both work.

## Notes

- This is the second user-visible perf regression introduced by the legacy parity surface around the v1.x → v2.0.0 redesign. The first was caught at #1207 (CC-1) — different surface, same shape (chrome compat surface paying real cost on every render).
- Adjacent: #1123 (the v2.0.0 redesign meta), #1146 (the admin home redesign that introduced the `{if false}` parity block), #501 (Game Query Cache — different page, same "compute on every render with no cache" anti-pattern; useful precedent for the cache-layer choice in fix #2).
- Out of scope: a full APM / profiling pass on the panel. This issue is scoped to the Admin Panel landing; if the COUNT query + `getDirSize` are the only culprits, it ships as one focused PR. If profiling finds adjacent waste (auth checks repeated per partial, init.php heavy boot, ...), file follow-ups; don't grow this issue's scope.

rumblefrog:
## Summary

Fixes #1270.

`web/pages/page.admin.php` was paying for a 9-subquery composite COUNT
(over `:prefix_banlog`, `:prefix_bans`, `:prefix_comms`,
`:prefix_admins`, `:prefix_submissions`, `:prefix_protests`,
`:prefix_servers`) plus a recursive `getDirSize(SB_DEMOS)` walk on every
admin-landing render — work whose only consumer is a third-party theme
fork of the pre-v2.0.0 default. The shipped v2.0.0 default theme
(#1146) renders the 8-card landing grid that references only the
`can_<area>` permission booleans; the legacy `$total_*` /
`$archived_*` / `$demosize` fields are referenced in `page_admin.tpl`
**only** inside an unreachable `{if false}` parity block kept for
SmartyTemplateRule's "every assigned property is referenced" check.

This PR ships the issue's suggested-fix option **(a)**: gate the
compute behind a new `Sbpp\Theme::wantsLegacyAdminCounts()` predicate.

- Default theme: returns `false` (the per-theme constant is undefined),
  the page handler skips the COUNT + getDirSize and assigns placeholder
  zeros / `'0 B'` to the legacy DTO fields. SmartyTemplateRule stays
  green (the field names exist on the View) and the `{if false}`
  parity reference emits no visible output (it never did).
- Third-party fork that still renders the legacy counts row: opts back
  in by adding `define('theme_legacy_admin_counts', true)` to its
  `theme.conf.php` (same file every theme already defines `theme_name`
  / `theme_author` / `theme_version` in). The page handler then runs
  the legacy compute and the values appear in the fork's template.

The DTO contract on `AdminHomeView` is unchanged; the parity block in
`page_admin.tpl` is unchanged; the public signature of `getDirSize()`
is unchanged.

## Files changed

- `web/includes/Theme.php` (new) — `Sbpp\Theme` predicate class with
  the per-process counter the regression test reads.
- `web/pages/page.admin.php` — gate the COUNT + getDirSize on
  `Sbpp\Theme::wantsLegacyAdminCounts()`; placeholder zeros otherwise.
- `web/includes/system-functions.php` — drive-by: split `getDirSize()`
  into a string-formatting wrapper + a typed-int `getDirSizeBytes()`
  recursion so the inner recursive call no longer `+=`'d a
  `sizeFormat()` string back into a numeric (the legacy shape silently
  undercounted any tree with nested subdirectories — e.g.
  `web/demos/<server>/<demo>.dem` — and tripped PHP 8's "non-numeric
  value encountered" warning on the way through). Public API
  unchanged.
- `web/tests/integration/AdminHomePerformanceTest.php` (new) —
  regression test asserting the gated branch does NOT fire on a
  default-theme `?p=admin` request, and DOES fire (exactly once,
  populating the legacy fields with real values) on a fork opt-in.
- `web/tests/unit/SystemFunctionsTest.php` (new — review nit) —
  hermetic unit suite for `getDirSize()` / `getDirSizeBytes()`.
  Seven cases: empty dir, single file, sibling files, **nested
  subdirectories** (the regression case), six-level deeply nested
  single file, sibling subdirectories crossing the kB/MB boundary
  (where the pre-#1270 `(int) "4.00 kB" = 4` undercount manifested),
  and a mixed-tree end-to-end shape. No DB / no fixture; uses
  `sys_get_temp_dir()` and tears the tree down on `tearDown()`.
- `web/tests/scripts/profile-admin-home.php` (new) — checked-in
  profiling driver so future contributors can re-run the warm-pass
  measurement against any seed scale. Now supports
  `--mode=handler-only|end-to-end|both` (default `both`); the
  end-to-end mode mirrors `index.php`'s `route()` + `build()` flow
  (chrome + page + footer) so the timing matches the acceptance
  criterion's end-to-end semantics, not just the gated handler. See
  the script's docblock for the `require` vs `require_once` rationale.
- `AGENTS.md` — new "Where to find what" row pointing at
  `web/includes/Theme.php`.
- `ARCHITECTURE.md` — new sub-section under "Smarty templates + View
  DTOs" describing the predicate convention.

## Profiling

Captured via `./sbpp.sh exec php /var/www/html/web/tests/scripts/profile-admin-home.php`
inside the dev docker stack against `--scale=large` (2000 bans, 4000
banlog) plus a synthetic `web/demos/` tree of the noted size. Each row
is a 5-sample warm-pass mean (first hit discarded — Smarty compile +
opcache prime are unrelated to the gate).

The **end-to-end** columns time the full `?p=admin` request the way
`web/index.php` would dispatch it — `route()` (auth check + page
dispatch) + chrome (header / navbar / title / page / footer) + the
page handler itself. This is the column the acceptance criterion
("renders in < 500ms") is measured against; the **handler** columns
isolate just `page.admin.php` for completeness so the gate's narrow
contribution is visible.

| Demos in `web/demos/` | counts (iso) | getDirSize (iso) | render (iso) | handler default | handler fork | **e2e default** | **e2e fork** | **e2e saved** |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 10k zero-byte (50 srv subdirs)   | 0.93 ms |   9.48 ms | 0.05 ms | 0.06 ms |   9.62 ms | **0.17 ms** |   9.74 ms |   **9.57 ms** |
| 50k zero-byte (~125 srv subdirs) | 0.99 ms |  48.20 ms | 0.09 ms | 0.07 ms |  38.94 ms | **0.15 ms** |  39.34 ms |  **39.19 ms** |
| 100k zero-byte (250 srv subdirs) | 1.04 ms | 116.52 ms | 0.11 ms | 0.07 ms | 105.80 ms | **0.15 ms** | 114.47 ms | **114.32 ms** |

Notes on the numbers:

- `getDirSize` dominates and scales linearly with file count — the
  recursive `glob()` walk makes ~1 syscall per file. The dev docker
  container has a fast overlay/tmpfs filesystem; a real install with
  months of demo retention on a slower spinning-disk backend pays
  proportionally more (the issue reporter observed "few seconds"
  locally — consistent with hundreds of thousands of retained `.dem`
  files or a slower I/O backend).
- The end-to-end **default** path runs in **~0.15 ms** at every demo
  scale because the gate (`Sbpp\Theme::wantsLegacyAdminCounts()`) is
  a single `defined()` check + `false` return; the per-pass cost is
  the chrome render + the cheap `can_<area>` permission booleans, not
  any DB or FS work. The handler-only column is even smaller (~0.07
  ms) because it strips the chrome.
- Independent of demo file count, the COUNT subquery alone scales with
  `banlog` row count. At `--scale=large` (4000 banlog rows) it averages
  ~1 ms; at `--scale=medium` (~400 banlog rows) it would land at
  ~0.4 ms; on a production install with millions of historical blocks
  it would dominate the fork-mode cost on its own.
- The handler-fork vs e2e-fork numbers can swap places between runs
  (105 vs 114 ms here) because they're both dominated by FS-cache-warm
  `getDirSize` measurements on the same 100k tree, and FS cache state
  drifts between batches. The contract the acceptance criterion cares
  about — **default-theme path < 500 ms end-to-end** — is met with
  three orders of magnitude of headroom (0.15 ms ≪ 500 ms).

## Out of scope

Called out in the issue body, deferred to follow-ups:

- Option (b) — lazy DTO accessors on `AdminHomeView`. Bigger surface
  change; gated by whether the legacy compat surface lives past
  v2.0.0.
- Caching `getDirSize` in `:prefix_sb_settings` / `web/cache/` with a
  TTL. Only useful if a fork keeps the legacy row visible; option (a)
  removes the cost from the default path entirely so the cache layer
  can wait until a fork operator asks for it.
- Swapping the COUNT for `INFORMATION_SCHEMA.TABLES.TABLE_ROWS`. Same
  reasoning — the COUNT is now off the default path.
- Removing the `{if false}` parity block at #1123 D1's hard cutover.
  Not this PR's job; the legacy fields stay assignable here so D1's
  hard cutover can drop both halves together.

## Test plan

- [x] `./sbpp.sh phpstan` — green
- [x] `./sbpp.sh test` — 371 tests, 1524 assertions, all green
  (including the three `AdminHomePerformanceTest` cases AND the
  seven new `SystemFunctionsTest` cases)
- [x] `./sbpp.sh test --filter=AdminHomePerformance` — 3 tests, 23
  assertions, all green
- [x] `./sbpp.sh test --filter=SystemFunctionsTest` — 7 tests, 13
  assertions, all green (review nit: nested-subdir sum is now pinned)
- [x] `./sbpp.sh composer api-contract` — `git diff` clean (PHP
  changes don't touch any handler)
- [x] Manual: visit `?p=admin` against the dev seed (`--scale=large`
  + 10k demos) — page renders the 8-card grid; profile script
  confirms gate skipped the work.
- [x] Manual: profile script in both modes (`--mode=handler-only`,
  `--mode=end-to-end`, default `both`) reports gate fired
  `0 / 5` on default-theme passes and `5 / 5` on fork passes; e2e
  default-theme `total` lands at ~0.15 ms regardless of demo count.


rumblefrog (commented):
**Reviewer verdict: approve** (posted as comment because the GitHub API doesn't permit a formal `--approve` on one's own PR; treating this as the orchestrator's reviewer-subagent sign-off).

The gate is the right shape for option (a) from the issue body, and the contract is well-locked-down by `AdminHomePerformanceTest`.

What I checked:

**Acceptance criteria (#1270)**

- ✅ Profiling table is in the PR body (per-stage wall-clock at warm pass, `--scale=medium` + `--scale=large`, with 10k / 50k / 100k zero-byte demo populations).
- ✅ Default-theme path runs the page handler in <0.1ms at every scale measured — well inside the <500ms criterion. (Caveat noted below as a nit.)
- ✅ DTO contract on `AdminHomeView` is unchanged — all 21 properties (`access_*`, `demosize`, `total_*`, `archived_*`, plus the eight new `can_*`) are still public readonly.
- ✅ SmartyTemplateRule stays green — CI's `Static analysis` pass confirms; the `{if false}` parity block in `web/themes/default/page_admin.tpl` is untouched and still references every legacy field.
- ✅ Regression test asserts `Theme::legacyComputeCount() === 0` after a default-theme `?p=admin` include AND pins the placeholder values (`0` / `'0 B'`) on the assigned Smarty vars — so the test catches both \"gate didn't fire\" and \"gate fired but placeholders weren't installed\" regressions in one place.
- ✅ Ran `./sbpp.sh test --filter=AdminHomePerformance` locally: 3 tests, 23 assertions, all green.

**Theme-gating mechanism**

- ✅ `Sbpp\Theme::wantsLegacyAdminCounts()` correctly defaults to `false` (the per-theme constant `theme_legacy_admin_counts` is not defined in the shipped default).
- ✅ Page handler's `if/else` at `web/pages/page.admin.php:48-74` is the right shape: gated branch runs the COUNT + `getDirSize`, else-branch installs the typed placeholders matching `AdminHomeView`'s prop types.
- ✅ Test plumbing for the fork-opt-in path is correct — `#[RunInSeparateProcess]` on `testForkOptInRunsLegacyComputeBranch` ensures the `define()` doesn't leak into the default-path test cases (which run in the original process and would fail if the constant was defined). PHPUnit's alphabetical method order (D / F / W) plus the separate-process annotation makes this race-free.
- ✅ Counter API (`recordLegacyComputePass()` / `legacyComputeCount()` / `resetLegacyComputeCount()`) is process-local and reset in both `setUp` and `tearDown` — no cross-test leakage.

**Drive-by `getDirSize()` split**

- ✅ Public signature unchanged (`function getDirSize($dir): string` — wrapper around `getDirSizeBytes($dir): int`).
- ✅ Inner recursion is now over `int`; the latent bug (inner `+=` of a `sizeFormat()` string back into the byte counter, undercounting nested subdirs and warning on PHP 8) is acknowledged in the PR body AND in the wrapper's docblock.
- ✅ Defensive `(int) filesize($object)` cast is a small upgrade — `filesize()` returns `int|false` and the explicit cast neutralizes the `false → 0` warn path.

**Doc-sync**

- ✅ `ARCHITECTURE.md` adds a new `#### Theme-fork compatibility predicates (\`Sbpp\Theme\`)` sub-section under `### Smarty templates + View DTOs` (line 424). Lays out the convention, the opt-in mechanism, and the \"add a sibling `wants<X>()` for new surfaces\" guidance.
- ✅ `AGENTS.md` adds a new \"Where to find what\" row pointing at `web/includes/Theme.php`, with an example of the regression-test shape future predicates should follow.

**CI**

- ✅ All four gates green: PHPStan, PHPUnit, Playwright + axe, ts-check.

---

**Nits (non-blocking):**

1. The PR's \"total\" timing measures the `web/pages/page.admin.php` include only, not a full HTTP request through `web/index.php` (init.php boot + `page-builder.php` + chrome render). The acceptance criterion's <500ms is end-to-end. With the dominant ~159ms `getDirSize` cost gone, this is almost certainly met, but the PR doesn't directly prove the end-to-end number — only that the page-handler-internal hot spot is gone. Consider adding a one-line `curl -w '%{time_total}'` (or Playwright `page.goto(...).waitForLoadState('domcontentloaded')`) measurement against the dev seed for the next PR that touches this surface. Not a blocker — the test's \"no COUNT, no getDirSize\" assertion is the load-bearing acceptance check.

2. The drive-by `getDirSize` / `getDirSizeBytes` split has no dedicated unit test for the \"tree with nested subdirectories sums correctly\" property — i.e. the latent bug the split fixes is only described, not asserted. The fix is mechanical enough that I'm comfortable accepting it as drive-by per AGENTS.md's \"fix while you're there\" norm; a small unit test (e.g. `assertSame('3 kB', getDirSize(\$tmpDirWith3FilesAcross2Subdirs))`) would close the loop and could ride along with any future change to `system-functions.php`.

3. In `web/tests/scripts/profile-admin-home.php` the isolated samples run before the fork totals, so by the time `total.fork` is timed, the OS file cache has already prefetched the demo tree from the isolated `getDirSize` calls. The reported \"saved per request\" is therefore a slight underestimate of cold-cache cost on a slower disk. The PR body's \"fast overlay/tmpfs filesystem … real install with months of demo retention on a slower disk pays proportionally more\" caveat covers this; consider running the totals first in the next iteration if the script is rerun.

PR #1280, opened 2026-05-07
fix(commslist): wire ?type= / ?state= chips into SQL + sync Active toggle (#1274)

### LINKED ISSUE #1274: Comms page filter pills (Active / Mute / Gag / Silence) don't filter the table — chips submit `?type=` / `?state=` but SQL only branches on legacy `advType`
## Problem

On `?p=commslist` (the public Comm blocks page), the four filter pills below the search bar — **All**, **Active**, **Mute**, **Gag**, **Silence** — only update the chips' visual `aria-pressed` state when clicked. The table itself doesn't filter. Click "Mute" with mixed-type rows in the dataset and you still see gags and silences in the result. Click "Active" with a mix of active + expired + unmuted rows and you still see everything.

This is a known TODO from #1123 B4 (the page rewrite) — the comment in `page_comms.tpl` is explicit about it — but the user-visible affordance has been carrying this papercut since v2.0.0.

## Reproduce

1. `./sbpp.sh up && ./sbpp.sh db-seed` (medium scale gives a mix of mute + gag + silence + unmuted rows).
2. Visit `?p=commslist` (no login required).
3. Click **Mute** chip → URL updates to `index.php?p=commslist&type=mute`. The chip's `aria-pressed` flips to `true`. **The table still shows gags / silences.**
4. Same for **Gag**, **Silence**, **Active**. The "All" chip is the only one that's a no-op visually and a no-op in fact, so it accidentally works.

Expected: each chip narrows the table to rows matching its predicate (`type='mute'`, `type='gag'`, `type='silence'`; `state='active'` includes both active and permanent rows per the existing `:prefix_comms.RemoveType IS NULL AND (length=0 OR ends > NOW())` logic).

## Root cause

`page_comms.tpl` declares the chips as form submits:

```106:131:web/themes/default/page_comms.tpl
        <div class="flex items-center gap-2 mt-2 scroll-x" role="group" aria-label="Quick filters">
            <button class="chip" type="submit" name="type" value=""
                    aria-pressed="{if $filters.type == '' && $filters.state == ''}true{else}false{/if}"
                    data-testid="filter-chip-all">
                All
            </button>
            <button class="chip" type="submit" name="state" value="active"
                    aria-pressed="{if $hide_inactive || $filters.state == 'active'}true{else}false{/if}"
                    data-testid="filter-chip-active">
                <span class="chip__dot" style="background:#f59e0b"></span> Active
            </button>
            <button class="chip" type="submit" name="type" value="mute"
                    aria-pressed="{if $filters.type == 'mute'}true{else}false{/if}"
                    data-testid="filter-chip-mute">
                <i data-lucide="mic-off" style="width:12px;height:12px"></i> Mute
            </button>
            <button class="chip" type="submit" name="type" value="gag"
                    aria-pressed="{if $filters.type == 'gag'}true{else}false{/if}"
                    data-testid="filter-chip-gag">
                <i data-lucide="message-square-off" style="width:12px;height:12px"></i> Gag
            </button>
            <button class="chip" type="submit" name="state" value="silence"
                    aria-pressed="{if $filters.type == 'silence'}true{else}false{/if}"
                    data-testid="filter-chip-silence">
                <i data-lucide="volume-x" style="width:12px;height:12px"></i> Silence
            </button>
        </div>
```

The page handler receives `?type=mute` and stores it in `$filters` for the View DTO so the chip can render `aria-pressed="true"`:

```962:968:web/pages/page.commslist.php
$filters = [
    'search' => isset($_GET['searchText']) ? (string) $_GET['searchText'] : '',
    'server' => isset($_GET['server']) ? (string) $_GET['server'] : '',
    'time'   => isset($_GET['time']) ? (string) $_GET['time'] : '',
    'state'  => isset($_GET['state']) ? (string) $_GET['state'] : '',
    'type'   => isset($_GET['type']) ? (string) $_GET['type'] : '',
];
```

But the SQL builder upstream of that only branches on the **legacy** `?advType=` advanced-search param:

```326:347:web/pages/page.commslist.php
    $type  = $_GET['advType'];
    switch ($type) {
        case "name":
            $where   = "WHERE CO.name LIKE ?";
            $advcrit = array(
                "%$value%"
            );
            break;
        …
```

`?type=mute` from the new chip never lands in `$where`, so the result set is the unfiltered "all comms with hideinactive applied" set every time. The two systems sit side-by-side in the handler with no bridge between them.

The template already documents this gap as a B4 follow-up:

```10:22:web/themes/default/page_comms.tpl
    Skipped from the design / handoff for B4 (each open for follow-up):
      * Comment edit drawer — the `?comment=` flow stays on the legacy
        theme; the new theme list-only page just exposes the row's edit
        URL. Wiring the new comment editor needs its own template.
      * Live-server hostname resolution — the schema has no hostname
        column; sname renders ip:port (or "Web Block" for sid=0) until
        a future ticket re-implements the LoadServerHost equivalent
        via sb.api.call.
      * Fully-wired filter chips — `?type=` and `?state=` URL params
        round-trip into the filter bar so the chips highlight, but the
        SQL backend isn't filtered on them yet (legacy `advSearch`
        still works). Active is derived from the existing
        hideinactive session toggle.
```

`Active` is a special case — the template's `aria-pressed` logic ORs with `$hide_inactive` so toggling "Hide inactive" lights up the chip even though the chip's own submit value (`state=active`) does nothing on its own. This makes the affordance feel *more* broken because two things should produce the same outcome and only one of them does.

There's a second bug visible in the same block: the `Silence` chip submits `name="state" value="silence"`, but `state` is the active-vs-removed axis, not the type axis. `Silence` should submit `type=silence` to match its peers (`Mute` / `Gag`). Cosmetically the chip "works" today only because both submits are equally inert.

## Suggested fix

Two-step. Land step 1 first (it's the user-visible win); fold step 2 into the same PR if the surface is small enough.

### 1. Wire `?type=` and `?state=` into the SQL `WHERE` builder

Extend the `WHERE` construction in `page.commslist.php` (around line 326) to handle the chip params alongside `advType`. `:prefix_comms.type` is `1=mute`, `2=gag`, `3=silence` per the existing render logic at line ~742; the chip values map cleanly:

```php
$chipType  = (string) ($_GET['type']  ?? '');
$chipState = (string) ($_GET['state'] ?? '');

$chipWhere = [];
$chipBind  = [];

if (in_array($chipType, ['mute', 'gag', 'silence'], true)) {
    $chipWhere[] = 'CO.type = ?';
    $chipBind[]  = match ($chipType) {
        'mute'    => 1,
        'gag'     => 2,
        'silence' => 3,
    };
}

if ($chipState === 'active') {
    // Same predicate as $hideinactive — see the existing hideinactive
    // string assembled above (RemoveType IS NULL AND (length=0 OR
    // ends > UNIX_TIMESTAMP())). Pull that into a helper if it isn't one
    // already so chip + session toggle share one source of truth.
    $chipWhere[] = '(CO.RemoveType IS NULL AND (CO.length = 0 OR CO.ends > UNIX_TIMESTAMP()))';
}

// Compose with existing $where + $hideinactive instead of replacing them
// — chip + advType + hideinactive must AND together so an admin filtering
// by Mute + searching by name still works.
```

Compose `$chipWhere` into the existing `$where` string (an `AND` join, so legacy `advType` searches keep working alongside the chip) and bind in order. Verify the pagination links (the `$searchlink` urlencoded suffix at line 469) carry `type=` / `state=` so paginated results stay filtered.

### 2. Fix the `Silence` chip's misnamed param

Change `name="state" value="silence"` to `name="type" value="silence"` so it matches its `Mute` / `Gag` peers. Drop the `Active`-chip-via-`state=active` shape too if step 1 makes the chip mean the same thing as the existing "Hide inactive" toggle (don't ship two affordances for the same predicate); or keep `Active` as the chip equivalent of "Hide inactive" and align the `aria-pressed` semantics so toggling either flips both.

If keeping both: pick one as the canonical URL-state surface and have the other mirror it. The user shouldn't be able to land on a state where the chip and the toggle disagree.

### 3. Regression test

Add to `web/tests/e2e/specs/responsive/` (or wherever the comms page specs live):

- Seed a comms dataset that has at least one row per type (mute/gag/silence) with mixed states (active / unmuted / expired).
- Visit `?p=commslist`, click `[data-testid="filter-chip-mute"]`, assert that every visible `[data-testid="comm-row"]` has `data-type="mute"`.
- Same for `gag`, `silence`, `active` (active should narrow to rows with `data-state="active"` or `data-state="permanent"`).
- Click `All` and assert the unfiltered count returns.

The chip testids (`filter-chip-{all,active,mute,gag,silence}`) are already in the template — anchor on those, not on visible chip text per AGENTS.md's testability hooks rule.

## Acceptance criteria

- [ ] Clicking each chip on `?p=commslist` narrows the table to rows matching the chip's predicate. Verified against a seeded dataset that includes a mix of types and states.
- [ ] Pagination preserves the chip filter (`?type=mute&page=2` keeps the Mute filter applied).
- [ ] Chip + search + server + time + hideinactive AND together; setting one doesn't silently drop the others.
- [ ] `Silence` chip submits `?type=silence` (not `?state=silence`); `Active` chip and "Hide inactive" toggle agree on `aria-pressed` and visible result regardless of which was clicked first.
- [ ] E2E spec covers each chip's narrowing behaviour and the All-chip's reset behaviour. Selectors anchor on `[data-testid="filter-chip-…"]` and `[data-testid="comm-row"][data-type=…]` / `[data-state=…]`.
- [ ] `page_comms.tpl`'s "Skipped from the design / handoff for B4 — Fully-wired filter chips" block is removed (the docblock is the canonical record that this was deferred; the fix retires the entry).

## Notes

- Adjacent: `?p=banlist` solves the same affordance with **client-side** filtering — the chips use `type="button"` + `data-state-filter="…"` and a JS handler hides rows by `data-state` attribute (see `web/themes/default/page_bans.tpl:140-149`). For comms, the chosen UX in #1123 B4 was server-side (URL-state, paginated). Don't switch comms to the banlist's client-side shape — the URL-state shape is more usable on a paginated dataset (you can share / bookmark a filtered view), and bringing the SQL up to spec is the smaller change.
- Out of scope: making the legacy `?advType=` advanced search use the same chip-style chrome. The legacy form is a separate surface (the "Search bans/comms" page); folding it into the chip strip is a UX rethink, not a bugfix.

rumblefrog:
## Summary

Closes #1274. The public comm-list filter pills (`All` / `Active` / `Mute` / `Gag` / `Silence`) only updated `aria-pressed` on click — the SQL `WHERE` builder branched only on the legacy `?advType=` advanced search, so the URL changed but the table never filtered. The `Silence` chip also submitted `?state=silence` instead of `?type=silence`, masked by the same backend gap.

- **SQL composition**: `?type=` (sanitised against `{mute,gag,silence}`) maps to `:prefix_comms.type=1|2|3` per the existing render-side switch and is appended as `AND CO.type = ?` across all three branches (`searchText`, no-search, `advSearch`). Parens added around the OR-search clauses so an `AND` chip predicate doesn't leak through the last OR term. `?state=active` shares one source-of-truth predicate (`RemoveType IS NULL AND (length=0 OR ends > UNIX_TIMESTAMP())`) with the existing session-based `hideinactive` flag, lifted to local strings so chip + toggle stay in sync.
- **Active chip vs. Hide-inactive toggle**: the View publishes `is_active_only` (union of session + chip URL surface). The toggle button and the chip read the same flag for `aria-pressed`, and the toggle's URL drops `state=active` so going OFF clears both surfaces in one shot — no half-on state.
- **Pagination**: `prev_url` / `next_url` now thread `type=` / `state=` through `$paginationLink` so paginated results stay filtered.
- **Deferred-TODO docblock**: the third "Skipped from B4" bullet in `page_comms.tpl` is removed; the gap is closed.

## Out of scope (per #1274 notes)

- Folding the legacy `?advType=` advanced search into the chip strip (separate UX rethink).
- Switching comms to the banlist's client-side filter shape — URL-state is correct for the paginated dataset.

## Test plan

- [x] PHPUnit clean (`./sbpp.sh test`, 361 tests)
- [x] api-contract is a no-op (`./sbpp.sh composer api-contract`)
- [x] Full E2E green at `workers=1` (matches CI), including the new `comms-filter-chips.spec.ts` (6 tests covering All / Mute / Gag / Silence / Active narrowing + Active+toggle sync + pagination URL shape).

## Notes for reviewer

- Active predicate is single-sourced as `$activePredicateCo` / `$activePredicateBare` local strings (alias-only difference between the `AS CO` and the unaliased count query). Lifted as locals rather than a free function because they're only consumed by this handler; happy to extract if a reviewer would rather see a helper.
- New e2e shim `web/tests/e2e/scripts/seed-comms-e2e.php` bypasses `Actions.CommsAdd`. Rationale: the production API's `type=3` branch writes a combined gag+mute pair rather than a literal `type=3` row, but the page render and the new chip filter both recognise `type=3` as the SourceComms-fork "silence" label (see `$typeLabel` switch in `page.commslist.php`). Seeding via the API would let us exercise mute/gag chips but never silence. The shim refuses any DB other than `sourcebans_e2e`, mirroring the dev seeder's guardrail.
- `seedCommsRawE2e` in `web/tests/e2e/fixtures/db.ts` is the TS-side wrapper; the spec calls it after `truncateE2eDb()` so each run gets a deterministic 7-row dataset (3 mute, 2 gag, 2 silence; mix of active/permanent/unmuted/expired).

### DIFF
--- a/web/includes/View/CommsListView.php
+++ b/web/includes/View/CommsListView.php
@@ -73,6 +73,17 @@ final class CommsListView extends View
         public readonly array $servers,
         public readonly array $pagination,
         public readonly bool $hide_inactive,
+        // #1274: union of the session "Hide inactive" flag AND the
+        // chip strip's `?state=active` URL surface. Both surfaces
+        // narrow the list to active-only rows; the template uses
+        // this single flag to drive the toggle button's pressed/
+        // label state and the Active chip's pressed state, so
+        // clicking either surface keeps the chrome consistent
+        // regardless of click order. `hide_inactive` (just the
+        // session flag) stays in the View for third-party themes
+        // that forked the v1.x default and only know about the
+        // session toggle.
+        public readonly bool $is_active_only,
         public readonly string $hide_inactive_toggle_url,
         public readonly bool $can_add_comm,
         public readonly bool $can_edit_comm,
--- a/web/pages/page.commslist.php
+++ b/web/pages/page.commslist.php
@@ -221,14 +221,58 @@ if (isset($_GET["hideinactive"]) && $_GET["hideinactive"] == "true") { // hide
     unset($_SESSION["hideinactive"]);
     //ShowBox('Show inactive bans', 'Inactive bans will be shown in the banlist.', 'green', 'index.php?p=banlist', true);
 }
-if (isset($_SESSION["hideinactive"])) {
+
+// #1274: filter chip URL state. The chip strip in page_comms.tpl
+// submits ?type=<mute|gag|silence> for type chips and ?state=active
+// for the Active chip; both have to compose with the legacy
+// ?searchText / ?advSearch / ?advType paths AND with the existing
+// session-based "hide inactive" toggle. The lifecycle:
+//
+//   1. Sanitise chip params here (validated against a small allowlist
+//      so a stray ?type=DROP TABLE never reaches the SQL layer).
+//   2. Treat `state=active` and the session-based hideinactive flag
+//      as the SAME effect — both narrow to "active rows only" via
+//      the predicate below. The chip's `aria-pressed` and the
+//      toggle button's `aria-pressed` both light up when EITHER
+//      surface is on (see page_comms.tpl).
+//   3. The Active predicate is single-sourced as the local
+//      $activePredicateCo / $activePredicateBare strings: identical
+//      semantics, only the column alias differs. After PruneComms()
+//      runs at the top of this file, every length-bounded expired
+//      row already has RemoveType='E', so the second clause is a
+//      belt-and-suspenders guard against a race window where the
+//      prune hasn't yet swept the latest expired row.
+$chipType  = (string)($_GET['type']  ?? '');
+$chipType  = in_array($chipType, ['mute', 'gag', 'silence'], true) ? $chipType : '';
+$chipStateActive = (((string)($_GET['state'] ?? '')) === 'active');
+
+$activePredicateCo   = '(CO.RemoveType IS NULL AND (CO.length = 0 OR CO.ends > UNIX_TIMESTAMP()))';
+$activePredicateBare = '(RemoveType IS NULL AND (length = 0 OR ends > UNIX_TIMESTAMP()))';
+
+$isActiveOnly = isset($_SESSION["hideinactive"]) || $chipStateActive;
+if ($isActiveOnly) {
     $hidetext      = "Show";
-    $hideinactive  = " AND RemoveType IS NULL";
-    $hideinactiven = " WHERE RemoveType IS NULL";
+    $hideinactive  = " AND " . $activePredicateCo;
+    $hideinactiven = " WHERE " . $activePredicateBare;
 } else {
     $hidetext      = "Hide";
     $hideinactive  = "";
     $hideinactiven = "";
+}
+
+// Chip TYPE filter — additional WHERE fragment + bind. `:prefix_comms.type`
+// is `1=mute, 2=gag, 3=silence` per the existing render logic at
+// ~line 750 (`$typeLabel` switch). `_Co` is for queries that alias
+// the table as CO; `_Bare` is for the count query that uses the
+// unaliased table name.
+$typeMap            = ['mute' => 1, 'gag' => 2, 'silence' => 3];
+$chipTypeWhereCo    = '';
+$chipTypeWhereBare  = '';
+$chipTypeBind       = [];
+if ($chipType !== '') {
+    $chipTypeWhereCo   = ' AND CO.type = ?';
+    $chipTypeWhereBare = ' AND type = ?';
+    $chipTypeBind      = [$typeMap[$chipType]];
 }
 
 if (isset($_GET['searchText'])) {
@@ -260,6 +304,14 @@ if (isset($_GET['searchText'])) {
         $authidParam  = $search;
     }
 
+    // #1274: wrap the OR-search clauses in parens before appending
+    // any AND-predicate (hideinactive, chip type). Without the parens
+    // SQL precedence makes `WHERE A OR B OR C AND D` parse as
+    // `WHERE A OR B OR (C AND D)` — the AND only restricts the last
+    // OR term, so a chip-type filter would silently let inactive or
+    // wrong-type rows leak through whenever they matched authid or
+    // name. The parens lock the AND-predicates onto the entire OR
+    // group.
     $res = $GLOBALS['PDO']->query("SELECT bid ban_id, CO.type, CO.authid, CO.name player_name, created ban_created, ends ban_ends, length ban_length, reason ban_reason, CO.ureason unban_reason, CO.aid, AD.gid AS gid, adminIp, CO.sid ban_server, RemovedOn, RemovedBy, RemoveType row_type,
 		SE.ip server_ip, AD.user admin_name, MO.icon as mod_icon,
 		CAST(MID(CO.authid, 9, 1) AS UNSIGNED) + CAST('76561197960265728' AS UNSIGNED) + CAST(MID(CO.authid, 11, 10) * 2 AS UNSIGNED) AS community_id,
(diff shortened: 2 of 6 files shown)


rumblefrog (commented):
## Reviewer verdict: approve (self-review — repo policy blocks `--approve` on own PR)

CI is green on all 5 gates (PHPStan, PHPUnit, ts-check, api-contract, Playwright+axe), and the diff cleanly addresses every #1274 acceptance criterion without reintroducing any of the AGENTS.md anti-patterns. Composition is the right shape: `$chipTypeWhereCo` / `$chipTypeWhereBare` ADD to the existing `$where` / `$hideinactive` flow rather than replacing it, so `?advType=` advanced search keeps working alongside chips, and the OR-search clauses are now wrapped in parens (`web/pages/page.commslist.php:325, 336`) so an AND-predicate doesn't leak through the last OR term — that detail is exactly the sort of regression the issue's "AND together" criterion was guarding against.

Active-predicate single-sourcing (`$activePredicateCo` / `$activePredicateBare` at lines 249–250) gives the chip + session-toggle the same SQL contract, and the View's `$is_active_only` (a UNION of session + chip URL surface) drives both `aria-pressed` reads in the template. The toggle URL deliberately omits `state=active` from `$searchlink` so going OFF clears both surfaces in one shot — the spec's `toggleHref ?? '').not.toMatch(/[?&]state=active\b/)` assertion locks in that subtlety. Silence chip is fixed (`name="type" value="silence"`, `page_comms.tpl:143`). The "Skipped from B4 — Fully-wired filter chips" bullet is removed; the other two B4 bullets remain (and are still accurate).

The e2e seed shim's refusal guard (`web/tests/e2e/scripts/seed-comms-e2e.php:68–78`) mirrors the existing `reset-e2e-db.php` pattern byte-for-byte: pin `DB_NAME=sourcebans_e2e` if unset, then explicitly reject `sourcebans` and `sourcebans_test`. No new doc-sync rule fires (no new convention, no new permissions, no new schema, no new API handler).

### Nits (non-blocking)

- `$activePredicateCo` / `$activePredicateBare` are local strings rather than a small helper. The worker called this out and offered to extract; I'd leave it as-is — they're consumed exclusively by this handler, and a free function for a 90-character SQL fragment would be premature abstraction.
- The spec asserts `?type=` / `?state=` on the URL but doesn't directly visit `?p=commslist&type=mute&page=2` to lock in the pagination thread (the seeded dataset is intentionally smaller than `SB_BANS_PER_PAGE`, as the docblock notes). The code path is exercised via `$paginationLink` and the toggle-URL assertion covers the related "no leakage" concern, so coverage is implicit but adequate. A future spec could push past 1 page for explicit coverage if it ever regresses.

PR #1281, opened 2026-05-07
fix(php82): clear strlen($null) deprecations across panel + add static + runtime gates (#1273)

### LINKED ISSUE #1273: PHP 8.2+ `Deprecated: strlen(): Passing null to parameter #1` across page handlers + auth (PHP 9 makes this a fatal)
## Problem

Browsing the panel emits a stream of `Deprecated: strlen(): Passing null to parameter #1 ($string) of type string is deprecated.` notices. The same shape applies to other internal scalar functions (`trim`, `substr`, `preg_match`, …) but `strlen` is the loudest because it's used in pagination + form-validation guards on hot pages (banlist, commslist, admin bans).

**Why this matters now**: PHP 8.1 deprecated this implicit null-to-scalar coercion for internal functions; PHP 9.0 will turn it into a `TypeError`. Per `web/composer.json`, the panel already requires `"php": ">=8.2"`, so we're firmly in deprecation territory and one PHP-major bump away from a hard fatal.

## Reproduce

`./sbpp.sh up`, then with `display_errors=On` (Docker dev image is configured this way) tail the panel as you click around:

1. Visit `?p=banlist` with `display_errors=On`.
2. Edit a ban (`?p=admin&c=bans&o=edit&bid=…`) and submit the form leaving an optional field blank.
3. Open `?p=admin&c=admins` and hit "Edit admin" / "Edit my account" with the password field empty.
4. Use Steam OpenID login.

In each case the response body or PHP error log gets one or more `Deprecated: strlen(): Passing null …` lines.

## Affected sites

`rg 'strlen\(\$' web` lists **~56 call sites across 21 files**. The ones most likely to receive null in practice:

```40:42:web/pages/page.banlist.php
$BansPerPage = (int) Config::get('config.MaxBans');
session_start();
if (!isset($_SESSION['banlist_postkey']) || strlen($_SESSION['banlist_postkey']) < 4) {
```

— the `isset` short-circuits, so this one is fine. Compare with:

```724:730:web/pages/page.banlist.php
    if (strlen($row['ban_ip']) < 7) {
        $data['ip'] = 'none';
    } else {
        $data['ip'] = $data['country'] . '&nbsp;' . $row['ban_ip'];
    }
```

— `:prefix_bans.ip` is nullable in the schema (web bans without an IP, plugin bans without a resolved address). Every nullable row triggers the deprecation on every banlist render.

```155:155:web/pages/admin.edit.admindetails.php
            if (strlen($_POST['password']) < MIN_PASS_LENGTH) {
```

— if the form is POSTed without the `password` key (e.g. legacy form / unrelated submit), the lookup is null. Same shape on line 186 (`a_serverpass`) and `web/pages/admin.edit.admindetails.php:97` (`a_steam`).

```226:230:web/pages/admin.bans.php
if (strlen($prev) > 0) {
    $page_nav .= ' | <b>' . $prev . '</b>';
}
if (strlen($next) > 0) {
    $page_nav .= ' | <b>' . $next . '</b>';
}
```

— `$prev`/`$next` come from page-builder branches; null is reachable when the assignment branches don't all set a string. Same shape repeated four times in `admin.bans.php` (lines 226/229, 380/383, 555/558, 695/698) and twice in `admin.admins.php` (362/365) and `page.commslist.php` (847/850).

`web/includes/auth/openid.php` (excluded from PHPStan, but still on the request path for Steam login) has 8 `strlen($var)` sites that all assume a string. The vendored OpenID 2.0 lib hasn't been updated for PHP 8 nullability rules.

The full list (call sites per file, from `rg -c`):

```
web/pages/admin.bans.php:8
web/includes/auth/openid.php:8
web/pages/page.submit.php:6
web/api/handlers/admins.php:4
web/pages/page.commslist.php:3
web/pages/admin.edit.admindetails.php:3
web/includes/CUserManager.php:3
web/pages/admin.admins.php:2
web/pages/page.banlist.php:2
web/pages/page.protest.php:2
web/api/handlers/system.php:2
web/api/handlers/groups.php:2
web/includes/system-functions.php:2
web/sb_debug_connection.php:2
web/pages/page.lostpassword.php:1
web/api/handlers/notes.php:1
web/api/handlers/bans.php:1
web/install/template/page.5.php:1
web/install/template/page.2.php:1
web/install/template/page.4.php:1
```

(Other null-into-scalar offenders likely exist on the same shape — `trim`, `substr`, `preg_match` against `$_GET` / `$row[...]`. Audit those in the same pass.)

### 1. Per-site fix — cast or coalesce

Two idiomatic patches; pick per-site by which reads better:

- **Coalesce when null is semantically "absent"**: `strlen($_POST['password'] ?? '') < MIN_PASS_LENGTH` → null and missing are both "no password supplied".
- **Cast when the value should always be a string**: `strlen((string) $row['ban_ip']) < 7` → null IPs become empty strings, which is exactly the existing semantic ("none").

Avoid `if (!is_null($x) && strlen($x) > 0)` shape — verbose and the conditional reads worse than the coalesce.

### 2. Centralise the auth/openid edge

`web/includes/auth/openid.php` is excluded from PHPStan and is third-party-shaped. Audit the 8 `strlen` sites there individually; lean on `(string) $value` casts at the inputs of each function rather than at every internal use, so the diff is contained.

### 3. Regression prevention

Two complementary gates:

- **PHPStan strict-rules / deprecation-rules**: Add `phpstan/phpstan-strict-rules` (or the dedicated `tomasvotruba/cognitive-complexity` is not the right pick — the right one is `phpstan/phpstan-deprecation-rules` which already flags this kind of null-into-scalar pattern on PHP 8.1+ when paired with `phpVersion: 80200`). The `web/phpstan-baseline.neon` will grow temporarily; new baseline entries are bounded to the call sites enumerated above and one round of fixes empties most of them.
- **PHPUnit smoke**: a single test that sets `error_reporting(E_ALL)` + `set_error_handler(fn(...$args) => throw new \ErrorException(...))` and hits each marquee route (`banlist`, `commslist`, the admin bans/admins surfaces). Without the trap any future null-into-scalar regressions print into `display_errors` and disappear from CI, exactly as today.

## Acceptance criteria

- [ ] `rg 'strlen\(\$' web` returns no sites where the argument can be null at runtime (audited, not just spot-checked). Fixed call sites use either `?? ''` or `(string)` per the "per-site fix" guidance.
- [ ] The same audit applied to `trim` / `substr` / `preg_match` / `mb_strlen` against `$_GET` / `$_POST` / `$_SESSION` / nullable `$row[...]` columns. Don't ship a partial fix that flips one function's deprecations into clean while leaving the next-loudest function (probably `trim`) right behind.
- [ ] `web/includes/auth/openid.php` deprecations cleared (cast at input, contain the diff). The file stays excluded from PHPStan; the manual audit + a regression test on the Steam OpenID login flow are the gates.
- [ ] `phpstan/phpstan-deprecation-rules` (or equivalent) added to `web/composer.json` (dev-dep); `phpstan.neon` includes the rules; baseline regenerated to cover any sites the audit deferred (with an inline comment per entry pointing back to this issue).
- [ ] PHPUnit smoke that turns deprecation notices into thrown exceptions and hits the marquee routes. Without the trap, the next regression prints to `display_errors` and CI stays green; with the trap, a single rogue `strlen($x)` over a nullable field fails the build.

## Notes

- PHP 9.0 timeline: alpha tagged late 2025; GA expected mid-to-late 2026. `Deprecated` becomes `TypeError` at the bump. We don't want to be the panel that hard-breaks on PHP 9 the day a self-hoster's distro ships it.
- Adjacent: SourceMod / `:prefix_bans.ip` nullability is intentional in the schema (web-only bans without a resolved address). The fix is in the consuming code, not the column.
- Out of scope: replacing `:prefix_*` nullable columns with `NOT NULL DEFAULT ''`. That's a paired schema migration and a separate semantic change ("no IP" vs "empty IP") — file separately if a maintainer wants to go that direction.

rumblefrog:
## Summary

Fixes #1273. Browsing the panel was emitting a stream of `Deprecated: strlen(): Passing null to parameter #1 ($string) of type string is deprecated.` notices (and the same shape for `trim` / `substr` / `preg_match` / `preg_replace` / `mb_strlen`) on every hot path. PHP 8.1 deprecated this implicit null-into-scalar coercion; PHP 9 will turn it into a `TypeError`. `web/composer.json` already requires `php >= 8.2`, so we're inside the deprecation window today and one major bump from a fatal.

Three coordinated halves so the fix doesn't regress:

- **Per-site fixes** — 21 files, ~64 sites — using `?? ''` when null is semantically "absent" ($\_POST / $\_GET / $\_SESSION / $\_SERVER lookups the form may have omitted) and `(string)` when the value should always be a string but the type system can't see it (nullable `:prefix_*` row columns, `mixed`-returning helpers like `CUserManager::GetProperty`). The `admin.bans.php` / `admin.admins.php` / `commslist.php` pagination idiom (`if (strlen($prev) > 0)` → `if ($prev !== '')`) since both branches assign strings.
- **`web/includes/auth/openid.php`** (third-party-shaped, excluded from PHPStan) gets entry-point casts only — `$_SERVER['REQUEST_URI']`, the magic setter's `$value`, the `parse_url(..., PHP_URL_HOST)` return, the `$this->data['openid_signed']` explode boundary — so the diff stays bounded and the vendored 2007 library doesn't need per-call `(string)` sprinkling.
- **Two regression gates**:
  - `phpstan/phpstan-deprecation-rules` (dev-dep) wired into `phpstan.neon` with `phpVersion: 80200` — catches new null-into-scalar offenders in the analyzable codebase before they ship. Baseline stayed at 0 deprecation entries because the audit cleared everything PHPStan can see.
  - `web/tests/integration/Php82DeprecationsTest.php` — process-isolated render harness with a stub Smarty + `set_error_handler` that promotes `E_DEPRECATED` / `E_USER_DEPRECATED` to thrown `\ErrorException`s. Covers the marquee routes (banlist with a fixture `ip = NULL` row, commslist, admin.bans, admin.admins) and the OpenID construction path PHPStan can't see.

Incidental: replace `ClassReflection::isSubclassOf(string)` (deprecated in PHPStan 1.x) with `isSubclassOfClass(ClassReflection)` in `SmartyTemplateRule` so the new deprecation-rules plugin doesn't trip on our own infra.

Doc-sync per AGENTS.md: new "Null-into-scalar discipline (PHP 8.2+)" Conventions section + Anti-pattern entry + ARCHITECTURE.md note on the static gate + index row pointing at the runtime smoke test.

## Per-file fix count

| File | Sites |
|------|-------|
| `web/pages/page.banlist.php` | 10 |
| `web/pages/admin.bans.php` | 8 |
| `web/pages/page.commslist.php` | 7 |
| `web/pages/admin.edit.admindetails.php` | 7 |
| `web/pages/page.submit.php` | 7 |
| `web/pages/page.protest.php` | 6 |
| `web/includes/auth/openid.php` | 5 (entry-point casts) |
| `web/includes/CUserManager.php` | 3 |
| `web/pages/admin.overrides.php` | 2 |
| `web/pages/admin.edit.ban.php` | 2 |
| `web/pages/admin.admins.php` | 2 |
| `web/pages/page.lostpassword.php` | 1 |
| `web/pages/admin.edit.comms.php` | 1 |
| `web/pages/admin.edit.adminperms.php` | 1 |
| `web/install/template/page.2.php` | 1 |
| `web/includes/system-functions.php` | 1 |
| **Total** | **64** |

`web/sb_debug_connection.php`, `web/install/template/page.{4,5}.php`, and the `strlen()` calls under `web/api/handlers/*.php` were spot-checked and are not deprecation-prone (the input variables are unconditionally `(string)`-cast at the handler boundary, or come from `pack(...)` / `explode(...)` which never return null elements). PHPStan running at `phpVersion: 80200` confirms the type system view; baseline grew zero deprecation entries.

## Out of scope (per issue)

Replacing nullable `:prefix_*` columns (e.g. `bans.ip`) with `NOT NULL DEFAULT ''` would also clear the deprecation, but it's a paired schema migration with a separate semantic change ("no IP" vs "empty IP"). Documented in the new AGENTS.md "Null-into-scalar discipline" section as the deferred follow-up.

## Test plan

- [x] `./sbpp.sh phpstan` — clean. Deprecation-rules plugin wired in, `phpVersion: 80200`, baseline unchanged (zero new entries).
- [x] `./sbpp.sh test` — full suite green (366 tests, 1493 assertions). New `Php82DeprecationsTest` adds 5 cases (banlist + null-IP fixture row, commslist, admin.bans, admin.admins, OpenID construction).
- [x] `./sbpp.sh composer api-contract` — clean (no API surface changed).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -279,6 +279,49 @@ Playwright E2E specifics:
 - Pattern: `query` → `bind` → `execute` / `single` / `resultset`.
 - ADOdb was fully removed (commit `b9c812b2`). **Do not reintroduce it.**
 
+### Null-into-scalar discipline (PHP 8.2+)
+
+`web/composer.json` requires `php >= 8.2`, so PHP's
+"`Deprecated: <fn>(): Passing null to parameter #1 of type string`"
+surface is active. PHP 9 will turn it into a `TypeError`. Every
+`strlen` / `trim` / `substr` / `preg_match` / `mb_strlen` / etc.
+call against a value that can be `null` at runtime needs one of
+two idiomatic shapes (#1273):
+
+- **Coalesce** when null is semantically "absent" — e.g. a `$_POST`
+  / `$_GET` / `$_SESSION` / `$_SERVER` lookup the form may have
+  omitted:
+
+  ```php
+  if (strlen($_POST['password'] ?? '') < MIN_PASS_LENGTH) { … }
+  $name = trim((string) ($_POST['name'] ?? ''));
+  ```
+
+- **Cast** when the value should always be a string but the type
+  system can't see it (most often a nullable `:prefix_*` row column,
+  or a `mixed`-returning helper like `CUserManager::GetProperty`):
+
+  ```php
+  if (strlen((string) $row['ban_ip']) < 7) { … }
+  $steam = trim((string) $userbank->GetProperty('authid', $aid));
+  ```
+
+- Never `if (!is_null($x) && strlen($x) > 0)` — verbose and the
+  conditional reads worse than the coalesce.
+- `phpstan/phpstan-deprecation-rules` + `phpVersion: 80200` is the
+  static gate; `Php82DeprecationsTest` (PHPUnit) is the runtime gate
+  for the bits PHPStan doesn't see (excluded paths like
+  `includes/auth/openid.php`, runtime values that look non-null to
+  the type system but actually aren't).
+- `web/includes/auth/openid.php` is excluded from PHPStan and is
+  third-party-shaped: cast at function inputs (entry points) so the
+  diff stays bounded; never sprinkle `(string)` at every internal
+  call.
+- Replacing nullable `:prefix_*` columns with `NOT NULL DEFAULT ''`
+  would also clear the deprecation, but it's a paired schema
+  migration with a separate semantic change ("no IP" vs "empty IP")
+  — file separately, not as part of a deprecation sweep.
+
 ### Dev DB seeder (`./sbpp.sh db-seed`)
 
 `./sbpp.sh db-seed` populates the dev DB (`sourcebans`) with a deterministic,
@@ -828,6 +871,16 @@ audit (#1207) locked in. New CTAs:
   value is safe HTML; without a `{* nofilter: <why> *}` comment above
   it, future readers can't tell whether it's a real escape hatch or a
   copy-paste accident waiting to be exploited (#1113 audit).
+- `strlen($_POST['x'])` / `trim($_POST['x'])` / `substr($row['col'], …)`
+  on values that can be `null` at runtime → coalesce
+  (`strlen($_POST['x'] ?? '')`) when null is "absent", or cast
+  (`strlen((string) $row['col'])`) when the value should always be a
+  string. PHP 8.1 deprecated this implicit null-into-scalar coercion;
+  PHP 9 makes it a `TypeError` (#1273). The static gate is
+  `phpstan/phpstan-deprecation-rules` + `phpVersion: 80200`; the
+  runtime gate (for PHPStan-excluded files like `auth/openid.php`) is
+  `Php82DeprecationsTest`. See "Null-into-scalar discipline" in
+  Conventions for the per-shape idiom.
 - `setTimeout` / `waitForTimeout` waits in E2E specs → wait on
   terminal attributes (`[data-loading="false"]` settled, `[data-skeleton]`
   removed) per #1123's "Testability hooks" rule.
@@ -901,6 +954,7 @@ audit (#1207) locked in. New CTAs:
 | Populate the dev DB with realistic synthetic data (banlist > 1 page, drawer history, moderation queues, audit log) | `./sbpp.sh db-seed` → `web/tests/scripts/seed-dev-db.php` (CLI driver) → `web/tests/Synthesizer.php` (`Sbpp\Tests\Synthesizer`). Dev-only: refuses any `DB_NAME` other than `sourcebans` (so `sourcebans_test` / `sourcebans_e2e` stay untouched). Idempotent; deterministic given a fixed `--seed` (default `Synthesizer::DEFAULT_SEED`). Does NOT share plumbing with `Fixture::truncateAndReseed` — the e2e hot path stays minimal. |
 | API wire-format snapshots              | `web/tests/api/__snapshots__/<topic>/<scenario>.json`    |
 | Action -> permission lock              | `web/tests/api/PermissionMatrixTest.php`                 |
+| Trap PHP 8.1 null-into-scalar deprecations at runtime (the bits PHPStan can't see) | `web/tests/integration/Php82DeprecationsTest.php` (#1273) — process-isolated render harness with a stub Smarty + `set_error_handler` that promotes `E_DEPRECATED` / `E_USER_DEPRECATED` to `\ErrorException`. Mirrors the LostPasswordChromeTest stub-Smarty pattern; each test method runs in a separate process because the page handlers declare top-level helpers (`setPostKey()` etc.) that PHP can't redeclare in one process. Add a marquee route here whenever a new high-traffic page handler ships, especially if it reads nullable `:prefix_*` columns or `$_POST` / `$_GET` lookups. |
 | Add an E2E spec                        | `web/tests/e2e/specs/<smoke|flows|a11y|responsive>/...` + `web/tests/e2e/pages/...` |
 | Add a route to the screenshot gallery  | `web/tests/e2e/specs/_screenshots.spec.ts` (`ROUTES` array) |
 | Tweak mobile (<=768px) chrome layout   | `web/themes/default/css/theme.css` — see the `#1207` `@media (max-width: 768px)` blocks for the canonical shapes (icon-only topbar search, full-width drawer + scroll lock). Sub-paged admin routes (servers / mods / groups / settings) use the `<details open>` accordion in the `#1259` `@media (min-width: 1024px)` block (sidebar inline at `<1024px`, sticky 14rem rail at `>=1024px`); see "Sub-paged admin routes" in Conventions. |
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -736,6 +736,16 @@ removes an entry or when bumping the level.
 strings against the schema. Set `PHPSTAN_DBA_DISABLE=1` to skip it; CI
 sets `DBA_REQUIRE=1` so credential drift fails loudly.
 
+`phpstan/phpstan-deprecation-rules` (#1273) is wired in via
+`phpstan.neon` with `phpVersion: 80200` so the analyser flags the
+PHP 8.1 null-into-scalar deprecation surface (`strlen($null)`,
+`trim($null)`, `substr($null, ...)`, `preg_match($null, ...)`, …)
+before it bites us on the PHP 9 bump. `web/includes/auth/openid.php`
+is excluded from PHPStan, so the same surface there is gated by the
+runtime smoke test in `web/tests/integration/Php82DeprecationsTest.php`
+(per-process `set_error_handler` that promotes `E_DEPRECATED` to a
+thrown `ErrorException` while it requires each marquee page handler).
+
 ## Test architecture
 
 Tests live in `web/tests/`:
--- a/web/composer.json
+++ b/web/composer.json
@@ -40,6 +40,7 @@
     },
     "require-dev": {
         "phpstan/phpstan": "^2.1",
+        "phpstan/phpstan-deprecation-rules": "^2.0",
         "phpunit/phpunit": "^11",
         "staabm/phpstan-dba": "^0.4"
     },
--- a/web/composer.lock
+++ b/web/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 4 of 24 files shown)


rumblefrog (commented):
## Reviewer summary — APPROVE (gh blocks self-approve, posting as comment)

Audited per the issue's acceptance criteria. All gates clean.

### Acceptance criteria
- **`strlen` audit**: `rg 'strlen\(\$' web/pages web/includes web/api` returns 28 residual sites; all verified safe (type-hinted `string` params, cast at handler boundary, or pulled from already-cast inputs — see API handlers' `(string)($params['x'] ?? '')` boundary, `system-functions.php::trunc(string $text, …)`, `Version.php`'s `callable(): string` defaults, etc.). Coalesce + cast idioms applied per the issue's "per-site fix" guidance.
- **`trim` / `substr` / `preg_match` / `mb_strlen` audit**: same-shape audit applied.
  - `rg 'trim\(\$_(POST|GET|SESSION|SERVER)' web` → **0 hits**
  - `rg 'preg_match\([^,]+,\s*\$_(POST|GET|SESSION|SERVER)' web` → **0 hits**
  - `rg 'preg_replace\([^,]+,[^,]+,\s*\$_(POST|GET|SESSION|SERVER)' web` → **0 hits**
  - `rg 'substr\(\$_(POST|GET|SESSION|SERVER)' web` → **0 hits**
  - `rg 'mb_strlen\(\$_(POST|GET|SESSION|SERVER)' web` → **0 hits**
- **OpenID entry-point casts**: 5 documented entry points verified (constructor's `$_SERVER['REQUEST_URI']` + `REQUEST_METHOD`, `__set` magic for `realm`/`trustRoot`, `get_provider_name` `parse_url` boundary, both `openid_signed` `explode` boundaries in `getAxAttributes` / `getSregAttributes`). Diff stays inside `openid.php` per the strategy.
- **`phpstan/phpstan-deprecation-rules`**: `^2.0` in `require-dev`, locked to `2.0.4` in `composer.lock`. `phpstan.neon` includes the rules.neon path with a `#1273` comment, `phpVersion: 80200` set. `phpstan-baseline.neon` is unchanged — audit cleared everything PHPStan can see, so the "annotate deferred entries" requirement collapses to "no deferred entries to annotate".
- **PHPUnit smoke**: `web/tests/integration/Php82DeprecationsTest.php` traps `E_DEPRECATED` / `E_USER_DEPRECATED` via `set_error_handler` + `error_reporting(E_ALL)`, hits banlist (with seeded `ip = NULL` ban row to actually traverse the cast path), commslist, admin.bans, admin.admins, and the OpenID `LightOpenID` construction + `__set` write path. Each test runs `#[RunInSeparateProcess]` because page handlers declare top-level helpers PHP can't redeclare in one process.

### Spot-checks
- `page.banlist.php:726` — `strlen((string) $row['ban_ip']) < 7` ✓
- `admin.bans.php` lines 226/229/380/383/555/558/695/698 — all 4 pagination clusters flipped to `if ($prev !== '')` / `if ($next !== '')` (cleaner than a cast — the page-builder branches always assign strings) ✓
- `admin.edit.admindetails.php:97` (`strlen((string) $a_steam) < 10`), `:155` (`strlen((string) $_POST['password'])`), `:186` (`strlen((string) ($_POST['a_serverpass'] ?? ''))`) ✓
- `openid.php` entry-point casts at the 5 documented boundaries ✓
- `SmartyTemplateRule` swap from `isSubclassOf(string)` to `isSubclassOfClass(ClassReflection)` — small, correct, scoped to the new deprecation-rules plugin not tripping on our own infra ✓

### AGENTS.md adherence
- New "Null-into-scalar discipline (PHP 8.2+)" Conventions section is concise (42 lines) and matches the per-site fix idioms used in the diff (coalesce / cast, anti-pattern call-out, openid entry-point strategy, out-of-scope schema migration).
- Anti-patterns section gained a paired entry pointing back to the convention.
- ARCHITECTURE.md notes the static gate; "Where to find what" gained a row for the runtime smoke test.
- No `:prefix_*` columns replaced with `NOT NULL DEFAULT ''` — confirmed out-of-scope schema migration not introduced (no SQL files touched).

### CI
All 5 gates green: PHPStan, PHPUnit (366 tests / 1493 assertions including the 5 new `Php82DeprecationsTest` cases), Playwright + axe, ts-check, api-contract.

### Per-file count
The PR body's per-file table (16 files × 64 sites) reconciles with the diff stat (23 files changed; the delta is 2 docs + composer.{json,lock} + phpstan.neon + the new test file + SmartyTemplateRule).

### Nit (non-blocking)
`web/includes/auth/handler/SteamAuthHandler.php:29` — `preg_match($pattern, $this->openid->data['openid_claimed_id'])` reaches into `$this->openid->data` (which is `$_POST` or `$_GET`). The OpenID `validate()` flow guarantees `openid_claimed_id` is set on a valid Steam round-trip (`openid.php:892`), but PHPStan can't see that (the array is `mixed`), and the new runtime smoke test only exercises construction, not the validation path. If a malformed Steam callback ever reaches `validate()` returning true without that key, this would emit `Deprecated: preg_match(): Passing null to parameter #2`. Unlikely in practice, but worth either an `(string) ($this->openid->data['openid_claimed_id'] ?? '')` cast or extending `Php82DeprecationsTest` to run a stub validate flow. Not blocking — the issue's audit list doesn't enumerate this site, and the entry-point cast strategy at `openid.php:56` (where `$this->data` is assigned from $_POST/$_GET) is technically the canonical fix point if we wanted to thread a `(string)` through every value, which is exactly the "diff sprawl" the worker correctly avoided.

Nothing blocking. Nice that the deprecation gate is now enforced both statically and at runtime — this should keep us off the PHP 9 fatal-error cliff.

PR #1284, opened 2026-05-07
fix(chrome): land the structural half of #1271 — footer inside .app so the sidebar stays sticky-pinned to the document bottom

### LINKED ISSUE #1271: Main sidebar (`.sidebar`) scrolls with page content — sticky breaks because flex `align-items: stretch` overrides `height: 100vh`

rumblefrog:
## Summary

- The original #1271 was misdiagnosed; #1278 added `align-self: flex-start` on a non-functional hypothesis (the commit message and CSS comment both admit this is "defensive parity" because explicit cross-sizes already win over flex `align-items: stretch` on evergreen browsers). Verified locally by reverting just `align-self: flex-start` and re-probing the layout — every scroll position produced byte-for-byte identical sidebar bounding-box coordinates with and without the declaration.
- The actual root cause is structural: `<footer class=\"app-footer\">` was rendered as a body-level sibling of `<div class=\"app\">`. `.sidebar`'s sticky containing block is `.app`, so `.app`'s height fell short of the document by `footerHeight`. Sticky correctly released as `.app`'s bottom entered the viewport, sliding the sidebar up and chopping off the brand. On barely-tall pages where `docHeight - viewport ≤ footerHeight` (e.g. `?p=admin&c=audit` on the bare e2e seed: doc=795, viewport=720), the entire scroll range fell inside the sticky-release phase and the sidebar visibly tracked the scroll — exactly the symptom rumblefrog originally reported.
- Move `<footer class=\"app-footer\">` inside `.app` (specifically as the last flex column item of `<div class=\"main\">`). Pair with `margin-top: auto` (the classic CSS sticky-footer pattern) so the footer hits the viewport bottom on short pages instead of floating halfway up. The `align-self: flex-start` from #1278 is retained as belt-and-suspenders parity with `.admin-sidebar` (#1259); the CSS comment is rewritten to credit the structural fix as load-bearing.

## Verification

Probe (Playwright, viewport 1280x720, dev seed):

| page | doc | viewport | sidebar.top @ scroll=0 | @ scroll=500 | @ scroll=BOTTOM |
| --- | --- | --- | --- | --- | --- |
| `?p=admin&c=bans` | 2121 | 720 | 0 | 0 | **0** (was -75) |
| `?p=home` | 1312 | 720 | 0 | 0 | **0** (was -75) |
| `?p=admin&c=audit` | 720 | 720 | 0 | 0 (no scroll) | n/a (was -75 on the only available scroll position) |
| `?p=banlist` | 720 | 720 | 0 | 0 (no scroll) | n/a |

I temporarily reverted just the structural fix (kept the new spec) to confirm it catches the bug:

```
2 failed
  [chromium] › specs/responsive/sidebar-sticky.spec.ts:78:9 › sidebar stays pinned at viewport y=0 across the entire scroll range of admin-bans
    Received: -75   Expected: >= -1
  [chromium] › specs/responsive/sidebar-sticky.spec.ts:183:9 › barely-tall page (audit log on the bare e2e seed) does not regress to \"sidebar tracks scroll\"
    Received: -75   Expected: >= -1
```

Post-fix both pass, plus all 14 sidebar/responsive specs:

```
14 passed (4.8s)
```

## Test plan

- [x] PHPStan — clean
- [x] ts-check — clean
- [x] Playwright e2e (chromium + mobile-chromium, workers=1) — 187 passed
- [x] New strict assertions confirmed to FAIL on pre-fix code (-75) and PASS on post-fix code (0)
- [x] Visual smoke at desktop on `?p=admin&c=bans` scroll=BOTTOM — brand visible at top, footer at viewport bottom, sidebar pinned through the full scroll range
- [x] Visual smoke at desktop on `?p=banlist` (short page) — footer pushed to viewport bottom by `margin-top: auto`, no awkward empty band
- [x] AGENTS.md cheatsheet updated (Where-to-find-what entry + two anti-patterns: don't pull the footer back out, don't pull the drawer/palette in)

Closes #1271 (the actual bug, this time).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -980,6 +980,32 @@ per spec, not the shared `sourcebans_e2e`. Wrapper:
   the sibling theme-toggle's chrome. The `.topbar__search-label` /
   `.topbar__search-kbd` spans stay in the DOM for SR users + the
   Mac glyph swap, but `display: none` everywhere — don't unhide them.
+- Moving `<footer class="app-footer">` back outside `<div class="app">`
+  (the body-level sibling shape from before #1271's structural fix) →
+  `.sidebar` is `position: sticky; top: 0; height: 100vh` and its
+  sticky containing block is `.app`. Pulling the footer out leaves
+  `.app` `footerHeight` short of the document, so on tall pages the
+  sidebar releases at the bottom (brand cuts off) and on barely-tall
+  pages where `docHeight - viewport ≤ footerHeight` (the bare-e2e
+  `?p=admin&c=audit` shape) the entire scroll range falls inside the
+  release phase and the sidebar appears to track the scroll — exactly
+  the symptom rumblefrog reported in #1271. The footer must stay as
+  the last flex column item of `<div class="main">`. The
+  `align-self: flex-start` on `.sidebar` (added by #1278) is
+  defensive parity with `.admin-sidebar`, NOT the load-bearing fix —
+  a future refactor that puts the footer back outside `.app` will
+  silently regress even with `align-self` in place. The regression
+  guard is `web/tests/e2e/specs/responsive/sidebar-sticky.spec.ts`'s
+  strict `top===0` assertion at scroll=`document.scrollHeight`.
+- Pinning `<aside id="drawer-root">` or `<dialog id="palette-root">`
+  inside `<div class="app">` "to be consistent with the footer" → the
+  drawer is `position: fixed; inset: 0` and `<dialog>` promotes itself
+  to the top layer when `showModal()`-ed; both are floating overlays
+  whose containing block is irrelevant to layout. Nesting the drawer
+  inside `.app` would extend `.app`'s flex cross-axis to the drawer's
+  `100vh` (and break the sticky containing-block math the same way
+  the pre-#1271 footer did, just from the opposite direction). Keep
+  them as direct children of `<body>`, after `</div>` closing `.app`.
 
 ## Where to find what
 
@@ -1026,6 +1052,7 @@ per spec, not the shared `sourcebans_e2e`. Wrapper:
 | Tweak mobile (<=768px) chrome layout   | `web/themes/default/css/theme.css` — see the `#1207` `@media (max-width: 768px)` blocks for the canonical shapes (icon-only topbar search, full-width drawer + scroll lock). Sub-paged admin routes (servers / mods / groups / settings) use the `<details open>` accordion in the `#1259` `@media (min-width: 1024px)` block (sidebar inline at `<1024px`, sticky 14rem rail at `>=1024px`); see "Sub-paged admin routes" in Conventions. |
 | Stop mobile browsers auto-linking SteamIDs / IPs as phone numbers | `web/themes/default/core/header.tpl` (`<meta name="format-detection" content="telephone=no…">` + `<meta name="x-apple-data-detectors">`) and the defensive `.drawer a[href^="tel:"]` reset in `theme.css` |
 | Lock page scroll while a modal-style chrome is open | `web/themes/default/css/theme.css` (`html:has(#drawer-root[data-drawer-open="true"]) { overflow: hidden; }` — pure-CSS, gates on the same `data-drawer-open` mirror theme.js sets, applies at every viewport so the drawer-open contract is symmetric desktop/mobile per the Linear/Vercel/Notion modal idiom) |
+| Keep the main sidebar sticky-pinned across the full document scroll (`<aside class="sidebar">`) | The structural half of #1271 lives in `web/themes/default/core/footer.tpl`: `<footer class="app-footer">` is rendered as the LAST flex column item of `<div class="main">`, INSIDE `<div class="app">`. `.sidebar`'s sticky containing block is `.app`; if the footer were a body-level sibling of `.app` (the pre-fix shape), `.app`'s height would fall short of the document by `footerHeight` and the sidebar would release at the bottom — brand cut off, on barely-tall pages (`docHeight - viewport ≤ footerHeight`, e.g. `?p=admin&c=audit` on the bare e2e seed) the entire scroll range would be in the release phase and the sidebar would track the scroll. Keeping the footer inside `.app` makes the sticky CB extend to the full document. The CSS half (`.sidebar { align-self: flex-start; }` from #1278) is defensive parity with `.admin-sidebar` and is RETAINED but not load-bearing on its own. The footer's `margin-top: auto` (`.app-footer` rule in `theme.css`) is the classic "sticky footer" pattern — pushes the footer to the bottom of `.main`'s flex column on short pages so the credit doesn't float halfway up the viewport. Regression guard: `web/tests/e2e/specs/responsive/sidebar-sticky.spec.ts` asserts strict `top===0` at scroll=`document.scrollHeight` on `?p=admin&c=bans` (the canonical tall page) AND on `?p=admin&c=audit` (the barely-tall page that historically presented the bug most visibly). |
 | Disable the chrome's slide-in / fade animations for `prefers-reduced-motion` users | `web/themes/default/css/theme.css` (`@media (prefers-reduced-motion: reduce)` global block — see the matching note in "Playwright E2E specifics" / Conventions) |
 | Run a stack in parallel with another worktree | Worktree-local `docker-compose.override.yml` (see "Parallel stacks") |
 | Local dev stack details                | `docker/README.md`                                       |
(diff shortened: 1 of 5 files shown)


rumblefrog:
Addressed adversarial review feedback in commit 05371e3e (one logical commit on top of the c1 structural fix; happy to squash on merge):

| # | Severity | What | Fix |
|---|---|---|---|
| 1 | nit | AGENTS.md anti-pattern claimed nesting drawer in `.app` "would extend `.app`'s flex cross-axis to the drawer's `100vh`" — `position: fixed` removes from flow, can't grow ancestor's cross-axis | Rewritten to credit the real concern: a future ancestor declaring `transform`/`filter`/`contain: layout` re-establishes the containing block for `position: fixed` descendants per CSS Position Module §3.2 |
| 2 | nit | Drawer described as `position: fixed; inset: 0` in three places — actual rule is `right: 0; top: 0; height: 100%; width: min(35rem, 100vw)` (right-pinned panel); `inset: 0` is on the separate `.drawer-backdrop` | Fixed in `core/footer.tpl` (×2) and `AGENTS.md` (×1) |
| 3 | nit | `.app-footer` parenthetical claimed `margin-top: 1rem` would "sit alongside `auto`" — same property, can't both apply | Rewritten to be honest: 1rem dropped because the sticky-footer pattern needs `auto` exclusively on this property; documented `padding-top: 2rem` as the lever for restoring prior 36px-total breathing room |
| 4 | nit | `.sidebar` comment framed `align-self: flex-start` as "parity with `.admin-sidebar`'s `align-self: start`" — same property name, different mechanism (`.admin-sidebar` is in a CSS *grid* where `align-self: start` is load-bearing; `.sidebar` is in a flex row with explicit `height: 100vh` where `align-self: flex-start` is a no-op) | Tightened the comment: `align-self: flex-start` here defends specifically against a future refactor that drops `height: 100vh` |
| 5 | nit | Same comment qualified its claim with "on evergreen browsers" — CSS Flexbox §9.4 is the spec, not a browser-version contingency | Replaced with the spec reference |
| 6 | nit | Spec used `.sidebar__brand` as a class-chain primary selector — AGENTS.md's testability-hook rule prohibits this | Added `data-testid="sidebar-brand"` to `core/navbar.tpl` (with a comment explaining it's the canary for the regression guard) and updated all three spec usages to anchor on the testid |
| 7 | nit | Audit-page docstring oversold the conjunctive guard — doesn't catch the joint regression where footer goes back outside AND `min-height: 100vh` is shaved off so `docHeight === viewport` post-regression | Tightened docstring to be honest about the narrow band; documented the path to closing the gap (seed a `:prefix_log` row from the e2e `Fixture` so the audit page is reliably scrollable). Admin-bans guard above remains the load-bearing tall-page invariant |

Verification:
- ts-check: clean
- `./sbpp.sh e2e --grep "sidebar" --workers=1`: 14 passed across chromium + mobile-chromium projects, no regressions
- Verified `data-testid="sidebar-brand"` lands in the rendered HTML via `curl | rg`

No structural / behaviour changes from c1 — this is purely correctness-of-comments plus one testability hook.

Issue #1275, opened 2026-05-07
Sub-route navigation is inconsistent: admin-admins + admin-bans use `#fragment` anchor jumps; admin-servers/-mods/-groups/-settings use `?section=` URL routing — unify on Pattern A everywhere


rumblefrog:
## Problem

The chrome on the admin sub-route sidebar is now visually identical across every dense admin route (post-#1259 / #1266 unification: 14rem rail at >=1024px, accordion at <1024px, iconed pill rows, dark-pill / brand-orange `aria-current="page"` treatment). But the **routing semantics underneath the same-looking chrome differ by route**:

| Route | Pattern | URL on click | Reload behaviour |
| ----- | ------- | ------------ | ---------------- |
| Admin → Servers | A | `?p=admin&c=servers&section=list` | Section preserved |
| Admin → Mods    | A | `?p=admin&c=mods&section=list`    | Section preserved |
| Admin → Groups  | A | `?p=admin&c=groups&section=list`  | Section preserved |
| Admin → Comms   | A (single-section) | `?p=admin&c=comms` | n/a |
| Admin → Settings | A | `?p=admin&c=settings&section=settings` | Section preserved |
| **Admin → Admins** | **B** | `?p=admin&c=admins#admins` | **Scroll position lost; URL hash drops** |
| **Admin → Bans**   | **B** | `?p=admin&c=bans#protests` | **Scroll position lost; URL hash drops** |

A user navigating from one admin sub-route to another (a common path — Groups → Admins → Bans is the canonical onboard flow for a new admin) experiences:

- Click a Groups sidebar entry → URL changes, page reloads, browser back button works.
- Click an Admins or Bans sidebar entry → URL gets a `#fragment`, page jumps via anchor, browser back goes to the *previous page* (not the previous section), and reloading the page lands you back at the top of the long scroll if the hash didn't survive the reload.

Same chrome, different mental model. The user-flagged report is "inconsistent UX" — the divergence reads as broken. **All admin sub-routes should use Pattern A** (real `?section=…` URL routing, like Admin → Groups already does).

## Reproduce

1. `./sbpp.sh up && ./sbpp.sh db-seed`. Sign in as `admin/admin`.
2. Visit Admin → Groups. Click `Add a group` in the sidebar.
3. Observe: URL changes to `?p=admin&c=groups&section=add`. Reload the page. The "Add a group" form is still active; sidebar still has it pressed.
[...]

PR #1317, opened 2026-05-10
Fix #1307: reset config.theme to default + document composer install on v2.0 upgrade

### LINKED ISSUE #1307: v2.0 upgrade UX gaps: force config.theme=default and document composer install prerequisite
## Problem

An operator upgrading a real v1.8.x production install to v2.0.0 hits two UX gaps that the updater pipeline doesn't currently catch. Both are upgrade-day defects: the panel doesn't crash in dev because the dev stack always starts fresh on `default` and always has a `vendor/` populated by `web-entrypoint.sh`. Real upgrades carry over operator state the dev stack doesn't.

### 1. `config.theme` is not reset to `default` after the upgrader runs

`config.theme` lives in `:prefix_settings` (seeded by `web/install/includes/sql/data.sql` line 48: `('config.theme', 'default')`). On a v1.x install with a forked theme it holds the fork's directory name (e.g. `tf2c`, `darkred`).

The updater UI itself runs against `default` — `web/init.php` lines 216-220:

```216:220:web/init.php
$theme_name = (Config::getBool('config.theme')) ? Config::get('config.theme') : 'default';
if (defined("IS_UPDATE")) {
    $theme_name = "default";
}
define('SB_THEME', $theme_name);
```

But that override is scoped to the `IS_UPDATE` request lifecycle (set in `web/updater/index.php`). Once the updater finishes and the operator clicks **Return to panel** (`updater.tpl` line 69), the next request goes through normal `init.php` bootstrap, reads `config.theme` from DB, and tries to render the v2.0 chrome against the fork theme's directory.

The v2.0 default theme is a complete rewrite (#1123 / #1207 / #1259 / #1275): new typed View DTOs (`Sbpp\View\*`), new sub-paged admin sidebar partial (`core/admin_sidebar.tpl`), `core/admin_tabs.tpl` reduced to a back-link, every template signature changed, MooTools / `xajax` / `sb-callback.php` removed, `openTab()` JS deleted. A fork theme inherited from v1.x literally does not contain the templates v2.0 expects to render — best case the operator gets `Smarty: Unable to load template …` fatals; worst case Smarty falls through to a half-rendered page where every variable is undefined.

The reporter's request: the updater migration that takes existing installs from v1.x → v2.0 should override `config.theme` to `default` so the panel actually loads after the upgrade. Forks that want their theme back can re-select it from Admin → Settings → Themes once they've ported it.

#1115 currently advises fork-theme authors to "diff against `web/themes/default/` and reconcile" but provides no tooling and doesn't reset the active theme. That's correct for the **port** path, but the v1.x fork theme still being **active** between "operator finishes the updater" and "operator finishes the port" is the broken state this issue is about.

### 2. `composer install` is undocumented for the v1.x → v2.0 upgrade path

v2.0 ships new PHP dependencies that v1.x didn't have (`web/composer.json` requires):

- `symfony/mailer ^7.2` (the v1.x `PHPMailer` was removed at v2.0)
- `league/commonmark ^2.5` (the dashboard intro renderer added at #1113)
- `lcobucci/jwt ^5.0` major-bumped from v1.x's `^4`
- `smarty/smarty ^v5.4` major-bumped from v1.x's `^v3`
- `php >= 8.5` (raised at #1289 from v1.x's `>=7.4`)

`web/init.php` lines 60-62 hard-`die()` if `vendor/autoload.php` is missing:

```60:62:web/init.php
if (!file_exists(INCLUDES_PATH.'/vendor/autoload.php')) {
    die('Compose autoload not found! Run `composer install` in the root directory of your SourceBans++ installation.');
}
```

That message is fine for a fresh install with no `vendor/`, but the realistic upgrade shape is **`vendor/` is present but stale** — the operator pulled v2.0 over an existing v1.x git tree, or unzipped a tarball into a tree that already had a `vendor/` from v1.x's `composer install`. In that case the autoload check passes, and the panel fatals at runtime the first time v2.0 code references `Symfony\Component\Mailer\…` or `League\CommonMark\…` — usually mid-way through a page render, not at bootstrap, so the operator hits a 500 with the actual class-not-found buried in their PHP error log.

What the docs say today:

- `README.md` mentions `composer install` only for fresh master-branch installs and explicitly says **"release versions come bundled with all required code dependencies"** — operators reading those instructions reasonably expect tarball upgrades to "just work" without a composer step.
- `UPGRADING.md` does not mention `composer install` at all (it currently only covers the telemetry-disclosure surface).
- The upgrader UI (`web/themes/default/updater.tpl`) does not surface the requirement either.

The reporter typo'd "composer updater" but the intent is clear: they want this called out somewhere the operator will actually read **before** the upgrade.

### Theme drift after upgrade

1. Stand up a v1.8.x panel (release tarball or local 1.x build). Sign in.
2. Admin → Settings → Themes → switch to a forked theme (any non-`default` directory under `web/themes/`). Verify `config.theme` in `:prefix_settings` is the fork's directory name.
3. Drop a v2.0 working tree onto the install (mirror the operator-facing `git pull` / unzip-tarball-over-tree path).
4. Visit `web/updater/index.php`. Watch the wizard run cleanly through migrations 801 → 807 (the wizard itself renders correctly because of the `IS_UPDATE` override).
5. Click **Return to panel**.
6. Observed: `Smarty: Unable to load template 'page_admin_dashboard.tpl' in 'themes/<fork>/'` (or equivalent) — the fork theme has no v2.0 chrome.
7. Expected: the panel loads on `default`. The operator can re-select their fork theme from Admin → Settings → Themes once they've ported it (per #1115's "Theme authors" guidance).

### Stale `vendor/` after `git pull`

1. Stand up a v1.8.x panel, run `composer install` so `vendor/` is populated with v1.x deps.
2. `git pull origin main` (mirror the git-based upgrade path) to get v2.0. Don't run `composer install`.
3. Visit any panel route.
4. Observed: 500 with `Class "Symfony\Component\Mailer\Mailer" not found` (or league/commonmark equivalent) in the PHP error log — the bootstrap autoload check passes because `vendor/autoload.php` exists, but the contents are v1.x.
5. Expected: a docs-side warning (UPGRADING.md / README.md) the operator hits **before** they pull, AND/OR a runtime banner in the updater UI noting "v2.0 introduces new PHP dependencies — run `composer install` from `web/` before continuing."

## Suggested fix

Both halves land as one PR (they share the same reviewer mental model: "v2.0 upgrade-day UX") but are independently scoped.

### Half 1 — paired updater migration to reset `config.theme`

Add `web/updater/data/808.php`:

```php
<?php

// Issue #NNNN: v2.0 ships a complete chrome rewrite (#1123 / #1207 /
// #1259 / #1275) — new typed View DTOs, new admin sidebar partial,
// every template signature changed. Fork themes carried over from
// v1.x do not contain the v2.0 templates and fatal on the first
// post-upgrade page load. Force config.theme back to 'default' so the
// panel actually loads after the upgrade; operators re-select their
// fork from Admin → Settings → Themes once they've ported it (#1115).
//
// Idempotent: the WHERE clause matches no rows on a re-run after the
// first successful pass, and matches no rows on installs that were
// already on 'default' to begin with.
//
// `$this` is supplied by Updater::update() which loads this file
// inside the Updater instance scope; PHPStan can't see that, so the
// next two calls are suppressed in the same way every sibling
// migration would be.
// @phpstan-ignore variable.undefined
$this->dbs->query(
    "UPDATE `:prefix_settings` SET `value` = 'default' "
    . "WHERE `setting` = 'config.theme' AND `value` <> 'default'"
);
// @phpstan-ignore variable.undefined
$this->dbs->execute();

return true;
```

Register in `web/updater/store.json`:

```json
"808": "808.php"
```

Per AGENTS.md "Updater migrations": idempotent UPDATE, `:prefix_` placeholder, `$this->dbs` with `// @phpstan-ignore variable.undefined`, defaults match `data.sql` (which already seeds `('config.theme', 'default')`). The migration is the canonical "force the new default" mechanism — same shape #1290 phase D etc. used for similar one-time forward-converging changes.

Open question for the maintainer:

- **Aggressive (preferred, matches reporter's request)**: always reset to `'default'`. Operators re-select their fork from Admin → Settings → Themes if they've ported it.
- **Conservative**: only reset when `web/themes/<config.theme>/theme.conf.php` is missing on disk. Requires a `file_exists()` check inside the migration, which is doable but couples the DB-side migration to the filesystem layout. Skip unless the maintainer specifically asks.

### Half 2 — document the `composer install` prerequisite

`UPGRADING.md` gets a new section above the existing telemetry one:

```markdown
## Updating PHP dependencies (#1126, v2.0.0)

v2.0.0 introduces new PHP dependencies (Symfony Mailer, league/commonmark, …)
and bumps the major versions of others (Smarty 5, lcobucci/jwt 5). The
release tarball ships a fresh `web/includes/vendor/` so tarball upgrades
work out of the box; **git-based upgrades require `composer install` after
the pull**. Without it, the panel fatals on the first request that
references a v2.0-only class.

```sh
cd /path/to/sourcebans/web
composer install --no-dev --optimize-autoloader
```

Run this **before** visiting `web/updater/index.php`. The updater itself
loads `web/init.php`, which fails fast with `Compose autoload not found`
if `vendor/` is entirely missing — but a v1.x install with a stale
`vendor/` populated by an old `composer install` will pass the autoload
check and fatal at template-render time instead.
```

Optionally `README.md`'s `## How to install the current master branch version` section gets a one-line "Upgrading from 1.x? See [`UPGRADING.md`](UPGRADING.md)" cross-reference, parallel to the existing #1117 cross-reference plan.

Optionally a passive banner at the top of `web/themes/default/updater.tpl`'s "Progress" card calling the requirement out — but this is belt-and-suspenders; the docs are the load-bearing surface (the operator who didn't read UPGRADING.md may not read the banner either, and runtime detection of stale vendor isn't reliable from PHP).

## Notes

- **Reporter context**: 1.8 → 2.0 production upgrade.
- **Pair issues**: #1115 (Author UPGRADING.md for 1.x to 2.0, OPEN, milestone v2.0.0) — this issue's docs half is a concrete addition to that surface, not a competing one. #1117 (Stage 2.0.0 release readiness, OPEN, milestone v2.0.0) — relevant because the version bump shouldn't ship until the upgrade-UX surface is closed.
- **Why this isn't already covered by #1115**: #1115 is the umbrella "author UPGRADING.md" issue. Its body's `Theme authors` subsection talks about porting fork themes (the diff-and-reconcile workflow) but does **not** propose resetting `config.theme` server-side, and its `Operators` subsection lists the upgrade steps but does not call out `composer install`. Both gaps are real and documented separately here so the fix-PR has concrete scope.
- **Why this isn't covered by #1166 / #1269**: both ran (or were superseded by) the dry-run verification path; neither found these gaps because the dev stack always starts fresh on `default` and has `vendor/` populated by `web-entrypoint.sh` (`docker/php/web-entrypoint.sh` lines 79-81). Real upgrades from a 1.x tree don't.
- **Per AGENTS.md "Updater migrations"**: a paired migration is required whenever `data.sql` and existing-install state can diverge. `data.sql` seeds `config.theme = 'default'`; the new migration converges existing forks onto the same value at upgrade time.
- **Per AGENTS.md "Anti-patterns"**: editing `data.sql` (or `struc.sql`) without a paired `web/updater/data/<N>.php` is called out as one of the entries — this issue is the inverse case (DB state that diverges from `data.sql` for **operational** reasons, not schema reasons). Same convention applies; the `808.php` migration is the documented mechanism.
- **Out of scope**: changing how `config.theme` is set going forward (the Admin → Settings → Themes UI is fine — it's the v1.x-to-v2.0 transition that's broken, not the steady-state contract). Same for the `IS_UPDATE` override in `init.php`; that scoping is intentional and correct for the wizard surface.

rumblefrog:
Fixes #1307.

## Summary

v2.0 ships a complete chrome rewrite (#1123 / #1207 / #1259 / #1275).
Operators upgrading from v1.x hit two real-world UX gaps the dev stack
doesn't catch (it always starts fresh on `default` and always has a
populated `vendor/`). Both halves land here:

- **Half 1 — paired updater migration `web/updater/data/808.php`**:
  rewrites `:prefix_settings.config.theme` back to `default` for any
  install not already on the shipped theme. The updater wizard itself
  runs against `default` (`IS_UPDATE` override in `init.php`), but
  that scoping ends when the operator clicks "Return to panel" — the
  next request reads the stale fork value and Smarty fatals against
  v2.0 templates the v1.x fork doesn't contain. Idempotent (the
  WHERE clause excludes the already-set row), safe to re-run, and
  the default value matches the seed in
  `install/includes/sql/data.sql` so fresh and upgraded installs
  converge. Operators with a fork re-select it from Admin →
  Settings → Themes once they've ported it (per #1115).
- **Half 2 — `composer install` prerequisite documented**:
  `UPGRADING.md` gains a new "PHP dependencies" section above the
  existing telemetry one, plus a "Theme compatibility" section that
  surfaces half 1 for docs-first readers. `README.md` gets a
  one-line cross-reference at the head of the master-branch install
  instructions so admins reading those don't miss `UPGRADING.md`.
  v2.0's autoload check only verifies `vendor/autoload.php` exists,
  not that it's v2.0-shaped — so panels with a stale v1.x `vendor/`
  pass bootstrap and 500 mid-render with the actual class-not-found
  line buried in the PHP error log.

Per `AGENTS.md` "Updater migrations": idempotent UPDATE, `:prefix_`
placeholder, `// @phpstan-ignore variable.undefined` on each
`$this->dbs` call, defaults converge with `data.sql`, registered in
`store.json`. Per "Keep the docs in sync": `UPGRADING.md` is the
load-bearing operator surface for upgrade-day defects, so the docs
land in the same PR as the migration.

## Test plan

- [x] `./sbpp.sh phpstan` → no errors (no new violations introduced).
- [x] `./sbpp.sh test --filter=UpgradeThemeReset` → 4/4 pass:
  - `testForkThemeIsResetToDefault` — fork value rewritten on first run.
  - `testInstallAlreadyOnDefaultIsUntouched` — WHERE clause excludes
    the already-set row.
  - `testRerunImmediatelyAfterFirstPassIsNoOp` — second run matches
    zero rows (idempotency guarantee).
  - `testMigrationDefaultMatchesDataSqlSeed` — the migration's
    `'default'` literal matches `data.sql`'s `('config.theme',
    'default')` seed; future renames of the shipped theme directory
    won't silently diverge fresh and upgraded installs.
- [x] `./sbpp.sh test` (full suite) → 407 tests, 1778 assertions, all OK.

`ts-check` / `composer api-contract` / `e2e` weren't run because this
PR doesn't touch `web/scripts/`, `web/themes/*/js/`, `web/api/handlers/`,
or any user-facing UI surface.


rumblefrog:
Post-merge adversarial review (PR was already merged when I ran). One real finding plus a sweep of the rest.

## What checks out

- Migration shape per AGENTS.md "Updater migrations": idempotent UPDATE (WHERE excludes already-`default` rows), `:prefix_` placeholder, `// @phpstan-ignore variable.undefined` on each `$this->dbs` call, registered as `"808": "808.php"` (next integer above 807, no clash with #1300's telemetry migration).
- `'default'` literal matches `data.sql`'s seed; `testMigrationDefaultMatchesDataSqlSeed` locks that parity.
- Tests: 4/4 pass on the new file, 407/1778 on full suite. Confirmed adversarially — gutting the migration body to `<?php return true;` fails 3/4 (the no-op-on-default case trivially passes), and restoring it gets back to green.
- PHPStan: 229/229 files, no errors.
- Edge cases I poked at: empty string → reset (correct), case-variant like `'Default'` → not reset (collation-dependent, pre-existing concern not introduced here), NULL → impossible (`value` column is `NOT NULL`), case where operator runs migration on a fresh install that already has `default` → no-op via WHERE.
- Fork-theme directory left in place on disk — consistent with the doc's "your work isn't lost" framing.
- The aggressive (always-reset) approach matches the issue body's preference and the docs explain the re-select path clearly.

## Finding: doc inaccuracy in `UPGRADING.md`

The new "PHP dependencies" section claims:

> Release **tarballs** ship a fresh \`web/includes/vendor/\` so tarball upgrades work out of the box. **Git-based upgrades** ... need the explicit \`composer install\` above.

This is **factually incorrect** for the current release pipeline. Verified against the live 1.8.4 release zip:

\`\`\`
$ curl -sL https://github.com/sbpp/sourcebans-pp/releases/download/1.8.4/sourcebans-pp-1.8.4.webpanel-only.zip -o /tmp/x.zip
$ python3 -c \"import zipfile; z=zipfile.ZipFile('/tmp/x.zip'); print(f'total: {len(z.namelist())}, vendor entries: {sum(1 for n in z.namelist() if \\\"vendor\\\" in n)}')\"
total: 985, vendor entries: 0
\`\`\`

**Root cause**: \`.github/workflows/release.yml\`'s \`build-webpanel\` job (lines 62-79) does \`cp -R web /tmp/\${pkg}\` and zips, with **no \`composer install\` step**. \`web/includes/vendor\` is gitignored (`.gitignore` line 65), so the workflow's checkout doesn't have it and the zip doesn't either. This has been the case since the workflow's introduction in #1065 (May 2026), so all 1.8.x release zips lack vendor too — \`README.md\`'s "release version come bundled with all required code dependencies" claim has been quietly wrong for a while; this PR's UPGRADING.md inherits the same assumption.

**Operator impact**: an operator unzipping the v2.0 release tarball onto a fresh server hits \`web/init.php\`'s hard-die ("Compose autoload not found") because there's no vendor at all. They have to figure out \`composer install\` themselves. The PR's UPGRADING.md tells them this is only a git-pull problem, which is misleading.

**Suggested follow-up** (out of scope for #1307 strictly speaking — it's a release-pipeline gap that pre-dates this PR):

- **Preferred** — fix the workflow to actually ship vendor. Add to \`build-webpanel\`:
  \`\`\`yaml
        - name: Install PHP dependencies
          working-directory: web
          run: composer install --no-dev --optimize-autoloader --no-progress --no-interaction
  \`\`\`
  This makes the README and UPGRADING.md claims factually correct and unblocks tarball users.
- **Fallback** — clarify UPGRADING.md to acknowledge that BOTH paths require \`composer install\`. Keeps the doc honest without touching the pipeline, but leaves the README's "bundled" claim still wrong.

Probably worth filing as a separate issue against #1117's release-readiness umbrella. Happy to open it / a fix PR if useful.

## Status

Merge stands — the migration half is solid and the doc gap is fixable separately. Flagging here for traceability.

PR #1321, opened 2026-05-10
Fix #1305: reject stale tier-1 version.json below v2.0

### LINKED ISSUE #1305: Footer shows stale v1.8.x version after upgrade to v2.0 (resolver returns preserved `web/configs/version.json` verbatim)
## Problem

After upgrading an existing v1.8 production install to v2.0, the panel
footer still reads:

> SourceBans++ 1.8.1 | Git: 1434

instead of the expected v2.0 version string. `data-version="1.8.1"` is
also wrong on the rendered `<footer class=\"app-footer\">` — every
consumer that keys off the resolved version (telemetry, E2E specs,
bug-report templates) sees a phantom v1.8.1 install.

## Reproduce

1. Stand up a v1.x panel (any release that ships
   `web/configs/version.json` — i.e. anything before #1070, including
   the `1.8.1` tarball).
2. Upgrade in place to v2.0.0 by overlaying the v2.0 release tree on
   the v1.x install **without** explicitly overwriting
   `web/configs/version.json` (e.g. an FTP \"skip if exists\" upload, an
   `rsync` invocation that excludes `configs/`, or a manual
   directory-by-directory copy that treats `configs/` as user data
   alongside `permissions/web.json`).
3. Run `web/updater/index.php` to apply the schema migrations.
4. Sign in and look at the footer of any panel page.

Expected: footer reads `SourceBans++ 2.0.0 | Git: <short SHA>` (or
`dev` if no `version.json` is present).
Actual: footer reads `SourceBans++ 1.8.1 | Git: 1434`.

## Investigation

The footer is wired correctly end-to-end on the v2.0 side; the bug is
that `Sbpp\\Version::resolve()` accepts a stale v1.x `version.json`
verbatim with no sanity / version check.

- The chrome's footer template emits `{\$version}{\$git}` from the
  Smarty assigns in
  [`web/pages/core/footer.php:13-14`](web/pages/core/footer.php) which
  read `SB_VERSION` / `SB_GITREV`.
- Those constants are defined at
  [`web/init.php:95-98`](web/init.php) by unpacking
  `\\Sbpp\\Version::resolve(ROOT . 'configs/version.json')`.
- The resolver
  ([`web/includes/Version.php:63-94`](web/includes/Version.php))
  is a three-tier fallback: \`configs/version.json\` → \`git describe\`
  → the `'dev'` sentinel. Tier 1 wins whenever the file exists and
  decodes to a JSON object with a `version` key — there's **no
  validation that the value matches the panel's actual major
  version**.

The shipped v1.x repo carried
\`web/configs/version.json\` as a checked-in, hand-edited file with:

```json
{
    \"version\": \"1.8.1\",
    \"git\": \"1434\",
    \"dev\": true
}
```

That file was deleted from `main` in
[#1070](https://github.com/sbpp/sourcebans-pp/pull/1070) (commit
`9d1caefd`, May 2 2026): the release workflow now writes a fresh
`version.json` into each tarball at build time
([`.github/workflows/release.yml:65-72`](.github/workflows/release.yml)),
and `.gitignore` now ignores `web/configs/version.json` for dev clones
so the runtime falls through to `git describe`.

The v1.x `version.json` is the **exact** byte-for-byte source of the
reporter's symptom: `\"version\": \"1.8.1\"` → `1.8.1`,
`\"git\": \"1434\"` → ` | Git: 1434`. Nothing else in the codebase
emits the literal integer `1434` (verified with `rg \"1434\"` and
`rg \"Git: 1434\"`).

The upgrade path doesn't touch this file:

- The schema-only updater (`web/updater/data/*.php` +
  `web/updater/store.json`, walked by
  [`web/updater/Updater.php`](web/updater/Updater.php)) operates on
  the DB and never reads or writes `web/configs/version.json`.
- [`UPGRADING.md`](UPGRADING.md) doesn't call out this file as a
  manual replacement step.
- [`web/upgrade.php`](web/upgrade.php) only fires the
  `SB_SECRET_KEY` one-shot.

This means the resolver's tier-1 branch picks up whatever
`web/configs/version.json` happens to be on disk after the deploy. If
the deploy preserved the v1.x file (which several reasonable upgrade
methods will — operators commonly treat `configs/` as user data
because `configs/permissions/*.json` lives next to `version.json`),
the file's stale contents win.

The verification walk in #1269 didn't surface this because it used
`rsync -av --delete`, which removed the v1.x file (the v2.0 dev tree
doesn't contain it, so `--delete` wiped it and the resolver fell
through to git). A real-world admin doing a tarball overlay or a
selective rsync without `--delete` keeps the file.

This matches **Hypothesis B** from the triage prompt: the resolver is
reading `web/configs/version.json`, and a stale 1.8.x copy is on the
upgraded install. Hypothesis A (a `:prefix_settings` row) and
Hypothesis C (the footer rendering a different variable than the
resolver) are both ruled out by the code paths above.

## Suggested fix

Pick one of these (or pair them — they're not mutually exclusive):

1. **Updater migration** — add `web/updater/data/<N>.php` (next
   integer above `807`, registered in
   [`web/updater/store.json`](web/updater/store.json)) that
   **deletes** `web/configs/version.json` if it exists. Tarball
   installs immediately get the fresh file written by
   `.github/workflows/release.yml` over the deletion; git-checkout
   installs fall through to tier-2 (git describe); broken upgrades
   fall through to the `'dev'` sentinel — all three converge to a
   correct value, and the stale v1.x baked literal can never win
   again. Idempotent because the second run is a no-op (file already
   gone). Pair with `IF EXISTS`-style guard so the deletion is silent
   when the file isn't there.

2. **Resolver-side floor check** — `Sbpp\\Version::resolve()`
   compares the JSON's `version` to a hard-coded floor (e.g. `2.0.0`)
   via `version_compare(...)` and falls through to tier 2 if the
   on-disk value is below the floor. Cheap belt-and-suspenders, but
   the floor needs to track the panel version on every release — a
   future refactor that forgets to bump it silently re-opens the bug.

3. **Doc-only** — call out the manual replacement step in
   [`UPGRADING.md`](UPGRADING.md). Cheapest fix; relies on operators
   reading the doc and trusting they got the file. Not sufficient on
   its own — the structural fix (1) should still ship — but a useful
   second line of defence for in-the-wild installs that already shipped
   without it.

The cleanest contract is (1): the file is owned by the release
pipeline and the dev-clone fallback, never by upgraded installs.

## Notes

- Reporter symptom: `SourceBans++ 1.8.1 | Git: 1434` in the footer
  after upgrading a v1.8 install to v2.0.
- Root cause: `web/configs/version.json` was checked-in in v1.x
  (deleted from git in #1070); the resolver in
  [`web/includes/Version.php:63-94`](web/includes/Version.php) reads
  the file verbatim with no version sanity check.
- Verified by inspection — no existing tracking issue
  (`gh issue list --search 'version footer'`,
  `gh issue list --search 'configs/version.json'`,
  `gh issue list --search 'Git: 1434'` all empty).
- Surfaced after #1269's upgrade walk landed (which didn't catch it
  because the walk's `rsync --delete` deleted the stale file before
  the v2.0 panel ever rendered it).
- Once fixed, add a regression test on top of the existing
  [`web/tests/unit/VersionTest.php`](web/tests/unit/VersionTest.php)
  pinning whatever shape the chosen fix takes (e.g. \"version below
  the v2.0 floor falls through to tier 2\").

rumblefrog:
## Summary

`Sbpp\Version::resolve()` previously trusted any `configs/version.json`
verbatim. The v1.x repo carried that file as a checked-in, hand-edited
literal (`{"version": "1.8.1", "git": "1434"}`) until #1070 deleted it
and made the release pipeline its sole writer. An operator who
upgraded a v1.8 install to v2.0 with a "skip if exists" overlay tool
(FTP, `rsync` without `--delete`, manual directory copy that treats
`configs/` as user data alongside `configs/permissions/*.json`) keeps
the stale file on disk; the resolver returned its contents verbatim
and the chrome footer read `SourceBans++ 1.8.1 | Git: 1434` on a v2.0
install (with `data-version=\"1.8.1\"` poisoning telemetry, bug
reports, and E2E specs that key off the attribute).

This PR ships the issue's suggested fix (2) — resolver-side floor
check — paired with the doc-only fix (3) for already-upgraded
in-the-wild installs. It does not ship the migration-side fix (1):
the migration could only delete the stale file, which would
*regress* correctly-overlaid installs (the new `version.json` would
be wiped and the footer would read `dev`); the resolver-side check
fails closed for the broken case without harming the working case.

- Add `Version::MIN_TIER1_MAJOR = 2` and gate the tier-1 input on
  it via a new `isAcceptableTier1Version()` helper. A version
  whose major component is below the floor falls through to tier-2
  (`git describe`) or tier-3 (the `'dev'` sentinel) — the operator
  sees a self-describing fallback instead of phantom v1.x copy.
- Major-only comparison (vs. a strict `version_compare(..., '2.0.0',
  '>=')`) is deliberate so pre-release tarballs like `2.0.0-rc.1` —
  which sort BELOW `2.0.0` under semver's "pre-release < release"
  rule — stay accepted. The constant's docblock calls out the
  "bump on every MAJOR release" maintenance burden, paired with a
  `testFloorConstantIsTwo()` regression guard so the bump becomes
  a deliberate test edit.
- Pin every reject branch (stale v1.x with git available + with git
  absent) and every accept branch (boundary `2.0.0`, pre-release
  `2.0.0-rc.1`, future `3.0.0`, plus a malformed-version data
  provider with six edge cases) in `web/tests/unit/VersionTest.php`
  — bringing the suite from 6 to 18 tests, all passing.
- Document the v2.0.0 file-ownership contract (and the recovery
  procedure for already-upgraded installs) in `UPGRADING.md`; cross-
  link the floor from `AGENTS.md` \"Where to find what\" and
  `ARCHITECTURE.md`'s init.php lifecycle step 5.

Fixes #1305.

## Test plan

- [x] `./sbpp.sh test tests/unit/VersionTest.php` — 18 / 18 pass
  (6 pre-existing + 12 new across `testStaleV1JsonFallsThroughTo*`,
  `testTarballJsonAtFloorMajorIsAccepted`,
  `testTarballJsonPreReleaseWithinCurrentMajorIsAccepted`,
  `testTarballJsonFutureMajorIsAccepted`, the
  `provideMalformedTier1Versions` data provider, and
  `testFloorConstantIsTwo`).
- [x] `./sbpp.sh test` — 415 / 415 pass (the suite-wide PHPUnit
  deprecation that remains is `GroupsTest::testEditRoundTripsHighBitFlagsAsPositiveIntegers`'s
  `@dataProvider` doc-comment, pre-existing and out of scope).
- [x] `./sbpp.sh phpstan` — clean (no errors at level 5 + dba).
- [x] No `web/scripts/`, `web/themes/*/js/`, or `web/api/handlers/`
  touched, so `ts-check`, `composer api-contract`, and `e2e` are
  not in the diff path. The chrome footer template
  (`web/themes/default/core/footer.tpl`) still mirrors `SB_VERSION`
  unchanged; the only behaviour change is what `SB_VERSION` resolves
  to when `configs/version.json` carries a stale v1.x version.

## Reviewer notes

- The migration-side fix (issue option 1) was deliberately skipped
  — see Summary for the regression-on-correctly-deployed-installs
  reasoning. The resolver-side check + the doc note covers the
  failure mode without harming the working case.
- The floor needs to be bumped to `3` when v3.0 ships;
  `testFloorConstantIsTwo()` is the paired regression guard so
  forgetting to bump it shows up as a failing test rather than a
  silently re-opened bug. AGENTS.md's "Where to find what" row
  carries the same note.


rumblefrog:
Not a bug, the user was testing straight from main. On normal path, the release will contain CI generated version.json. We don't need to support other paths of cloning from main.

PR #1334, opened 2026-05-10
Fix #1332: friendly out-of-box installation

### LINKED ISSUE #1332: Make installation friendly out of the box: bundle vendor/ in release artifact, modernize install wizard, simplify README install path
## Context

A meaningful slice of self-hosters install SourceBans++ on shared
hosting (cPanel and similar) where they have File Manager and
phpMyAdmin but no SSH, no `composer`, and often no command-line
experience at all. The current install path has three friction points
that compound for those users — and the first two are bugs (not just
UX rough edges) where the shipped behaviour contradicts the docs.

This issue captures the install-friction problem and the **single PR**
that fixes it. The fix targets installation friendliness in general:
cPanel / shared-hosting users benefit automatically because they're
the most-friction slice, but the changes are not framed as
"cPanel-specific" — every user gets a friendlier on-ramp.

## The contradiction

`README.md` line 41 and `UPGRADING.md` lines 41–45 both promise that
release tarballs ship a fresh `web/includes/vendor/`:

> Release **tarballs** ship a fresh `web/includes/vendor/` so tarball
> upgrades work out of the box.

But the `build-webpanel` job in `.github/workflows/release.yml` just
runs `cp -R web /tmp/${pkg}` and zips the result. With
`/web/includes/vendor` in `.gitignore` (line 64–65), every published
release zip omits the directory entirely. So `web/init.php` lines
60–63 hard-fails with the documented `die()`:

```php
if (!file_exists(INCLUDES_PATH.'/vendor/autoload.php')) {
    die('Compose autoload not found! Run `composer install` in the root directory of your SourceBans++ installation.');
}
```

…the moment a non-technical user uploads the zip via cPanel File
Manager and visits `/install/`. Their option at that point is
\"learn SSH + composer\" or \"give up\".

The install wizard itself is also partially broken on `main`
post-#1123. `web/install/template/header.php` lines 8–9 still
reference deleted scripts:

```html
<script type=\"text/javascript\" src=\"../scripts/mootools.js\"></script>
<script type=\"text/javascript\" src=\"./scripts/sourcebans.js\"></script>
```

`web/scripts/mootools.js` was deleted alongside the v1.x
`web/scripts/sourcebans.js` at #1123 D1 (per `AGENTS.md`'s
\"Anti-patterns\" entry). The wizard-local
`web/install/scripts/sourcebans.js` survives but its `FadeElIn` /
`ButtonOver` / etc. helpers all depend on MooTools' `Fx.Transitions`,
`setStyle`, `setOpacity` — none of which load. Every `ShowBox(...)`,
`$E('html').onkeydown`, and `$('id')` call in `page.1.php` through
`page.6.php` (six pages, dozens of sites) references a function that
no longer exists. The forms still submit (they're plain HTML), but
every error popup, every Enter-to-submit shortcut, and every
keyboard navigation hint is silently dead. The wizard also still
renders pre-2.0 chrome (inline `listtable_top`, `submit-main` divs,
no design-system tokens) that visually contradicts the post-#1123
default theme.

The README's \"How to install\" path is also structured for
developers, not end users. Lines 38–62 give the release-tarball flow
~3 lines and the master-branch + `composer install` flow ~7 lines.
A non-technical user skimming the README has no clean primary path:
the very first install instructions they see still mention
`composer`. The fix isn't a dedicated \"shared hosting\" section —
it's making the **default** install path obvious and friction-free
for everyone, which incidentally solves the cPanel case.

## The PR

The three workstreams below ship as a single PR. They share a
\"make install work without command-line skills\" thesis and only
deliver value when shipped together: bundling `vendor/` without
fixing the wizard means users still hit the broken wizard on first
run; fixing the wizard without bundling `vendor/` means they never
reach it; cleaning up the README without either still points users
at a release zip that doesn't work. A single coherent PR is the
smallest unit that produces a working release-zip → working
install-wizard → working panel path.

### Workstream A — Bundle composer dependencies in the release artifact

- [ ] `.github/workflows/release.yml` `build-webpanel`: install PHP
      8.5 + composer (`shivammathur/setup-php@v2`) and run
      `composer install --no-dev --optimize-autoloader --no-interaction --prefer-dist`
      against the staged `/tmp/\${pkg}/` tree before zipping.
      Stage-not-source so the source checkout stays clean.
- [ ] Add a smoke step that asserts
      `\${pkg}/includes/vendor/autoload.php` and
      `\${pkg}/includes/vendor/smarty/smarty/libs/Smarty.php` are
      present in the artifact before upload, so this can never
      silently regress.
- [ ] Decision (cite-able for future contributors): do not strip
      `web/install/`, `web/updater/`, `web/tests/`, or
      `web/scripts/api-contract.js` from the release zip — they're
      load-bearing for self-hosters (installer wizard, upgrade path,
      frontend type-check artifact). The tarball stays a clean copy
      of `web/` with `vendor/` populated.

### Workstream B — Simplify the README install path

- [ ] Restructure `README.md` lines 38–62 so there is **one**
      primary install path (download the release zip, extract, visit
      `/install/`) presented first, in plain steps that don't
      reference `composer` or the command line.
- [ ] Move the master-branch / git-checkout developer flow (which
      still needs `composer install`) into a clearly-labelled \"For
      developers / contributors\" subsection beneath it, alongside
      the existing docker quickstart. No new \"shared hosting\" /
      \"cPanel\" section — the primary path itself is friendly enough
      that it works for any host with PHP + MySQL + a way to upload
      files.
- [ ] Drop the now-stale \"doesn't include the required dependencies\"
      caveat from the release-tarball mention (post-Workstream A it
      actively misleads users). The caveat survives in the
      developer-flow subsection where it correctly describes git
      checkouts.
- [ ] Optional within the same PR: ship one combined
      `sourcebans-pp-X.Y.Z.zip` artifact (panel + plugin in a single
      download) alongside the existing `.webpanel-only.zip` and
      `.plugin-only.tar.gz`, so the \"give me the one file\" path
      exists. Land if it doesn't bloat the PR; otherwise drop and
      this issue captures the follow-up.

### Workstream C — Fix + modernize the install wizard

- [ ] **C1 — kill the broken JS surface.** Remove the `<script>` tags
      from `web/install/template/header.php` lines 8–9 (refs to
      deleted `mootools.js` + the wizard-local `sourcebans.js`) plus
      every `ShowBox(...)` / `\$E(...)` / `\$('id')` site in
      `page.1.php` through `page.6.php`. Replace with vanilla-JS
      equivalents that match the `AGENTS.md` \"Frontend
      (`web/scripts/`)\" convention: `// @ts-check` + JSDoc,
      `document.querySelector`, native event listeners,
      `window.SBPP.showToast` for popups (so the wizard reuses the
      post-#1123 toast surface from the main panel).
- [ ] **C2 — chrome alignment.** Render the wizard against the v2.0
      default theme primitives: typed View DTOs under
      `Sbpp\\View\\Install\\*View.php`, Smarty templates under
      `web/themes/default/install/*.tpl`, the existing `theme.css`
      form / button / `.empty-state` rules, and the design-system
      tokens. Drop `web/install/scripts/sourcebans.js` entirely.
      Keep the routing wrapper at `web/install/index.php` — it's
      the public entry self-hosters bookmark.
- [ ] **C3 — in-wizard \"vendor/ missing\" recovery.** If a user lands
      on `/install/` from a git checkout (no `vendor/`), present a
      clear error with download-the-release-zip /
      SSH-to-run-composer instructions instead of the current
      `init.php` lines 60–63 `die()` text. Belt-and-suspenders for
      self-hosters who do checkout-based installs.

The MooTools, `ShowBox`, and `web/scripts/sourcebans.js` references
in the wizard are all already documented as forbidden in `AGENTS.md`'s
\"Anti-patterns\" section — the wizard is the last surface in the
codebase that violates them.

## PR-shaping note

Reviewer-friendly commit ordering: A (release.yml + smoke gate +
caveat tweak) → B (README install-path restructure) → C (wizard
rewrite, the bulkiest commit). C touches the most files; landing
C last keeps the diff legible and lets a `git bisect` cleanly land
on the wizard rewrite if a regression emerges. The PR description
should map commits to workstreams so reviewers can read each in
isolation.

## Out of scope (file separately if pursued)

- Self-extracting / web-based composer fallback (security trade-offs
  around running composer from a webroot context).
- One-click installer marketplace listings (Softaculous etc.) —
  business / packaging concern, not code.
- Auto-deletion of the `install/` directory after success. Would
  simplify the `init.php` lines 51–57 \"Please delete the install
  directory\" gate but raises permission-model questions about a
  webserver-writable install path.

## Acceptance criteria (PR-level)

- [ ] Downloading and unzipping the next `.webpanel-only.zip` to a
      fresh PHP 8.5 + MariaDB 10.11 host with no `composer install`
      step yields a working install wizard at `/install/`.
- [ ] The wizard's six pages render against the v2.0 default theme,
      with no browser-console JS errors, no 404s on `mootools.js` /
      `sourcebans.js`, and no references to `ShowBox` / `\$E()` /
      `\$('id')` in any wizard source file.
- [ ] The README's primary \"How to install\" path gets a user from
      \"I downloaded the release zip\" to a logged-in admin without
      referencing `composer` or the command line. The developer /
      git-checkout flow (which still needs `composer`) lives in a
      clearly-labelled subsection beneath it.
- [ ] CI: `release.yml` smoke gate fails the build if the staged
      zip is missing `includes/vendor/autoload.php`. `phpstan` /
      `ts-check` / `e2e` continue to pass on the wizard rewrite.

rumblefrog:
Closes #1332.

Three-workstream change. They share a "make install work without
command-line skills" thesis — A bundles the dependencies so the
release zip is a one-step install, B rewrites the README so a
non-technical operator finds the right path immediately, and C
fixes the wizard so it isn't visibly broken (and isn't running
its own JS bundle that's been dead since #1123 D1).

## Commits

| Commit | Workstream | Summary |
| --- | --- | --- |
| [`06af8fe8`](https://github.com/sbpp/sourcebans-pp/commit/06af8fe8) | A | `ci(release): bundle composer dependencies in webpanel release artifact` — `release.yml`'s `build-webpanel` job now installs PHP 8.5 + Composer, runs `composer install --no-dev --optimize-autoloader --no-interaction --prefer-dist` against the staged tree before zipping, and asserts both `includes/vendor/autoload.php` and `includes/vendor/smarty/smarty/libs/Smarty.php` exist before upload. The smoke step is the regression guard so this can never silently break again. |
| [`9e017928`](https://github.com/sbpp/sourcebans-pp/commit/9e017928) | B | `docs(readme): simplify primary install path` — README now leads with one plain-language path (download release zip → upload → visit `/install/`) with no `composer` mention. The git-checkout / `composer install` flow is moved into "For developers / contributors" beneath it. The misleading "release tarball doesn't include the required dependencies" caveat is dropped from the release section (post-A it actively misleads); it survives in the developer flow where it correctly describes git checkouts. |
| [`23e3f4b0`](https://github.com/sbpp/sourcebans-pp/commit/23e3f4b0) | C | `refactor(install): modernize wizard against v2.0 chrome` — Full rewrite of `web/install/`. Typed `Sbpp\View\Install\Install*View` DTOs + Smarty templates under `web/themes/default/install/` + procedural page handlers under `web/install/pages/page.<N>.php`. Adds a self-contained "vendor/ missing" recovery surface (`web/install/recovery.php`) for git-checkout / partial-upload installs. Deletes the legacy MooTools / wizard-local `sourcebans.js` / `web/install/template/*.php` procedural-PHP-template surface. Doc sync to `AGENTS.md` (anti-pattern + convention block + "Where to find what" rows) and `ARCHITECTURE.md` (directory layout + legacy-pattern table). |

## Test plan

The issue's acceptance criteria as a checklist:

- [x] `release.yml` builds the webpanel zip with `vendor/` baked in (smoke gate fails CI if either `autoload.php` or `Smarty.php` is missing).
- [x] `README.md`'s primary install path doesn't mention `composer` or the command line; downloading the release zip → uploading → visiting `/install/` is the one path presented first.
- [x] Developer / contributor flow (git checkout + `composer install` + `./sbpp.sh up`) survives in its own subsection.
- [x] No `<script src="../scripts/mootools.js">` or `<script src="./scripts/sourcebans.js">` in the rendered wizard chrome.
- [x] No `ShowBox(`, `$E(`, or `$('id')` calls anywhere under `web/install/`.
- [x] Wizard chrome uses the panel's `theme.css` design tokens (button / input / card / typography); install-only chrome (`.install-shell`, `.install-alert`, …) is inline in `_chrome.tpl`.
- [x] Each step is a typed `Sbpp\View\Install\Install*View` DTO rendered through `Sbpp\View\Renderer::render($theme, $view)`.
- [x] Vendor-missing path serves `recovery.php` (HTTP 503 + clear download-the-release-zip / SSH-and-run-composer instructions) BEFORE any namespaced class is referenced.
- [x] Existing wizard step structure preserved: 1=licence → 2=DB details → 3=requirements → 4=schema apply → 5=admin form + final config write → 6=optional AMXBans import.
- [x] `data.sql` and `struc.sql` schema content untouched (out of scope per the issue body).

Quality gates run from the worktree:

- [x] `./sbpp.sh phpstan` — clean.
- [x] `./sbpp.sh test` — 417 tests / 1871 assertions, all green.
- [x] `./sbpp.sh composer api-contract` — `git status` clean after regen.
- Manual smoke against the worktree's docker stack:
  - [x] `GET /install/?step=1` — licence page renders with the panel's design tokens, accept checkbox is required-gated.
  - [x] `POST /install/?step=2` with valid creds → renders the auto-submit handoff to step 3.
  - [x] `POST /install/?step=2` with bad creds → form re-renders with `data-testid="install-database-error"` populated.
  - [x] `POST /install/?step=3` with valid creds → requirements table renders, errors are pill-coded, blocked-by-errors state disables the Continue button.
  - [x] `GET /install/recovery.php` → HTTP 503 + the recovery copy ("Setup needed", `data-testid="install-recovery"`).

E2E suite (`./sbpp.sh e2e`) was NOT run because the wizard isn't covered by Playwright per `AGENTS.md` ("Playwright E2E specifics" — the suite reads against `sourcebans_e2e` which is post-install) and the change doesn't touch any panel surface that's exercised by the existing specs. None of the diff lands outside `web/install/`, `web/themes/default/install/`, `web/includes/View/Install/`, `.github/workflows/release.yml`, `README.md`, `AGENTS.md`, or `ARCHITECTURE.md`.

## Decisions / deferrals

- **Combined `sourcebans-pp-X.Y.Z.zip` artifact (panel + plugin)**: dropped from this PR per the issue's "land only if it doesn't bloat the PR" guidance. The current diff is already substantial (the wizard rewrite is the bulkiest commit). Filed as a follow-up — easy add to `release.yml`'s `build-combined-package` job once this lands.
- **CSRF in the wizard**: stays off. The panel's `Sbpp\Security\CSRF` reads from session + DB-loaded settings; both are unavailable pre-install. The pre-install attack surface is intrinsically limited (anyone who can reach `/install/` can also overwrite files via the same upload channel they used to deploy the panel) and the install-folder-still-exists guard in `web/init.php` blocks the panel from booting if the operator doesn't delete `/install/` after the wizard completes. Documented in the `AGENTS.md` install-wizard convention block.
- **State across steps**: hidden POST fields, never `$_SESSION`. Rationale documented inline on `InstallDatabaseView` + the new `AGENTS.md` convention block. A session-based handoff would silently leak credentials into the operator's tmp dir on an abandoned install.
- **Wizard-local JS scope**: vanilla only, no external dependency. The wizard has no logged-in user / no Config / no `$userbank`, so the panel's chrome JS (theme.js / palette / lucide.min.js) is structurally unavailable. Step 1's licence-accept checkbox guard and the auto-submit handoff form are the only JS in the rewrite — both are page-tail, both work without any global namespace.
- **`web/install/includes/converter.inc.php`**: deleted, AMXBans logic inlined into `pages/page.6.php` as `sbpp_install_amxbans_import()` — the legacy file `echo`-ed progress and called `die()` mid-render, which broke the new template handoff. The structured rewrite returns a status string the template surfaces in an alert.
- **`web/install/includes/main.css` + `web/install/includes/system-functions.php` + `web/install/images/`**: deleted. None referenced from the new wizard, all leftover from the v1.x procedural shape; `theme.css`'s design tokens cover what the wizard needs visually.


rumblefrog:
Adversarial review pass surfaced 3 critical SQLi paths + 4 major correctness regressions + 4 minor cleanups + 2 doc-sync items. All addressed in two follow-up commits, then rebased cleanly onto current `main`.

## Critical (all blocked + verified live against the worktree's docker stack)

| # | Finding | Fix | Verification |
|---|---|---|---|
| 1 | Step 4 schema install: prefix SQLi via `str_replace('{prefix}', $prefix, $sql)` against `struc.sql` | `sbpp_install_validate_prefix` re-run at top of `page.4.php`; rejects via 302 → step=2 BEFORE any SQL substitution | `prefix=evil`(`x` int);DROP TABLE secret_data;--` POST → 302 Location: ?step=2 |
| 2 | Step 5 data.sql replay: same shape, `str_replace('{prefix}', $prefix, $sql)` against `data.sql` | Same validator at top of `page.5.php` | Same 302 short-circuit |
| 3 | Step 6 AMXBans import: `amx_prefix` SQLi via `Database::setPrefix()`'s `str_replace(':prefix', $this->prefix, $query)` on the SOURCE-side SELECT | Same validator on `prefix` AND `amx_prefix` (the latter is operator input on the import page itself) | `amx_prefix=bans`+UNION+SELECT+1--` POST → form re-rendered with "AMXBans prefix must be 1-9 letters, digits, or underscores." inline error |

Single source of truth: new `sbpp_install_validate_prefix` in `web/install/includes/helpers.php` (1-9 ASCII letters/digits/underscores). Step 2's submit gate now calls the same helper so the validation can never drift between submit-time and defence-in-depth sites.

## Major (all addressed)

1. **`Database::__construct` calls `die()` on PDOException** → bypasses chrome with bare-text `SQLSTATE[HY000] [1045] Access denied` on credential typo. Fix: new `sbpp_install_open_db()` helper probes connectivity with a raw `\PDO` first so the failure surfaces as a catchable `\PDOException`; steps 3-6 wrap it in their existing try/catch and render the message inside the chrome. Step 2 already used a raw PDO probe by design — generalised the pattern. Verified: bad-creds POST to step 3 now renders the error inside the requirements table.

2. **`databases.cfg` snippet broken when DB password contains `"`** → e.g. `"pass" "p"a"ss"` (5 unescaped quotes). Fix: new `sbpp_install_kv_escape()` applies Valve KeyValues quoting (`\` → `\\`, `"` → `\"`) to host / database / user / password. Verified byte-level: `kv_escape(a\b"c)` → hex `61 5c 5c 62 5c 22 63` (matches Valve quoting rules exactly).

3. **`data.sql` replay fails with 1062 duplicate key, leaves `:prefix_settings` half-populated.** Fix: `INSERT` → `INSERT IGNORE` on all three rows (mods / settings / admins). Verified: PDO replay against the seeded dev DB → OK: 4, FAILED: 0.

4. **`install/init.php` pre-creates an empty `config.php`**, leaving a sharp edge if operator deletes `install/` mid-wizard (`web/init.php` sees zero-byte file as installed → `Undefined constant DB_HOST`). Fix: drop the pre-create. Page 3's existing `is_writable($configPath) || is_writable(PANEL_ROOT)` check already covers the file-doesn't-exist case via the second clause.

## Minor (4 of 6 addressed; 2 deferred with rationale)

* **#1 Lucide glyphs render as empty `<i>` tags** → 13 occurrences across 8 templates. The wizard chrome deliberately doesn't load `lucide.min.js` (the docblock in `_chrome.tpl` explains why), so each `<i>` rendered as a blank inline element. Stripped them all — buttons read fine without ("Back" / "Continue" / "Open the panel" carry the affordance).
* **#3 Stale `converter.inc.php` reference in `page.6.php`** → dropped.
* **#5 Auto-submit `<script>` outside `</body></html>` in `page_handoff.tpl`** → moved above `_chrome_close.tpl` include.
* **#7 `release.yml` smoke check verifies staged tree, not the actual zip** → switched to `unzip -l "${pkg}.zip" "${f}"` so a future `zip` flag that drops `vendor/` fails the build.
* *Deferred:* **#2 Stepper visual states identical** (UX nice-to-have, doesn't block functionality).
* *Deferred:* **#4 DB password in hidden input on done page** — there's no `<input type="password">` on `page_done.tpl`, only hidden inputs (which browsers don't auto-save). The reviewer's `autocomplete="off"` recommendation targeted password inputs specifically, not hidden fields.
* *Deferred:* **#6 `display_errors=1` leaks PHP warnings** — reviewer noted "acceptable trade-off" given the install context.

## Doc-sync (2 addressed; 2 deferred as pre-existing on main)

* **#1 CSRF rationale** in the install-wizard convention block claimed "session + DB-loaded settings; both are unavailable pre-install" — the DB-loaded-settings half is wrong. Rewrote to lead with the actual reasons (no logged-in user / bounded threat model / install-folder-existence gate) and cross-link to the `sbpp_install_validate_prefix` defence-in-depth pattern.
* **#2 Missing wizard row in "Keep the docs in sync" table** → added.
* *Deferred:* **#3 README MariaDB version inconsistency** — pre-existing on `main`, file separately.
* *Deferred:* **#4 README Upgrade section references stale paths** — pre-existing on `main`, file separately.

## Quality gates (re-run post-fix, post-rebase)

| Gate | Result |
|---|---|
| `./sbpp.sh phpstan` | clean (237 files, level 5 + dba) |
| `./sbpp.sh test` | 417 tests / 1871 assertions, all green (1 PHPUnit deprecation pre-existing on `main`) |
| `./sbpp.sh composer api-contract` | regen + `git status` clean |
| Manual smoke (curl against worktree's docker stack) | step 1 GET → 200, step 4 SQLi POST → 302, step 6 amx_prefix SQLi → form re-rendered with inline error, step 3 bad-creds → requirements table renders error inside chrome |

Branch rebased onto current `main` (b5d291be) with two AGENTS.md conflicts resolved (anti-pattern entry + "Where to find what" row — both clean "added entries on both sides" cases).

PR #1336, opened 2026-05-11
Fix #1335: post-install-wizard audit (panel-takeover paths + 8 UX gaps)

### LINKED ISSUE #1335: Post-#1332 install-wizard human-flow audit: panel-takeover paths via localhost host-check bypass + /install/ replayable on installed panels, plus 8 UX gaps
## Context

Follow-up to #1332 (merged via #1334). I drove a real Chromium browser through
the install wizard end-to-end against an empty `wizard_e2e` DB on the v2.0
codebase to audit the human flow that a non-technical self-hoster would
actually walk. The happy path is solid — license → DB → requirements → schema
(19 tables) → admin → done in ~7 seconds, no JS console errors, no uncaught
PHP errors, every error path renders a visible banner.

The walkthrough surfaced two paths that together let anyone who can reach a
panel at the right `Host` value with `install/` left in place commandeer it,
plus a handful of post-install UX papercuts that bite right after the user
clicks "Open the panel".

This issue rolls them up because the underlying symptom is the same: the
install wizard and the panel's first-boot guard don't share a security model,
so the friendly install we landed in #1332 doesn't quite carry through to
"install is done, panel is up".

---

### C1. Localhost bypass on the post-install safety check

`web/init.php:51` gates the `install/` + `updater/` presence check on
`HTTP_HOST != "localhost"`:

```php
if ($_SERVER['HTTP_HOST'] != "localhost" && !defined("IS_UPDATE")) {
    if (file_exists(ROOT."/install")) {
        die('Please delete the install directory before you use SourceBans++.');
    } else if (file_exists(ROOT."/updater")) {
        die('Please delete the updater directory before using SourceBans++.');
    }
}
```

Verified by curl against the live worktree: with `Host: localhost`, the panel
boots a normal dashboard even with `install/` still present. The
`README.md` install step 5 (#1332) reads "the panel refuses to load until you
do" — that claim is false on localhost, on any panel reachable via a
`localhost` Host header (port-forward, SSH tunnel, ngrok, Cloudflare Tunnel),
and on local-development workflows where the operator never sees the warning
they'll need to act on once they deploy.

The exception likely originated as "let local dev work without manual
chmod-style ceremony", but the panel doesn't actually need it: the install
wizard itself runs fine with `install/` in place; the guard only fires on the
*panel runtime* after the wizard completes.

**Suggested fix:** drop the `HTTP_HOST != "localhost"` exception, or move the
exception behind an explicit opt-in (e.g. `define('SBPP_DEV_KEEP_INSTALL',
true)` in `config.php`). Pair with a regression test that hits
`/index.php` with `Host: localhost` *and* `install/` present and asserts the
guard fires.

### C2. Wizard runs unconditionally over an already-installed panel

`web/install/index.php` has no "is the panel already installed?" gate. After
a successful wizard run, re-visiting `/install/` re-renders step 1 and lets
the operator walk the entire flow again, including overwriting `config.php`
(when writable), creating a new admin account, and re-pointing the panel at
a different DB. There's no warning and no confirmation.

Combined with C1 (or with any operator who simply forgot to delete
`install/`), this is a panel-takeover path: anyone who can reach
`/install/` can re-install over the live panel with arbitrary credentials.

**Suggested fix:** at the top of `web/install/index.php`, after `init.php`
runs, check `file_exists(PANEL_ROOT . 'config.php')` and either:
- refuse to start with a clear "this panel is already installed; if you want
  to reset it, delete `config.php` first" page (and link to the recovery /
  reinstall docs), or
- require an explicit `?confirm-reinstall=1` query param plus a typed-text
  confirmation on a soft-block screen.

The first option is safer (no in-band override). Pair with a regression test
that hits `/install/` with `config.php` present and asserts the guard fires
before the license page renders.

---

### M1. "Open the panel" CTA on the done page leads to bare-text `die()` errors

The done page's primary CTA (`a[data-testid="install-done-finish"]`) sends
the operator to `/`. `web/init.php` at that point may `die()` with one of:

- `SourceBans++ is not installed.` (`web/init.php:47`) — fires if `config.php`
  wasn't writable and the operator didn't yet paste in the manual snippet
  from done page section 3.
- `Please delete the install directory before you use SourceBans++.`
  (`web/init.php:53`) — fires if `install/` is still present (and we're not on
  the C1 localhost bypass).

Both `die()` paths return HTTP 200 with bare text — no chrome, no link to
docs, no `<a href="install/">`, no "here's how to fix it". A novice clicking
"Open the panel" before completing both manual post-install steps sees a
stark white page that reads like the server crashed.

**Suggested fix:** replace both `die()` calls with a small self-contained
HTML page (same shape as `web/install/recovery.php`'s pre-bootstrap
recovery surface — inline CSS, no Smarty / no vendor deps because we're
upstream of the autoload). Each page should:

- explain *what* happened (no `config.php` / `install/` still present)
- explain *how* to fix it (paste the manual snippet from the wizard's
  step 5 done page / delete the `install/` directory via FTP, SSH, or the
  hosting File Manager)
- link back to `/install/` so the operator can revisit the wizard if they
  closed the tab

Bonus: redirect missing-config to `/install/` instead of dying, so a user
who lands on `/` first gets dropped into the wizard automatically.

### M2. Wizard's writable-folder failures provide no remediation hint

`web/install/pages/page.3.php:159-161` builds the detail message for failing
filesystem checks as:

```php
$detail = !$exists
    ? 'Missing: ' . $path
    : ($writable ? 'Writable' : 'Not writable: ' . $path);
```

For the four hard-fail rows (`demos`, `cache`, `images/games`, `images/maps`),
the operator gets `Not writable: /path/to/web/demos` and a disabled
"Continue" button — and no instructions on how to make it writable. Combined
with the README never mentioning the writable-folder requirement (see m7
below), a non-technical operator on shared hosting can stall here with no
actionable next step.

**Suggested fix:** extend the detail message with a one-liner remediation:
"Not writable — set permissions to 0775 (or 0777 on shared hosting where
you don't control the PHP user) via your hosting File Manager / FTP client /
`chmod`." Optionally link to a docs page that walks through the cPanel /
DirectAdmin / Plesk variants.

### M3. Step 5 wipes both password fields on every validation re-render

Reproduced via the browser walkthrough — fill in the admin form with valid
username + email + a 32-char password + a *bad* SteamID, click Create admin.
The form re-renders with:

| Field | Preserved? |
|---|---|
| Username | yes |
| Email | yes |
| Steam ID (the bad value) | yes |
| Password | **wiped** |
| Confirm password | **wiped** |

The `InstallAdminView` DTO intentionally doesn't echo the password values
back into the template (correct — `nofilter`-into-HTML avoidance), but the
UX cost is that any of {bad email, bad steam ID, mismatched passwords,
short password} forces re-typing both passwords from scratch. With
`autocomplete="new-password"` set, browsers won't autofill them either.

**Suggested fix (preferred):** move the SteamID format check + email shape
check into client-side JS (page-tail script next to the existing
mismatched-password guard) so they fail before round-tripping to the
server. Server-side validation stays as the second line of defense, but
the round-trip-with-wiped-passwords path stops being the common case.

**Alternative:** keep server-side validation but reorder it so password
validations fire *first*; if those pass but a non-password field is
invalid, surface the error with a shorter "the rest of your form is fine,
just fix X" framing rather than an empty form.

---

### m1. `/install/recovery.php` returns HTTP 503 unconditionally

`web/install/recovery.php:25` always sets `http_response_code(503)` because
its contract is "we already know vendor/ is missing — the dispatcher in
`index.php` only require_once's us in that case". When the file is hit
directly (curious user, bookmarked URL, link share), it always shows the
"Setup needed — vendor missing" page even when vendor IS present.

Fix: at the top of `recovery.php`, self-check
`file_exists(PANEL_INCLUDES_PATH . '/vendor/autoload.php')` and 302 to
`/install/` if vendor is found.

### m2. "Licence" vs "License" inconsistency

Step 1 of the wizard uses British "Licence" exclusively (page title,
step title, step label, body copy, checkbox label, textarea heading).
Everywhere else in the repo — README, file paths (`page_license.tpl`),
test IDs (`install-license-*`), the project's own LICENSE.md — uses
American "License". Pick one (#1332 went American everywhere except
this one screen, so this is probably an oversight from the wizard
rewrite).

### m3. Step 2 `data-testid` prefix mismatch

Field test IDs use one prefix, form/button test IDs use another:

- Fields: `install-db-server`, `install-db-port`, `install-db-username`,
  `install-db-password`, `install-db-database`, `install-db-prefix`,
  `install-db-apikey`, `install-db-email`
- Form / buttons: `install-database-form`, `install-database-back`,
  `install-database-continue`, `install-database-error`

Every other step uses one consistent prefix (e.g. `install-license-*`,
`install-requirements-*`, `install-schema-*`, `install-admin-*`,
`install-done-*`). Friction for E2E spec authors — caught me writing
the audit script.

Fix: pick one (`install-database-*` is more consistent with the rest;
the field IDs `install-db-server` etc. were probably shortened
casually).

### m4. Bad-password error is the raw PDOException

`web/install/pages/page.2.php:68` surfaces:

> `Could not connect to the database: SQLSTATE[HY000] [1045] Access denied for user 'sourcebans'@'192.168.96.5' (using password: YES)`

`SQLSTATE[HY000] [1045]` is gibberish to non-DBAs, and the IP address is
the panel-as-seen-by-DB internal address — minor information disclosure.

Fix: pattern-match common PDO error codes ([1045] access denied, [2002]
host unreachable, [1049] unknown database, [1044] denied for user on
database) and emit a friendly translation; fall back to the raw message
for unrecognised codes.

### m5. License textarea is sized inconsistently with content

`web/themes/default/install/page_license.tpl:27` sets `rows="20"` on the
license textarea, but the rendered height is ~5 lines — the `.input`
class CSS in `theme.css` overrides the rows attribute. Cosmetic, but the
textarea looks suspiciously small given the wizard has a whole card
chrome around it.

Fix: either drop `rows="20"` (and use `min-height` in the inline style)
or override `.input`'s height for `textarea.input` specifically.

### m6. AMXBans form (step 6) has no field-level helper text

Step 2 has nice "Default for MySQL is 3306" / "Usually `localhost` on
shared hosting" / "Up to 9 letters / digits / underscores" hints under
each field. Step 6 (AMXBans import) has labels only — same fields, no
hints. Inconsistent voice; an operator unfamiliar with AMXBans gets no
guidance.

Fix: copy the helper-text pattern from `page_database.tpl` into
`page_import.tpl` for the matching fields, plus an AMXBans-specific
"Look in `addons/amxmodx/configs/sql.cfg` for these values" hint at the
top of the form.

### m7. README never mentions writable folders or chmod

`README.md` install steps 1-5 cover create-DB → download → unzip → run
wizard → delete install/, but never warn the operator that their PHP
user might need write access to `web/{demos,cache,images}` for step 4
(requirements check) to pass. On most cPanel / DirectAdmin / Plesk
shared hosts the upload + PHP user are the same, so things work
naturally — but on hosts where PHP runs as `nobody` or `apache` and the
extracted tree is owned by the operator's SSH user, they get stuck at
the requirements step with M2's no-remediation-hint detail and no
README signpost to recover.

Fix: add a short "If the wizard's environment check flags any folder
as not writable, set permissions to 0775 (or use your hosting File
Manager's Properties dialog)" line between steps 3 and 4.

---

## Methodology (one-off audit, NOT added to the e2e suite)

A standalone Playwright spec drove a real Chromium against the worktree's
`sbpp-1332-*` stack:

1. Removed `web/config.php` to simulate fresh install
2. Created an empty `wizard_e2e` DB granted to the `sourcebans` user
3. Walked the wizard step-by-step, capturing screenshots + console events at
   each step + every error path (bad credentials, bad prefix, mismatched
   passwords, bad SteamID, short password)
4. Verified post-install panel state (with `install/` present, with/without
   `config.php`, with `Host: localhost` vs other Host headers)
5. Restored `config.php` + dropped the test DB

The walkthrough script lived under `web/tests/e2e/_oneoff_install/` for the
duration of the audit and was deleted before report-out — the existing e2e
suite covers post-install panel behaviour against a Fixture-seeded DB and
deliberately doesn't drive the wizard itself (the wizard runs *before* the
panel is set up; the suite's auth + storage-state fixtures presume an
installed panel).

## What worked well (not findings, just for context)

- Happy path completes cleanly with no JS errors, no PHP errors, no broken
  resources
- Schema install creates 19 tables successfully; success banner is clear
- `databases.cfg` snippet renders correctly with `sbpp_install_kv_escape()`
  doing its job (verified the password values round-trip cleanly into the
  KeyValues snippet)
- Manual `config.php` fallback path works (when wizard can't write the
  file, the success page surfaces the rendered text for copy-paste)
- The defense-in-depth from the #1334 review (`sbpp_install_validate_prefix()`
  on every step, `sbpp_install_open_db()` wrapping `\Database` so connect
  failures land in the wizard chrome instead of a bare-text page,
  idempotent `data.sql` via `INSERT IGNORE`) all hold up under real
  browser interaction
- The bundled `vendor/` (the original ask of #1332) is paying off — no
  missing-autoload errors during the live walkthrough; the recovery page
  exists for the C3 fallback but isn't needed in the happy path

rumblefrog:
## Summary

Closes #1335.

End-to-end fix for the post-#1332 install-wizard human-flow audit. Two
critical security findings (panel-takeover paths) + three major UX
findings + seven minor papercuts, plus the paired regression tests
and doc updates.

### Critical (security)

- **C1** — Drop the `HTTP_HOST != "localhost"` exemption on `init.php`'s
  install/+updater/-presence guard. The exemption was a panel-takeover
  path on any panel reachable via a `localhost` Host header
  (port-forward, SSH tunnel, ngrok, Cloudflare Tunnel). Replaced with
  an explicit `SBPP_DEV_KEEP_INSTALL` constant (loud-named, only
  defined automatically by `docker/php/dev-prepend.php` for the dev
  stack). `IS_UPDATE` exemption preserved.
- **C2** — Refuse to start the wizard over an already-installed panel.
  Pre-fix, anyone reaching `/install/` after the wizard finished could
  walk it again and overwrite `config.php` / admin / DB. The new
  `web/install/already-installed.php` surface (pure inline HTML + CSS,
  same shape as `recovery.php` since it runs upstream of Composer)
  emits 409 Conflict and explains the intentional-reinstall path
  (delete `config.php` first).

### Major (UX friction at the worst possible moment)

- **M1** — Replace bare `die('SourceBans++ is not installed')` /
  `die('Please delete the install directory')` /
  `die('Composer autoload not found')` calls with self-contained chrome
  in `web/init-recovery.php`'s `sbpp_render_install_blocked_page()`.
  Bonus: missing-`config.php` now redirects to `/install/` instead of
  dying.
- **M2** — Extend page 3's writable-folder fail message with a chmod
  remediation hint.
- **M3** — Step 5's admin form: page-tail vanilla JS now validates
  SteamID format + email shape on submit (alongside the existing
  password-match check) so the round-trip-with-wiped-passwords path
  stops being the common case. Server-side validation stays as the
  second line of defense.

### Minor (papercuts)

- **m1** — `recovery.php` 302s to `/install/` when vendor IS present
  on a direct hit (was always 503).
- **m2** — Standardize "License" (American) across step 1.
- **m3** — Standardize step 2 field testids on `install-database-*`
  (was a mix of `install-db-*` for fields, `install-database-*` for
  form/buttons).
- **m4** — `sbpp_install_translate_pdo_error()` translates the four
  common connect-error codes (1045 / 2002 / 1049 / 1044) into
  operator-friendly messages; raw message preserved as fallback for
  unrecognised codes.
- **m5** — Step 1 license textarea: switch from `.input` (which
  overrides height) to `.textarea` with `min-height: 24rem` so the
  license is readable without scroll-noise.
- **m6** — Step 6 (AMXBans import): top-of-form pointer to
  `addons/amxmodx/configs/sql.cfg` + per-field helper text mirroring
  step 2's shape.
- **m7** — README install steps now signpost the writable-folder
  remediation between steps 3 and 4.

### Tests

`web/tests/integration/InstallGuardTest.php` — pure-function unit
tests pinning:

- `localhost` Host header does NOT bypass the install guard (C1
  regression)
- `IS_UPDATE` skips the guard (preserved legacy contract)
- `SBPP_DEV_KEEP_INSTALL` skips the guard (new dev escape hatch)
- the C2 already-installed predicate fires on `config.php` presence
- `sbpp_install_translate_pdo_error()` covers all four common codes
  and falls back gracefully on unknowns

### Doc updates

- **AGENTS.md** — Install wizard lifecycle, `Where to find what`
  (three new rows + refresh of the existing wizard-step row), four
  new Anti-patterns entries, and refresh of the wizard-vanilla-JS
  anti-pattern (license spelling + M3 step-5 additions).
- **ARCHITECTURE.md** — Web-panel directory layout (surface the new
  files), bootstrap step 2 (rewrite for the C1 + M1 contract),
  local dev-stack note (`dev-prepend.php` rewrite), legacy-patterns
  table (three new rows).
- **README.md** — m7 writable-folder signpost.

## Test plan

- [x] phpstan green (level 5 + dba, 239 files, 0 errors)
- [x] phpunit green (467 tests, 2089 assertions; 1 pre-existing
  deprecation, untouched by this PR)
- [x] ts-check green
- [x] api-contract clean (no diff after regen)
- [ ] CI e2e green — running on PR
- [x] Manual walkthrough: dev-stack panel boots ✓, `/install/` → 409
  Already-installed ✓, deleting `config.php` lets the wizard run ✓,
  `web/init.php` recovery surfaces render under all three scenarios ✓.

## Deviations from the issue

None. The issue suggested either dropping the localhost exception
unconditionally OR moving it behind a `define('SBPP_DEV_KEEP_INSTALL',
true)` opt-in; this PR takes the second path so the docker dev
workflow keeps working without operator intervention. Rationale:
the constant is loud-named (visibly wrong if defined in production),
defined only by `docker/php/dev-prepend.php`'s `auto_prepend_file`
(production panels have no path to define it), and documented as a
dev-only escape hatch in `AGENTS.md` Conventions + Anti-patterns.


rumblefrog:
Adversarial review findings addressed in three follow-up commits on this branch (be6d7b7a, a1943989, 8d5e6351). All four quality gates re-verified green; one new regression test method added.

## Fixes

- **M2 (P1) — split missing-folder hint from not-writable hint** (be6d7b7a). The first cut glued the same chmod hint onto both the `Missing:` and `Not writable:` branches; the missing-folder branch now points at re-uploading from the release zip / creating via File Manager (chmod can't fix something that doesn't exist). Lifted the detail-string construction into `sbpp_install_describe_filesystem_check()` in `web/install/includes/helpers.php` so the contract is unit-testable. New regression test `testFilesystemCheckEmitsDistinctRemediations` in `web/tests/integration/InstallGuardTest.php` pins both directions (missing branch must NOT mention chmod; not-writable branch must NOT suggest re-uploading).

- **M3a + M3b + M3c (P2) — drop `novalidate`, native attrs gate every input case** (a1943989). Picked **Option 1** of the two M3c choices: removed `novalidate` from the admin form so the browser's native `required` / `minlength="8"` / `pattern` / `type="email"` attrs become load-bearing again (per AGENTS.md's existing "native gate / JS polish" rule). Empty / short / pattern-mismatch / type-mismatch cases now surface native popovers before the JS handler runs; the JS shrinks to just the cross-field password-match check (the one validation native HTML can't express). M3a (short password) and M3b (empty fields) are both covered for free by the native-validation flip.

- **AGENTS.md update for the cross-field-validation contract** (8d5e6351). Added a `novalidate` carve-out to the install-wizard "Conventions for new wizard work" block spelling out the canonical cross-field shape, refreshed the wizard-vanilla-JS anti-pattern entry to drop the now-stale SteamID + email mentions, and added `sbpp_install_describe_filesystem_check` to the helper list in the "Edit a step of the install wizard" row. No anti-pattern exception note was needed (which would have been Option 2): the existing rule was already correct, the admin form's `novalidate` was just violating it.

## Quality gates re-run

| Gate | Status |
| --- | --- |
| PHPStan (level 5 + dba) | green (0 errors, 239 files) |
| PHPUnit | green (468 tests, 2096 assertions; +1 test method, +7 assertions vs review) |
| ts-check | green |
| api-contract | clean (no diff after regen) |

## P3 items

The reviewer's five P3 items are out-of-scope for this PR but tracked as a single follow-up at #1338 so they don't get lost — short summaries of each, ready for the maintainer to pick up later.

## Smoke tests

- M2: `./sbpp.sh exec php -r '...'` exercising `sbpp_install_describe_filesystem_check()` directly produces the three expected detail strings (missing → re-upload hint, not-writable → chmod hint, OK → bare "Writable").
- M3: walked the live wizard step 5 with a temporarily-removed `config.php`; confirmed the rendered `<form>` has no `novalidate`, all five inputs carry their native validation attrs (`required`, `minlength="8"` on both passwords, `pattern="STEAM_[01]:[01]:[0-9]+"` on SteamID, `type="email"` on email), and the page-tail JS is reduced to the password-match check only.

PR #1511, opened 2026-06-27
feat: add SBPP_BanPlayerBySteamId native for offline player bans

Sigibert:
## Summary

Adds `SBPP_BanPlayerBySteamId` — a new native for banning players who are no longer connected to the server.

`SBPP_BanPlayer` requires an in-game client index. This makes it impossible to ban players via automated systems (TK managers, anti-cheat, etc.) when the target disconnects before the ban fires. The new native accepts a SteamID string and player name directly, bypassing the need for an active client slot.

## API

```c
// sourcebanspp.inc
native void SBPP_BanPlayerBySteamId(int iAdmin, const char[] steamId, const char[] name, int iTime, const char[] sReason);
```

| Parameter | Description |
|-----------|-------------|
| `iAdmin` | Client index of the admin issuing the ban (0 = server/automated) |
| `steamId` | SteamID string in `STEAM_0:X:Y` format |
| `name` | Player name for display in the SBPP panel; may be empty string |
| `iTime` | Ban duration in minutes (0 = permanent) |
| `sReason` | Ban reason |

## Implementation

- Performs a duplicate-check SELECT first (same pattern as `CommandAddBan`) -- skips the INSERT if an active ban for that SteamID already exists
- Inserts into `sb_bans` with `length` in seconds (`iTime * 60`) and `ends = UNIX_TIMESTAMP() + (iTime * 60)`
- Uses `IFNULL((SELECT aid ...), '0')` for the admin ID lookup, matching the existing pattern
- Falls back to a server-lookup subquery for `sid` when `serverID == -1` (server not yet registered in SBPP)
- Fires the existing `SBPP_OnBanPlayer` forward with `iTarget = -1` to signal an offline ban to any listening plugins

## sourcebanspp.inc changes

- Declares the new native with a full doc comment
- Adds `MarkNativeAsOptional("SBPP_BanPlayerBySteamId")` in `__pl_sourcebanspp_SetNTVOptional` so plugins that use this native can still load when running without SBPP

### DIFF
--- a/game/addons/sourcemod/scripting/sbpp_main.sp
+++ b/game/addons/sourcemod/scripting/sbpp_main.sp
@@ -2399,6 +2399,12 @@ public int Native_SBPP_BanPlayerBySteamId(Handle plugin, int numParams)
 	GetNativeString(3, name, sizeof(name));
 	GetNativeString(5, reason, sizeof(reason));
 
+	if (strncmp(steamId, "STEAM_", 6, false) != 0)
+	{
+		ThrowNativeError(SP_ERROR_NATIVE, "SBPP_BanPlayerBySteamId: steamId must be in SteamID2 format (STEAM_X:Y:Z), got: %s", steamId);
+		return 0;
+	}
+
 	if (reason[0] == '\0')
 		strcopy(reason, sizeof(reason), "Banned by SourceBans");
 
--- a/game/addons/sourcemod/scripting/sbpp_main.sp
+++ b/game/addons/sourcemod/scripting/sbpp_main.sp
@@ -144,6 +144,7 @@ public APLRes AskPluginLoad2(Handle myself, bool late, char[] error, int err_max
 
 	CreateNative("SBBanPlayer", Native_SBBanPlayer);
 	CreateNative("SBPP_BanPlayer", Native_SBBanPlayer);
+	CreateNative("SBPP_BanPlayerBySteamId", Native_SBPP_BanPlayerBySteamId);
 	CreateNative("SBPP_ReportPlayer", Native_SBReportPlayer);
 
 	g_hFwd_OnBanAdded = CreateGlobalForward("SBPP_OnBanPlayer", ET_Ignore, Param_Cell, Param_Cell, Param_Cell, Param_String);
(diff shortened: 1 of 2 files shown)


game/addons/sourcemod/scripting/sbpp_main.sp
```diff
@@ -2382,6 +2383,125 @@ public int Native_SBBanPlayer(Handle plugin, int numParams)
...
+
+	DataPack pack = new DataPack();
+	pack.WriteCell(iTime);
+	pack.WriteString(reason);
+	pack.WriteString(steamId);
+	pack.WriteString(name);
+	pack.WriteString(adminAuth);
+	pack.WriteString(adminIp);
+
+	char steamIdEscaped[MAX_AUTHID_LENGTH * 2 + 1];
```
rumblefrog, line 2432: Need to check for steamid type since enforcement is done with SteamID2

Sigibert: I added a SteamID2 format check to the steamId provided by the native. Hope that fits ur concerns.

game/addons/sourcemod/scripting/sbpp_main.sp
```diff
@@ -2382,6 +2383,131 @@ public int Native_SBBanPlayer(Handle plugin, int numParams)
...
+	DB.Escape(steamId, steamIdEscaped, sizeof(steamIdEscaped));
+
+	char query[512];
+	FormatEx(query, sizeof(query), "SELECT bid FROM %s_bans WHERE type = 0 AND authid = '%s' AND (length = 0 OR ends > UNIX_TIMESTAMP()) AND RemoveType IS NULL",
+		DatabasePrefix, steamIdEscaped);
+
+	DB.Query(DB_OnBanBySteamIdSelect, query, pack, DBPrio_High);
+
+	Call_StartForward(g_hFwd_OnBanAdded);
```
rumblefrog: i don't know your use case, but nit if this should be checking the insertion result before firing, otherwe lgtm

Sigibert: Fixed in 97ed2bfc — the forward now fires from the INSERT callback (DB_OnBanBySteamIdInsert) only once results != null, instead of firing synchronously right after issuing the SELECT.

Sigibert:
I want to give it a little push. Anything missing for that to add?

game/addons/sourcemod/scripting/sbpp_main.sp
rumblefrog: Fixed in d9fd0090. Validation now uses the shared Steam2-to-Steam3 parser, which rejects anything outside exact STEAM_[01]:[01]:Z syntax and validates the numeric account component. Duplicate detection now also matches the universe-insensitive Y:Z enforcement shape.

game/addons/sourcemod/scripting/sbpp_main.sp
rumblefrog: Verified and retained in d9fd0090. The forward fires only after a non-null INSERT result. The callback now also resolves the admin by user ID so an asynchronous insert cannot attribute the event to a recycled client slot.

rumblefrog:
Thanks, I added some fixes from main.

PR #1534, opened 2026-07-30
feat(admin-nav): move config items into a sidebar accordion menu

### LINKED ISSUE #1490: Move admin interface configuration items into accordion menu
All of the admin sections which have sub-sections should have all of the menu items in the main sidebar, accordion style. Currently it is very wonky where there is a double "menu", it just reduces the available space in the main window. This could also reduce duplication, example under Admin > Servers, there is the add server "menu" item on the left, but also the same thing as a button on the right, no need for that.

Crude edit in what I mean.

Image

**SourceBans++ Version: Main**

maxijabase:
## Summary

Closes #1490

Admin section nav no longer lives in a second rail next to the main sidebar. Pattern A pages (`?section=…`) keep the same URLs and routing, but their children now open as accordions under Admins / Servers / Bans / Groups / Settings / Mods in the main sidebar. That frees ~14rem of content width and kills the double-menu feel the issue called out.

- **`Sbpp\View\AdminNavCatalog`** is the single source for section catalogs (navbar accordion + page-handler slug defaults). Permission and `config.enable*` gates match the old `AdminTabs` filter.
- **`AdminTabs`** with a non-empty tab list is a no-op. Empty calls still render the edit-* Back link.
- Duplicate header **Add …** CTAs on servers / admins lists are gone. Those destinations already sit in the accordion.
- Accordion chrome: nested indent, vertical rail under the parent icon, chevron on the summary.
- Protests / submissions Current|Archive chips share `.page-section` inset so they line up with the page body after the second rail went away.

Docs (`AGENTS.md`, `ARCHITECTURE.md`), E2E selectors, and PHPUnit (including `AdminNavCatalogTest`) updated for the new testids (`nav-admin-<endpoint>-<slug>`).

image

image

## Test plan

- [ ] Log in as owner. Expand Admins / Servers / Bans / Settings. Confirm children render, active leaf has `aria-current="page"`, category summary does not when children exist.
- [ ] Visit `?p=admin&c=servers&section=add` and `?p=admin&c=admins&section=add-admin`. Confirm no second left rail and no duplicate Add button in the page header.
- [ ] Ban protests / submissions. Confirm Current|Archive chips align with the heading and empty state (not flush against the sidebar).
- [ ] Partial-permission admin (e.g. list servers only). Confirm accordion children match reachable sections.
- [ ] Toggle off `config.enableprotest` / `config.enablesubmit` / `config.enablegroupbanning`. Confirm those Bans children disappear from the accordion.
- [ ] Edit-* page (e.g. edit ban). Confirm Back link still works.
- [ ] `./sbpp.sh phpstan` and `./sbpp.sh test --filter=AdminNavCatalogTest|AdminBansFeatureToggleTest|AdminAdminsSearchTest`
- [ ] `./sbpp.sh e2e --grep "admin-tabs|sidebar-sticky|admin-admins-density|routing-truthiness"`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -424,7 +424,8 @@ matching its directory. PSR-4 autoloads from `web/includes/` →
 | `Sbpp\Config`                            | settings cache                        |
 | `Sbpp\Api\Api`                           | JSON API dispatcher                   |
 | `Sbpp\Api\ApiError`                      | structured API error                  |
-| `Sbpp\View\AdminTabs`                    | admin sub-route sidebar mounter       |
+| `Sbpp\View\AdminTabs`                    | edit-* Back-link mounter (empty tabs) |
+| `Sbpp\View\AdminNavCatalog`              | Pattern A section catalogs for the main-sidebar accordion (#1490) |
 | `Sbpp\View\BrandLogo`                    | `template.logo` resolver with `is_file()` fallback to `images/favicon.svg` — single source for the navbar + login chrome brand-mark renders; rejects path-traversal + null-bytes + the v1.x default (case-insensitive); fail-closed on missing `SB_THEMES` |
 | `Sbpp\View\Toast`                        | server-side toast emitter (`emit(kind, title, body, ?redirect)` — replaces v1.x `ShowBox(...)`) |
 | `Sbpp\View\*`                            | view DTOs (page-level + partials)     |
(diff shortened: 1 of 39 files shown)


Rushaway:
Can you resolve conflicts @maxijabase ? 👍

